PR #17, opened 2011-01-08
Extend Stream to provide remove(count) and peek(offset)

blalor:
I needed to extend HardwareSerial to get more access to the internal buffer; for my project, it was wasteful to have two statically-allocated buffers for incoming serial data, but I needed to be able read arbitrary bytes of the receive buffer.  This is the result.  The comments on the individual commits should make it pretty clear what was done. :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+app/bin/
+app/pde.jar
+build/.DS_Store
+build/macosx/work/
+core/bin/
+core/core.jar
--- a/hardware/arduino/cores/arduino/HardwareSerial.h
+++ b/hardware/arduino/cores/arduino/HardwareSerial.h
@@ -26,6 +26,16 @@
 
 #include "Stream.h"
 
+// Define constants and variables for buffering incoming serial data.  We're
+// using a ring buffer (I think), in which rx_buffer_head is the index of the
+// location to which to write the next incoming character and rx_buffer_tail
+// is the index of the location from which to read.
+#if (RAMEND < 1000)
+  #define RX_BUFFER_SIZE 32
+#else
+  #define RX_BUFFER_SIZE 128
+#endif
+
 struct ring_buffer;
 
 class HardwareSerial : public Stream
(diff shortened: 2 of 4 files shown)


hardware/arduino/cores/arduino/HardwareSerial.h
```diff
@@ -50,11 +60,16 @@ class HardwareSerial : public Stream
       uint8_t rxen, uint8_t txen, uint8_t rxcie, uint8_t udre, uint8_t u2x);
     void begin(long);
     void end();
-    virtual int available(void);
-    virtual int peek(void);
-    virtual int read(void);
-    virtual void flush(void);
-    virtual void write(uint8_t);
+    
+    // @todo I think these only need to be declared virtual if HWSerial's going to be extended
```
Chris--A: No, virtual is not needed.

If a derived object implements `peek()` this is the one the virtual methods will call. For that object to use the parent versions, it'll have to explicitly call `derived::peek()`.

I wouldn't use `virtual` for another reason: A client may want to use the derived version without implementing their own `using derived::peek;`.

matthijskooijman:
Should a default peek() implementation also be provided (calling peek(0))? That allows subclasses to only implement peek(count) and eventually perhaps making peek() non-virtual. The downside of this is that if a subclass implements neither peek method, it will end up as infinite recursion instead of a compiler error, but that's probably acceptable?

NicoHood:
The passed variable type should be int or size_t instead of uint8_t. I'd go for size_t.

Also I'd remove the peek() virtual function and make it a normal function that calls peek(0). Libraries might need to add using::peek(); or something like that. Using too many virtual functions adds too much overhead. If we can avoid it, please do so.

I'd also add a function that resets the buffer (discards all bytes).

.gitignore
```diff
@@ -0,0 +1,6 @@
+app/bin/
```
NicoHood: And this gitignore file is really not needed.

facchinm:
Hi @blalor , I know this PR is more than 8 years old (forgive us if this happened) but due to reorganization the patch should be moved to https://github.com/arduino/ArduinoCore-api . Is it still something that you (or anyone else) might be interested to discuss?
In the meantime, I'm closing the issue since it doesn't belong here anymore. To anyone interested, please open a new PR/issue on ArduinoCore-api linking this one, thanks.

PR #18, opened 2011-01-11
Adapt arduino-usbserial to LUFA changes

rafl:
LUFA, which is used by the usbserial firmware on the 8u2 on the Uno and Mega 2560 boards, has seen various backward-incompatible changes lately. Part one of this changeset updates arduin-usbserial to work again with the changed API.

We get 4K of flash on the 8u2 for the firmware. Unfortunately, LUFA grew slightly and a arduino-usbserial compiled against the latest version would exceed those 4K. Therefore the second part of this patchset attempts to save some space and gets the firmware to fit into 4K again.

Finally, the third part regenerates the .hex files for the Uno and the Mega 2560 based on the other two changes.

All of this was tested on the Mega 2560 only.


facchinm:
[This repo](https://github.com/facchinm/Arduino_avrusb_firmware) contains updated firmwares that will not be merged in the main repo [ https://github.com/arduino/ArduinoCore-avr ] unless we factory flash the new firmwares. They need to be validated and, eventually, updated to latest LUFA too.

PR #25, opened 2011-03-28
bitclock parameter for programmers, speedier uploads

madworm:
Since it is possible to set the programmer in 'boards.txt', it makes sense to use optimal settings as well.

Currently uploading with avrdude + USBtinyISP is _awfully_ slow. The same is true for writing a bootloader.
Some programmers support the bitclock option "-B", let us use it then:

programmers.txt:

usbtinyisp.name=USBtinyISP
usbtinyisp.protocol=usbtiny
usbtinyisp.bitclock=1


ffissore:
You can add additional params to the `extra_params` entry in `programmers.txt`: for example `usbtinyisp.program.extra_params`

PR #27, opened 2011-04-08
Don't set TWSTO in tw_stop for slave transactions. This fixes non-working

wallclimber21:
Don't set TWSTO in tw_stop for slave transactions. This fixes non-working repeated-start sequences.


cmaglie:
This has been fixed with #66.

Thanks for your submission!

PR #34, opened 2011-06-24
Support building packets in w5100 packet memory.

karlp:
This is a fix for http://code.google.com/p/arduino/issues/detail?id=563

It is based on work in the UDP library, even though that library no longer
contains support for prebuffering packets.


facchinm:
Superseded by Ethernet library v2 ( https://github.com/arduino-libraries/Ethernet )

PR #49, opened 2011-11-11
Wire library support for TWI General Call

edbaafi:
This patch adds TWI (I2C) "General Call" capabilities to the Wire library

General call is like a broadcast that can be received by many listeners at the same same. This is useful when you don't know the address of the device you want to talk to or it can be used as a sort of interrupt over the TWI. While it was possible to enable general call from an Arduino user program simply by setting the TWGCE bit in the TWAR, it was not possible to determine which address the data came from. 

TwoWire::onReceive is overridden to add an optional second callback which will be called whenever data is received over the general call address.  

This patch addresses issue 120 (http://code.google.com/p/arduino/issues/detail?id=120) and its use is reasonably well documented in the example code.  

Thanks,
Ed

### DIFF
--- a/libraries/SoftwareSerial/SoftwareSerial.cpp
+++ b/libraries/SoftwareSerial/SoftwareSerial.cpp
@@ -360,6 +360,9 @@ void SoftwareSerial::setTX(uint8_t tx)
   //if rx and tx shares a pin, only set pin to output when writing
   if(!_sharedRxTx){
     pinMode(tx, OUTPUT);
+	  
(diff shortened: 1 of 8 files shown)


q2dg:
Could it be expanded to ARM-based boards?

facchinm:
Patchset rebased and PR ready to be opened here ( https://github.com/arduino/ArduinoCore-avr/compare/pr_49?expand=1 ).
If someone is still interested in "sponsoring" this feature, please open the PR and adopt the code :wink:

PR #53, opened 2011-12-07
Fix for issue 669

kmpm:


PaulStoffregen:
Fixed in Ethernet 2.0.0

PR #58, opened 2012-01-15
Update libraries/SoftwareSerial/SoftwareSerial.h

kasperskaarhoj:
I wrapped _SS_MAX_RX_BUFF in "#ifndef // #endif" so that it can be customised in applications needing a larger buffer (for instance when reading SMS messages from a GPRS module)

### DIFF
--- a/libraries/SoftwareSerial/SoftwareSerial.h
+++ b/libraries/SoftwareSerial/SoftwareSerial.h
@@ -39,7 +39,18 @@ http://arduiniana.org.
 * Definitions
 ******************************************************************************/
 
+/******************************************************************************
+* Definitions
+******************************************************************************/
+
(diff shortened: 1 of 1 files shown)


sandeepmistry:
I could not merge and rebase this via the command line, so I've manually made the change in: https://github.com/arduino/Arduino/commit/b37430b47377687263ce893c0a75f502968be95d.

flatsiedatsie:
For some reason I wasn't able to override the value.
```
#define _SS_MAX_RX_BUFF 255                         // The size of the software serial buffer
#include <SoftwareSerial.h>                       // A software library for serial communication. In this case we use it to talk to the GSM modem.

```
This still left me with a 64 byte buffer. It's a [mystery](https://forum.mysensors.org/topic/10977/expanding-the-size-of-an-existing-array).

PR #61, opened 2012-02-07
Altered Stepper library to use 5 phase 5 wire stepper motor.

rdodesigns:
5 phase motors can now be used by calling a constructor with five pins.

This update does not change the function calls of the previous versions, so no previous projects will be broken.

Additionally the comments were cleaned into a unified style and the entire library was made to fit 80 columns.


cmaglie:

PR #63, opened 2012-02-16
Three new boards and serial monitor window behavior change

Cougar:
I added three more popular Arduino compatible boards to the boards.txt. Existing boards (with different names) can be actually used for Nanode and FREAKDUINO-CHIBI  but Freaduino ADK needs a new entry. I do not know  what is the policy for adding new clones and you are free to reject these commits if needed.

Another change I made which is much more useful at least for me is the serial monitor behavior change. It is very inconvenient that after every code upload serial console will be closed which means that a) you loose all existing text in this window and b) you have to reopen serial monitor every time. I added pause/unpause for that where during upload serial port will be closed and all field at serial monitor window disabled. After that serial port will be reopened and all fields enabled again.


ffissore:
We can't add non arduino boards to the official arduino boards list. However, with 1.6.4, you can list additional boards into boards manager. See http://blog.arduino.cc/2015/05/07/arduino-ide-1-6-4-released/
As for serial monitor, a more recent proposal is available at #2742.
Thank you!

PR #65, opened 2012-03-06
Fix for issues #846 and #847

pbouchet:
Make `Stream::timedRead()` protected instead of private
http://code.google.com/p/arduino/issues/detail?id=846

Add a `Stream::find(char)` method
http://code.google.com/p/arduino/issues/detail?id=847


facchinm:
replaced by #3269

PR #67, opened 2012-03-20
Patch to wait for WIZnet W5100 reset

Alkarex:
Patch for http://arduino.cc/forum/index.php/topic,85342.msg677517.html#msg677517
Proposed by hardcore.
See W5100 Datasheet V1.2.4 page 64 available from:
http://www.wiznet.co.kr/Sub_Modules/en/product/product_detail.asp?pid=1011


cmaglie:
This is very old, but I think it has been fixed long ago in https://github.com/arduino-libraries/Ethernet.
BTW if this PR is still relevant, please re-submit on the above repository.

PR #68, opened 2012-03-23
Add support for avrdude -Pavrdoper

ajd4096:
https://code.google.com/p/arduino/issues/detail?id=864

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -24,9 +24,11 @@
 package processing.app;
 
 import processing.app.debug.AvrdudeUploader;
+import processing.app.debug.BasicUploader;
 import processing.app.debug.Compiler;
 import processing.app.debug.RunnerException;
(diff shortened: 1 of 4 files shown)


ffissore:
You can achieve the same result taking advantage of the current boards/platforms definitions (txt files)

Create a custom core with a custom board (for example: a copy of the [Arduino Leonardo](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/boards.txt#L260-L295) called "my leonardo") and [replace `avrdude`](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/boards.txt#L270) with a custom uploader. Create a new platform.txt (copying [arduino one](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/platform.txt) and change [command line params](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/platform.txt#L94)

Select "my leonardo" from the Boards list and upload a sketch

PR #69, opened 2012-03-30
Adding HttpClient

amcewen:
Here's the HttpClient, updated as per discussions on the Developers list (including using completed() rather than eof().


cmaglie:
@amcewen 
I see that the HttpClient library is now in the library manager, can we close this PR?

amcewen:
@cmaglie yes, that makes sense (unless you'd like to take it into the core, in which case I could work up a new version...)

cmaglie:
Thanks!

> unless you'd like to take it into the core, in which case I could work up a new version...

Having it into the core is a maintenance burden and adds delays in the update process.
Believe me, it's 10 times better to have it maintained independently from the upstream author, especially if he's active like you :-), especially now that the library manager makes it very simple to install/upgrade.

amcewen:
Hehe, I'm not sure I'm "active", more sporadic I think ;-)

PR #77, opened 2012-04-23
XBee automatic reset

rossbiro:
Here's some much better code to automatically reset FIOs attached to XBee's.  It's useable, but still has all the debugging code in.  I'd still like to add code to allow the user to add arbitrary remote commands.  Make commands other than AC do a commit. There is also a couple of occasional bugs where the XBees get out of sync and the easiest solution is to just exit the app and restart.

I don't really want this pulled.  I'd like comments on the UI.  Any important code changes (such as there is currently no provision to internationalize any of the strings) requested and general comments.

Don't worry about this until after the release.


matthijskooijman:
I'm looking through some older PRs and came across this one. It's not entirely clear to me what this adds, but looking at the code it seems that a lot of special casing is added for connection to or through an XBee or Zigbee device, which does not really look like it belongs in the Arduino IDE. I'm closing this PR since it's old, unclear and does not look like it fits with the IDE. If you disagree, feel free to comment and explain more.

PR #78, opened 2012-04-25
Adding the "dont_erase" option to board preferences

s-leroux:
Some AVR require to be erased before uploading a new program. But
the Arduino IDE always pass the '-D' (dont erase) option to avrdude.

This patch allow to disable this feature on a per-board basis.

### DIFF
--- a/app/src/processing/app/debug/AvrdudeUploader.java
+++ b/app/src/processing/app/debug/AvrdudeUploader.java
@@ -80,7 +80,9 @@ public class AvrdudeUploader extends Uploader  {
       "-P" + (Base.isWindows() ? "\\\\.\\" : "") + Preferences.get("serial.port"));
     commandDownloader.add(
       "-b" + Integer.parseInt(boardPreferences.get("upload.speed")));
-    commandDownloader.add("-D"); // don't erase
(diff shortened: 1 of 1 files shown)


ffissore:
When uploading via bootloader, no erase is performed any more, only when uploading via external programmer.

ffissore:
Erase is performed, I missed the end of the command line

ffissore:
You can achieve the same result taking advantage of the current boards/platforms definitions (txt files)

Create a custom core with a custom board (for example: a copy of the [Arduino Leonardo](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/boards.txt#L260-L295) called "my leonardo") and [replace `avrdude`](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/boards.txt#L270) with a custom uploader (for example "non_erasing_avrdude"). Create a new platform.txt (copying [arduino one](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/platform.txt) and change [command line params](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/platform.txt#L94)

Select "my leonardo" from the Boards list and upload a sketch

PR #87, opened 2012-05-28
Lcd oop extensions

manfredjonsson:
This Version of the Liquid Crystal Library can be inherited from other classes.
A possible usage is outlined in http://github.com/manfredjonsson/LiquidCrystalAddons

Because of the many changes, each commit contains one step of the improvements.

Commit 1: Fixes
- Remove unneeded variables.
- Make setCursor work for 16x2 and 16x4 displays.
- I consider LCD_ENTRYSHIFTINCREMENT etc to be a missnomer. Use ...ENABLE and 
  ...DISABLE instead. 
- Different variable names were used in definition and declaration of begin().
  Use the name "row" instead of "line" in all places to have a consistent 
  wording.
- Remove the begin() call from init() because some transport may not be 
  initialized in a global constructor (Some versions of the Wire (I2C) 
  library don't work when their begin() method is called inside a constructor 
  of a global object).

Commit 2: Minimal set of changes for inheritance
- Move functions and variables from private to protected.
- Add the virtual keyword to all functions which may need an override in an
  derived class.

Commit 3: Make it smaller
- Use the "-1" trick from rw for the data lines too. This enables the use of 
  one writebits() function instead of two. Pay attention to the fact that 
  pinMode is still in writebits(), this slows down each write to the LCD but 
  code is smaller compared to an additional initialization loop.
- Use the 4 upper bits of a byte in writebits() instead of the 4 lower bits 
  for 4 bit writes. Leads to an almost identical initialization for 4 and 8 
  bit mode. 
- Use the same code in 4 and 8 bit mode to force display to 8 bit mode and use 
  a loop instead of three distinct calls to writebits().

### DIFF
--- a/libraries/LiquidCrystal/LiquidCrystal.cpp
+++ b/libraries/LiquidCrystal/LiquidCrystal.cpp
@@ -44,15 +44,15 @@ LiquidCrystal::LiquidCrystal(uint8_t rs, uint8_t enable,
 }
 
 LiquidCrystal::LiquidCrystal(uint8_t rs, uint8_t rw, uint8_t enable,
-			     uint8_t d0, uint8_t d1, uint8_t d2, uint8_t d3)
+			     uint8_t d4, uint8_t d5, uint8_t d6, uint8_t d7)
 {
-  init(1, rs, rw, enable, d0, d1, d2, d3, 0, 0, 0, 0);
+  init(1, rs, rw, enable, 255, 255, 255, 255, d4, d5, d6, d7);
 }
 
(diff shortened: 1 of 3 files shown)


agdl:
This issue was moved to arduino-libraries/LiquidCrystal#11

agdl:
Sorry closed by mistake

facchinm:
Patches moved to https://github.com/arduino-libraries/LiquidCrystal/tree/pr_87 (mostly for historical reference)

PR #108, opened 2012-08-12
Fix operator new

kibergus:
Implementation of operator new in arduino contradicts C++ standard.
If malloc returned NULL it MUST throw exception or cal std::terminate in case if exceptions are disabled.
If you want operator new to return NULL just use non throw version of it which is not provided by arduino. Beside that, there should be operator new[] and placement variants of the operators.

There is correct realization in uclibc++ and I suggest using it instead of current implementation. To achieve it gnu ABI  specific functions must be moved to separate header (see https://github.com/arduino/Arduino/pull/107). After that new.h and new.cpp must be deleted and uclibc++ should be used instead.

If you have strong objections against uclibc++ but are still concerned about quality of the library, I can separate implementation of operator new from uclibc++ and provide patches for integrating it to arduino. But this is not a right way because it would tear standard library implementation to peaces.


facchinm:
If still interesting, I ported the PR to https://github.com/arduino/ArduinoCore-avr . If someone wants to adopt it, please open a PR using this link https://github.com/arduino/ArduinoCore-avr/compare/pr_108?expand=1 , thanks.

matthijskooijman:
@facchinm, I think the changes proposed in comments are still interesting, not so much the PR itself. I've opened an issue in the AVR repo about this to keep track.

PR #119, opened 2012-10-09
Stream - virtual copy constructor added

timschneider:
I've added a virtual copy constructor to Stream.h / .cpp, to give the user the ability to pass derived classes as reference. 

Tim


soundanalogous:
This issue should be closed. This was fixed via a different means in Firmata v2.3.6.

cmaglie:
@soundanalogous 
thanks for the heads up!

PR #121, opened 2012-10-19
Make SD library board agnostic

fakufaku:
Changed Sd2PinMap.h to use the pin mapping defined in the variant header pins_arduino.h (instead of redefining everything).

This small change makes the SD library compatible with any board that has a variant defined.

The library was tested successfully for both read and write using both the hardware SPI and the soft SPI.

PS.
This pull request replaces pull request [#117](https://github.com/arduino/Arduino/pull/117). The content is the same, but now neatly grouped in a single commit.

### DIFF
--- a/libraries/SD/utility/Sd2PinMap.h
+++ b/libraries/SD/utility/Sd2PinMap.h
@@ -22,308 +22,36 @@
 #define Sd2PinMap_h
 #include <avr/io.h>
 
-//------------------------------------------------------------------------------
-/** struct for mapping digital pins */
-struct pin_map_t {
-  volatile uint8_t* ddr;
-  volatile uint8_t* pin;
-  volatile uint8_t* port;
-  uint8_t bit;
-};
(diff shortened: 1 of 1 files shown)


matthijskooijman:
This commit looks good to me. I haven't checked it thoroughly or tested it, but the diff looks fine.

cmaglie:
Rebased and merged here: 05a9750747fcc3d5218acba8ae6a4fdacc6d017a

@fakufaku, thank you!

cmaglie:
Ok, it seems that we called victory too fast: as @PaulStoffregen pointed out (https://github.com/arduino/Arduino/commit/05a9750747fcc3d5218acba8ae6a4fdacc6d017a#commitcomment-15657902) this patch has a big performance impact, so I'm reverting for now, until we find a better way to improve optimization of pin-mapping macros (at least when the argument is a constant known at compile-time).

PS: @JasonSheng I've merged your patch in the meantime, sorry for the confusion!

PR #126, opened 2012-11-02
Fixdefines

iMartyn:
Multiple locations in the source are #if defined(blah) || blah2 || blah3 instead of #if defined(blah) || defined(blah2) ||  defined(blah3)

This branch fixes a few of them.


cmaglie:

PR #1118, opened 2012-11-17
Patch SD library to allow re-initialization after power down

fakufaku:
This is a patch of the SD library to allow the SD card to be re-initialized after having been powered off.

This patch fixes issue [1103](http://code.google.com/p/arduino/issues/detail?id=1103&start=400).

The problem was that the root directory is still marked as open and thus its initialization fails. The fix changes SD.init to first check if the root dir is open, and if so, closes it.

### DIFF
--- a/libraries/SD/SD.cpp
+++ b/libraries/SD/SD.cpp
@@ -340,6 +340,13 @@ boolean SDClass::begin(uint8_t csPin) {
     Return true if initialization succeeds, false otherwise.
 
    */
+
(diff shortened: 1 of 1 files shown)


facchinm:
@sandeepmistry since you recently updated SD library, do you think this might still be interesting?

sandeepmistry:
@facchinm yes!

@fakufaku it would be great to have a sketch that can reproduce the issue.

PR #1179, opened 2012-12-19
Add rts dtr and baudrate public functions in CDC Serial_ class

ynsta:
This permit to use the CDC serial link like a modem. For example configuration
mode while DTR is low and data mode otherwise.

Signed-off-by: Stany MARCEL <stanypub@gmail.com>


cmaglie:
Fixed with 96f7b966bccd8b09acdd53a977183368983fb8e1

PR #1211, opened 2013-01-11
Portable Arduino

sgk:
Use Arduino on USB flash stick without affecting anything on your PC.

To use this feature, simply make a folder named "portable" in the application
root folder where there are "hardware", "examples", "lib" and others.
In this mode, file association setting is disabled in the preferences dialog.

The sketch book folder is inside the "portable" folder. This may be confusing
especially for Mac OSX "Arduino.app" folder.

explorer

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -118,6 +118,11 @@ public class Base {
   static public void main(String args[]) throws Exception {
     initPlatform();
 
+    // Portable folder
+    portableFolder = getContentFile("portable");
+    if (!portableFolder.exists())
+      portableFolder = null;
+
     // run static initialization that grabs all the prefs
     Preferences.init(null);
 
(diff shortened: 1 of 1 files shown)


WestfW:
I'd like to see the restriction on changing the sketchbook location removed, so that (for example) I can have all the executables on a flash drive or other portable location, and the sketchbook on a cloud directory like onedrive/dropbox/etc.
The choice to use a portable app is the user's (possibly restricted by policy), and the user should still be able to reconfigure anything that IS allowed by policies.

aethaniel:
Hi, great addition!
Do you think it should be proposed/introduced to http://portableapps.com/apps also (even if Windows only)?

ffissore:
@WestfW good advice. Please open a separate issue
@aethaniel we don't plan to do that on our own. portableapps is free to do it however

Testato:
actually if you use "portable" folder is it not possible use a sketchbook folder outside the same disk (on Windows, if you have Arduino+portale folder on C:\ you cannot select a sketchbook on D:\)

And, if you are agree, i propose to insert the "portable" folder inside the Arduino .zip download by default. So the people that download the .zip automatically have this option. If a people download the .zip version he think that all the software remain inside the unzipped folder

PR #1221, opened 2013-01-13
HardwareSerial modifications

agalliazzo:
Added the possibility to configure the hardware serial port for 9 bit comunication and added an event (processed like SerialEvent) for a txfinished interrupt (usefull if you're using the serial port with a rs485 transceiver to drive RE/DE line).
I've tested the modifications with an arduino mega2560 and a pic microcontroller as a receiver with Serial1.

for use the txFinished event you should have a function like this:

void transmissionFinishedEvent1(){
   doSomething...
}

### DIFF
--- a/hardware/arduino/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/cores/arduino/HardwareSerial.cpp
@@ -15,7 +15,7 @@
   You should have received a copy of the GNU Lesser General Public
   License along with this library; if not, write to the Free Software
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
-  
+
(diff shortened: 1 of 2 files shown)


CrabbyPete:
What happened with this?

Bouni:
I modified the Hardware serial for 1.0.X and 1.5.X and submitted pull requests for both:

https://github.com/arduino/Arduino/pull/2291
https://github.com/arduino/Arduino/pull/2292

But nothing happened since then, which is a little disappointing :(

If you need 9-bit serial for your projects, you can clone my repo and use that:

https://github.com/Bouni/Arduino/tree/ide-1.5.x-hardware-serial-9-bit

facchinm:
Superseded by arduino/Arduino#2291

PR #1240, opened 2013-01-21
Made Ethernet library non Blocking

DavyLandman:
Hi Guys,

I've extended the `Ethernet` library to make it possible to work with it in a non blocking way, which makes it easier to combine the library with handeling other task (handeling buttons, updating a screen etc) while a DHCP lease happens, or a packet is sent.

I've added non blocking variants for:
- `Dhcp::*`
- `Dns::getHostbyName`
- `Ethernet::begin (DHCP version)`
- `Ethernet::maintain`
- `EthernetClient::connect(*)`

I welcome comments on the api-style, not really glad with all the names, but what would you guys think? I also need to think about a non blocking send, since it now loops until enough memory is available, I could make a variant which actually returns the bytes it could send, but I still have to think if this is the best way to do this.

There are 2 commits in this pull request, the Ethernet library was a mix of 3 different indentation styles, so I looked at the other parts and made commit a2ca62d to first only fix the formatting of the code. Then there is a new commit which actually changes the code, commit 2ac4e2f . This is actually an import from the work I started in DavyLandman/NBEthernet where I found out that the existing Ethernet library was already half non blocking.

So when reviewing, make sure to select just the correct commit (2ac4e2f), else in the github summary will claim I changed almost everything..

Impact on the size is quite minimal (for example `DnsWebClient` ): 

```
old: Binary sketch size: 13,474 bytes (of a 30,720 byte maximum)
new: Binary sketch size: 14,980 bytes (of a 30,720 byte maximum)
```

The increase is primarily due to keeping a blocking variant.

### DIFF
--- a/libraries/Ethernet/Dhcp.cpp
+++ b/libraries/Ethernet/Dhcp.cpp
@@ -11,121 +11,121 @@
 
 int DhcpClass::beginWithDHCP(uint8_t *mac, unsigned long timeout, unsigned long responseTimeout)
 {
-    _dhcpLeaseTime=0;
-    _dhcpT1=0;
-    _dhcpT2=0;
-    _lastCheck=0;
-    _timeout = timeout;
-    _responseTimeout = responseTimeout;
+	_dhcpLeaseTime=0;
+	_dhcpT1=0;
+	_dhcpT2=0;
+	_lastCheck=0;
+	_timeout = timeout;
+	_responseTimeout = responseTimeout;
 
-    // zero out _dhcpMacAddr
-    memset(_dhcpMacAddr, 0, 6); 
-    reset_DHCP_lease();
+	// zero out _dhcpMacAddr
+	memset(_dhcpMacAddr, 0, 6); 
+	reset_DHCP_lease();
 
-    memcpy((void*)_dhcpMacAddr, (void*)mac, 6);
-    _dhcp_state = STATE_DHCP_START;
-    return request_DHCP_lease();
+	memcpy((void*)_dhcpMacAddr, (void*)mac, 6);
+	_dhcp_state = STATE_DHCP_START;
+	return request_DHCP_lease();
 }
 
 void DhcpClass::reset_DHCP_lease(){
-    // zero out _dhcpSubnetMask, _dhcpGatewayIp, _dhcpLocalIp, _dhcpDhcpServerIp, _dhcpDnsServerIp
-    memset(_dhcpLocalIp, 0, 20);
+	// zero out _dhcpSubnetMask, _dhcpGatewayIp, _dhcpLocalIp, _dhcpDhcpServerIp, _dhcpDnsServerIp
+	memset(_dhcpLocalIp, 0, 20);
 }
 
 //return:0 on error, 1 if request is sent and response is received
 int DhcpClass::request_DHCP_lease(){
-    
-    uint8_t messageType = 0;
-  
-    
-  
-    // Pick an initial transaction ID
-    _dhcpTransactionId = random(1UL, 2000UL);
-    _dhcpInitialTransactionId = _dhcpTransactionId;
 
-    _dhcpUdpSocket.stop();
-    if (_dhcpUdpSocket.begin(DHCP_CLIENT_PORT) == 0)
-    {
-      // Couldn't get a socket
-      return 0;
-    }
-    
-    presend_DHCP();
-    
-    int result = 0;
-    
-    unsigned long startTime = millis();
-    
-    while(_dhcp_state != STATE_DHCP_LEASED)
-    {
-        if(_dhcp_state == STATE_DHCP_START)
-        {
-            _dhcpTransactionId++;
-            
-            send_DHCP_MESSAGE(DHCP_DISCOVER, ((millis() - startTime) / 1000));
-            _dhcp_state = STATE_DHCP_DISCOVER;
-        }
-        else if(_dhcp_state == STATE_DHCP_REREQUEST){
-            _dhcpTransactionId++;
-            send_DHCP_MESSAGE(DHCP_REQUEST, ((millis() - startTime)/1000));
-            _dhcp_state = STATE_DHCP_REQUEST;
-        }
-        else if(_dhcp_state == STATE_DHCP_DISCOVER)
-        {
-            uint32_t respId;
-            messageType = parseDHCPResponse(_responseTimeout, respId);
-            if(messageType == DHCP_OFFER)
-            {
-                // We'll use the transaction ID that the offer came with,
-                // rather than the one we were up to
-                _dhcpTransactionId = respId;
-                send_DHCP_MESSAGE(DHCP_REQUEST, ((millis() - startTime) / 1000));
-                _dhcp_state = STATE_DHCP_REQUEST;
-            }
-        }
-        else if(_dhcp_state == STATE_DHCP_REQUEST)
-        {
-            uint32_t respId;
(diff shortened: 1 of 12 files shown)


TrippyLighting:
Somehow from the provided example sketches I can't  figure out how to use this library non-blocking.
The behavior I would like is very well described in the PollingDHCP.pde that can be found in the line provide above http://gkaindl.com/software/arduino-ethernet/dhcp
The code below does exactly what I would want and I've used it a long time ago (Arduino 0023 ?). The Ethernet.begin returns immediately and one can continue with the sketch. Then in the loop the state of the DHCP is being polled.

[code]
void setup()
{
  Serial.begin(9600);
  
  // Initiate a DHCP session. The argument is the MAC (hardware) address that
  // you want your Ethernet shield to use. The second argument enables polling
  // mode, which means that this call will not block like in the
  // SynchronousDHCP example, but will return immediately.
  // Within your loop(), you can then poll the DHCP library for its status,
  // finding out its state, so that you can tell when a lease has been
  // obtained. You can even find out when the library is in the process of
  // renewing your lease.
  EthernetDHCP.begin(mac, 1);
}

void loop()
{
  static DhcpState prevState = DhcpStateNone;
  static unsigned long prevTime = 0;
  
  // poll() queries the DHCP library for its current state (all possible values
  // are shown in the switch statement below). This way, you can find out if a
  // lease has been obtained or is in the process of being renewed, without
  // blocking your sketch. Therefore, you could display an error message or
  // something if a lease cannot be obtained within reasonable time.
  // Also, poll() will actually run the DHCP module, just like maintain(), so
  // you should call either of these two methods at least once within your
  // loop() section, or you risk losing your DHCP lease when it expires!
  DhcpState state = EthernetDHCP.poll();

  if (prevState != state) {
    Serial.println();

    switch (state) {
      case DhcpStateDiscovering:
        Serial.print("Discovering servers.");
        break;
      case DhcpStateRequesting:
        Serial.print("Requesting lease.");
        break;
      case DhcpStateRenewing:
        Serial.print("Renewing lease.");
        break;
      case DhcpStateLeased: {
        Serial.println("Obtained lease!");

        // Since we're here, it means that we now have a DHCP lease, so we
        // print out some information.
        const byte* ipAddr = EthernetDHCP.ipAddress();
        const byte* gatewayAddr = EthernetDHCP.gatewayIpAddress();
        const byte* dnsAddr = EthernetDHCP.dnsIpAddress();

        Serial.print("My IP address is ");
        Serial.println(ip_to_str(ipAddr));

        Serial.print("Gateway IP address is ");
        Serial.println(ip_to_str(gatewayAddr));

        Serial.print("DNS IP address is ");
        Serial.println(ip_to_str(dnsAddr));

        Serial.println();
        
        break;
      }
    }
  } else if (state != DhcpStateLeased && millis() - prevTime > 300) {
     prevTime = millis();
     Serial.print('.'); 
  }

  prevState = state;
}

// Just a utility function to nicely format an IP address.
const char* ip_to_str(const uint8_t* ipAddr)
{
  static char buf[16];
  sprintf(buf, "%d.%d.%d.%d\0", ipAddr[0], ipAddr[1], ipAddr[2], ipAddr[3]);
  return buf;
}
[/code]

DavyLandman:
Hi @TrippyLighting ,

I made the pull-request over 2 years ago. Back then, it was working nicely, with perhaps a bit of polishing needed on the API choices.

However, perhaps @cmaglie should just close the pull-request, since I would be surprised if the past 2 years haven't seen enough evolution to make the diffs almost useless. 

The work on making it non-blocking was not trivial, and quite useful for me at the time, but I haven't done any Arduino development since then, and won't expect in the coming future.

TrippyLighting:
Hi @DavyLandman ,

Thanks for the reply. Actually there are not that many things different now compared to the standard Ethernet library. I believe your additions would be well worth to be integrated into the standard Arduino Ethernet library. Alas the Arduino teams not known for their agility in integrating changes.

I've been looking through the code over the weekend and I think I understand now how this works. I've started margin it with the Ethernet library that comes with Teensyduino, which is the Arduino IDE plugin that  goes along with the Teensy boards from PJRC ( http://pjrc.com/teensy/index.html ) .

I know what you mean with non-trivial. Last year I did a complete overhaul of the Arduino Bonjour library to rely only on UDP calls instead of talking directly to he W5100/5200 chip. That was a heck of an undertaking given my level of programming abilities. Works nice now though.

Thanks for your work. I'll let you know how things work out!

PaulStoffregen:
FWIW, I've been holding off UDP multicast and a number of other Ethernet changes, until Arduino 1.6.0.... which is now released.

Before 1.6.0, Teensyduino had to overwrite the Ethernet library.  I was reluctant to do too much with it, since that meant changing it for all other boards.  Now it'll be possible to add this stuff, without unexpectedly changing the copy used for other boards.

At the moment, I'm crazy busy finalizing version 1.21 and releasing the new LC board.  But soon I'm going to look at this and lots of other things that just weren't feasible to do before 1.6.0's better library management.

PaulStoffregen:
Long-term, I have some very ambitious plans for Ethernet to use the W5200 interrupt signal and support special "server" libraries.  A lot of that work probably won't ever be feasible on memory-limited AVR.  But on ARM Cortex-M chips, we ought to be able to create libraries that act like daemons on Linux or Unix systems, and do almost all their work using nested priority interrupts, and provide nice Arduino-style APIs to the sketch code for configuration and status.

TrippyLighting:
I am looking over the example sketches, but these seem to be unchanged from the originals and do not work.
"if (Ethernet.begin()==0)" does not compile (using a Teesny 3.0 Arm based board) because in Ethernet.h the return value is "void" and not "int".

Is there any example of how this is supposed to work. Do I need to call Ethernet.maintain()  in loop() and respond to it's return value ?

TrippyLighting:
After upgrading the NBEthernet library to use the new SPI transactions it works nicely.
If it needs to work with Teensyduino then the contents of the /utility folder also have to be replaced with the equivalent files from Teensyduino.

[code]

boolean _ethernetInitialized = 0;
setup();
{
Serial.begin(9600;
Ethernet.begin(myMac);
}

loop(){
		if(Ethernet.initialized()){
			if( Ethernet.maintain()==0){
				if (_ethernetInitialized == 0){
					Serial.print("My IP is ");
					Serial.println(Ethernet.localIP());
					_ethernetInitialized = 1;
				}
			}
		}
}

[/code]

bennigraf:
@TrippyLighting,

I understand you use the NBEthernet library with a teensy? I'm trying the same thing but I seem to install it wrong. Where did you place the files and what did you #include? 

I'd love to see non-blocking ethernet in Arduino core as well, btw!

Best, Benni.

DavyLandman:
fyi, the NBEthernet library was abandoned, I started with this pull request
since it would take too much cloning.

TrippyLighting:
@bennigraf 
You are correct. However, you canot just download the version provided here by Davy. Some stuff has change in the since Davy released it, for example the Ethernet library - particularly the one that ships with Teesnyduino - now uses SPI transactions.
I updated this NBEtherent library to work with SPI transactions and it works nicely. In my initial search I created a thread on the Teensy Forum:
https://forum.pjrc.com/threads/27980-Non-Blocking-Ethernet-DHCP-library-(-)?highlight=blocking+Ethernet
Please post there as a reminder for me and I'll post a zipped file containing the updated library.

alphaonex86:
No news?

DavyLandman:
I think I'm gonna close it, 8y is enough time to get the message 😉

PR #1244, opened 2013-01-24
Feature: implement DHCP options provider and parser

valentinalexeev:
I have implemented a feature to allow Ethernet library user to supply and read DHCP options. 

This functionality is very useful when you need to provide run-time configuration to the Arduino board. Since objective of DHCP is "host configuration" it is an ideal choice to use to pass parameters to the board. This commit also allows board to send additional options as part of the request. In my case, for example, it was necessary to provide additional Vendor Class Identifier (option 60) for the DHCP server to answer.

Here is a sample code:

``` c
// Define two functions to provide the DHCP options:

/** DHCP custom option parser.
 * At the moment will extract DNS domain (option 15) and store it separately).
 *
 * \warning The code must read() through the option even if it is unknown.
 * @param optionType DHCP option type as per standard.
 * @param dhcpUdpSocket Reference to UDP socket where to read the data from.
 */
void dhcpOptionParser(const uint8_t optionType, EthernetUDP *dhcpUdpSocket) {
    uint8_t opt_len = dhcpUdpSocket->read();

    if (optionType == 15) { // domain name
        // read the value with dhcpUdpSocket->read()
    } else {
        while (opt_len--) {
            dhcpUdpSocket->read();
        }        
    }
}

/** DHCP option provider.
 * Send Vendor Class Identifier (60) option as part of DHCP request.
 * @param messageType type of outgoing DHCP message
 * @param dhcpUdpSocket socket to write to
 */
void dhcpOptionProvider(const uint8_t messageType, EthernetUDP *dhcpUdpSocket) {
    uint8_t buffer[] = {
        dhcpClassIdentifier,
        0x08,
        'M','S','F','T',' ','5','.','0'
    };
    dhcpUdpSocket->write(buffer, 10);
}

//...

// Initialize the Ethernet library in setup() phase
void setup() {
    // ...
    Ethernet.begin(mac, (DhcpOptionParser *) &dhcpOptionParser, (DhcpOptionProvider *) &dhcpOptionProvider);
    // ...
}
```

Please refer to dhcp-options(5) http://linux.die.net/man/5/dhcp-options chapter "Defining New Options" for an example on how to provide custom options with isc-dhcp server. Other DHCP servers should have similar support.

### DIFF
--- a/libraries/Ethernet/Dhcp.cpp
+++ b/libraries/Ethernet/Dhcp.cpp
@@ -242,10 +242,16 @@ void DhcpClass::send_DHCP_MESSAGE(uint8_t messageType, uint16_t secondsElapsed)
     buffer[5] = domainName;
     buffer[6] = dhcpT1value;
     buffer[7] = dhcpT2value;
-    buffer[8] = endOption;
-    
+
     //put data in W5100 transmit buffer
-    _dhcpUdpSocket.write(buffer, 9);
+    _dhcpUdpSocket.write(buffer, 8);
+
+    if (_optionProvider != NULL) {
+        // If we have a custom option provider - pass the socket to it.
+        (*_optionProvider)(messageType, &_dhcpUdpSocket);
+    }
+
+    _dhcpUdpSocket.write(endOption);
 
     _dhcpUdpSocket.endPacket();
 }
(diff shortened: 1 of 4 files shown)


koenvdheuvel:
Is this ever going to be implemented. This feature would be really useful.

valentinalexeev:
@koenvdheuvel I’m not sure if my patch is still compatible with the current Arduino code base. It may require FF and rework in my repo first.

Feel free to resubmit PR if you have working code for the same and then this PR can be closed.

PR #1267, opened 2013-02-05
Bugfix for garbage data coming from serial port after reset. (Arduino Due)

cmaglie:
After reset garbage may show up from Serial port due to a firmware bug on the atmega16u2.

Here the relevant forum topic:
http://arduino.cc/forum/index.php/topic,134847

Here the updated .hex for the 16u2:
https://github.com/arduino/Arduino/blob/976f7588b7f68bd3c5fce52c15ccfe738589085a/hardware/arduino/sam/firmwares/atmega16u2/Arduino-DUE-usbserial-prod-firmware-2013-02-05.hex

Please check if the updated firmware fix the issues and report back here.
Thanks.
C


dinvlad:
I'm seeing this issue on 2 separate Mega's 2560 R3. More specifically, using

    void setup() {
      Serial.begin(115200);
      delay(300);
      randomSeed(analogRead(0));
      Serial.print(F("Test output "));
      Serial.println(random(255));
    }
If I remove the delay, after each upload it would show "Test output XX" from a previous connection:

    Test otput 34
    ðTest output 199
and more importantly, these numbers are generally different from those of the previous upload. If I set the baud rate in setup() to be different from what I use for the upload (e.g. 57600 vs 115200), "first output" still appears but in a completely garbled form:

    `f†æø~˜f~f~`x˜˜€Test output 59
suggesting that this trash output is a relic from the serial connection used for the upload. Could it be somehow discarded before the real desirable output? Of course, I could always discard anything that comes out of the board before sending the first command to it, but I was also hoping to log some initialization messages and use Serial 1..3 for communicating with third-party devices that may not react well to such output. I also updated the firmware using Flip utility and https://raw.githubusercontent.com/arduino/Arduino/master/hardware/arduino/avr/firmwares/atmegaxxu2/arduino-usbserial/Arduino-usbserial-atmega16u2-Mega2560-Rev3.hex

dinvlad:
It looks like other ports are having the same issue:

    HardwareSerial* serial[] = {&Serial, &Serial1, &Serial2, &Serial3};
    
    void setup() {
      randomSeed(analogRead(0));
      long c = random(255);
      
      for (int i = 0; i <= 3; i++) {
        serial[i]->begin(115200);
        serial[i]->print(F("Test output port "));
        serial[i]->print(i);
        serial[i]->print(F(": "));
        serial[i]->println(c);
      }
    }

Output from Serial0 (connected via USB to PC):

    Test output port 0: 241
    ðTest output port 0: 59

Output from Serial1 (connected via MAX235 and a third-party USB adapter to PC):

    Test output port 1: 241                       
    Test output port 1: 59

Does that imply that 16U2 is not responsible for such behavior? If I understand correctly, it only serves to translate TX0/RX0 to USB and programming Arduino.

PR #1272, opened 2013-02-08
1.5.x enhancements: serial monitor autostart and editor window size autosave

MartyMacGyver:
I've created two small enhancements targeting 1.5.x:

* Add preference serial.autostart to start serial monitor after a successful upload (Martin Falatic)

* Save editor.window.{width|height}.default values based on last editor window closed (Martin Falatic)

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -875,6 +875,10 @@ public class Base {
     // Close the running window, avoid window boogers with multiple sketches
     editor.internalCloseRunner();
 
+	int[] location = editor.getPlacement();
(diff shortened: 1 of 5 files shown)


ffissore:
Hi @MartyMacGyver and sorry for the late reply. Autostarting serial monitor was an old feature request we never liked. Today we merged a similar feature, proposed in #2180 and later in #2742. However with your PR you're also proposing a way to save window size: would you like to review this PR (or make a brand new one) in order to include such feature only?

ffissore:
Closing. On second though, it's better to start from a brand new PR rather than fixing this. Thanks!

PR #1276, opened 2013-02-11
Send byte array from serial-monitor

labsin:
Add the ability to send an array of bytes from the serial-monitor.
You write them using a comma separated list of hex, oct, dec or binary numbers.
It gives an error when the string is malformated.
In this case a new exception, ByteStringException, occures which is getting caught.

arduino 1
arduino 2
arduino 3
arduino 4
arduino 5
arduino 6

### DIFF
--- a/app/src/processing/app/ByteStringException.java
+++ b/app/src/processing/app/ByteStringException.java
@@ -1,0 +1,42 @@
+/*
+ * ByteStringException.java
+ * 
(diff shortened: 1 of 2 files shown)


ffissore:
Sorry for the so-late reply. Indeed, everything we add to the user interface makes the IDE harder to understand, for novices and students especially. Escaping chars might be a better solution, although, if you have such a requirement, you shouldn't really use the built in serial monitor. You should instead use some specialized app like putty or cutecom.

PR #1333, opened 2013-03-26
Update LiquidCrystal.cpp

newbie15:
in my case i have to change it. then my duino work properly with my 16x4 LCD


cmaglie:
Thanks

ffissore:
Please upgrade version in library.properties

Issue #1354, opened 2013-04-07
I2C/TWI Library: Buffer Overflow


neophob:
Tested on Arduino v1.0.3, Using Arduino Duemillanove/Seeeduino

I did some I2c testing between RPI and Arduino (http://neophob.com/2013/04/i2c-communication-between-a-rpi-and-a-arduino/) and observed that when a I2C master (RPI) sends data to an Arduino I2C slave device that is larger than the default buffer (32 bytes) the Arduino just freeze. After sending a i2c packet larger than the default buffer a i2cdetect scan shows no more devices on the bus.
The Arduino needs a reset after this event.

sandeepmistry:
@mofux @matthijskooijman are there any action items or can we close this issue?

seebs:
I am not entirely sure, but so far as I can tell, the patch in question (1) is actually needed, (2) is not present in arduino 1.8.1's copy of that library yet.

PR #1363, opened 2013-04-13
fix bug #1354: I2C/TWI Library: Buffer Overflow

neophob:
If we send i2c data larger than the arduino buffer (default: 32bytes) the arduino waits infinitely in a while loop (freeze). This fixes this.

### DIFF
--- a/libraries/Wire/utility/twi.c
+++ b/libraries/Wire/utility/twi.c
@@ -475,8 +475,14 @@ ISR(TWI_vect)
       break;
     case TW_SR_DATA_NACK:       // data received, returned nack
     case TW_SR_GCALL_DATA_NACK: // data received generally, returned nack
-      // nack back at master
-      twi_reply(0);
+      // We replied a nack, this happens when the buffer overflows.
+      // After this interrupt we don't get a TW_SR_STOP interrupt, so
+      // clean up now
+
+      // Discard the data, it is incomplete
+      twi_rxBufferIndex = 0;
+      // ack future responses and leave slave receiver state
+      twi_releaseBus();    
       break;
     
     // Slave Transmitter


netguy204:
I've encountered this bug as well and would love to see this merged.

sandeepmistry:
@cmaglie this looks good to merge to me. It will resolve https://github.com/arduino/Arduino/issues/486 as well.

I've tested with a Uno to Uno Wire setup by artificially lowering the slave RX buffer size:

```
diff --git a/hardware/arduino/avr/libraries/Wire/utility/twi.c b/hardware/arduino/avr/libraries/Wire/utility/twi.c
index b436e69..600103a 100644
--- a/hardware/arduino/avr/libraries/Wire/utility/twi.c
+++ b/hardware/arduino/avr/libraries/Wire/utility/twi.c
@@ -466,7 +466,7 @@ ISR(TWI_vect)
     case TW_SR_DATA_ACK:       // data received, returned ack
     case TW_SR_GCALL_DATA_ACK: // data received generally, returned ack
       // if there is still room in the rx buffer
-      if(twi_rxBufferIndex < TWI_BUFFER_LENGTH){
+      if(twi_rxBufferIndex < 10){
         // put byte in buffer and ack
         twi_rxBuffer[twi_rxBufferIndex++] = TWDR;
         twi_reply(1);
```

and the following sketches:

Master (modified version of ``master_writer`` example):

```c
#include <Wire.h>

void setup() {
  Serial.begin(9600);
  Wire.begin(); // join i2c bus (address optional for master)
}

void loop() {
  Wire.beginTransmission(8); // transmit to device #8
  Wire.write("fedcba9876543210");
  Serial.println(Wire.endTransmission(), HEX);    // stop transmitting

  delay(500);

  Wire.beginTransmission(8); // transmit to device #8
  Wire.write("0123456789");
  Serial.println(Wire.endTransmission(), HEX);

  delay(500);
}
```

Slave (modified version of ``slave_receiver`` example):

```c
#include <Wire.h>

void setup() {
  Wire.begin(8);                // join i2c bus with address #8
  Wire.onReceive(receiveEvent); // register event
  Serial.begin(9600);           // start serial for output
}

void loop() {
  delay(100);
}

// function that executes whenever data is received from master
// this function is registered as an event, see setup()
void receiveEvent(int howMany) {
  Serial.println(howMany);
  while (Wire.available()) {
    char c = Wire.read();
    Serial.print(c);
  }

  Serial.println();
}
```

Output of master sketch alternates between 0 and 3 (success and data send, NACK received).

Issue #1386, opened 2013-04-27
Leonardo keyboard API only allows ASCII input


weizenspreu:
The Arduino Leonardo keyboard API only allows ASCII input and then converts it to the actual key codes sent to the computer via USB. This is a huge problem because it fails for non-QWERTY systems (e.g. in Germany, France) and even for non-Windows systems (e.g. when it comes to special characters on a Mac).

Take this code as a small example. On a German computer this will actually output "qwerty".

    void setup() {
      Keyboard.begin();
    
      pinMode(2, INPUT);
      digitalWrite(2, HIGH);
    }
    
    void loop() {
      while (digitalRead(2) == HIGH) {
        delay(500);
      }
      delay(1000);
      
      Keyboard.print("qwertz");
    }

It would be great to have something like Keyboard.pressKeycode() and Keyboard.releaseKeycode() and actually do the conversions ourselves. That way we could ask the user what keyboard layout he/she has and do everything appropriately.

sandeepmistry:
Closing, looking at [Keyboard.h](https://github.com/arduino-libraries/Keyboard/blob/master/src/Keyboard.h) the following API's are present:

```c
  size_t press(uint8_t k);
  size_t release(uint8_t k);
  void releaseAll(void);
```

trahloff:
How would you modify these APIs in Keyboard.h to achieve a conversion?

sandeepmistry:
Re-opening this, on closer inspection raw key codes are still not supported currently.

nospam2000:
look at [pull request 1391 ](https://github.com/arduino/Arduino/pull/1391) it contains the code for the raw key codes

NicoHood:
We decided that the arduino IDE keeps the library simple while my HID Project will add all those new features. That was made with 1.6.6 or 1.6.7 (?) when the new usb core came out.
https://github.com/NicoHood/HID

I think you can close this issue as "wont fix"

PaulStoffregen:
@NicoHood - I see you've used my international keyboard layouts code in your project.  Please be aware the license for that code includes a requirement to have Teensy installed in the Tools > Boards menu.  It may be used on non-Teensy boards, but only when full Teensy support is available to the user in the menu.

NicoHood:
which international keyboard layouts code? `TeensyKeyboardAPI.h` etc? That is not working. And well if you say we cannot use it like that i will remove the code in the next release :/

limpkin:
[...]

PR #1391, opened 2013-04-28
Solution for: Leonardo keyboard API only allows ASCII input

weizenspreu:
Add pressKeycode(k, send) and releaseKeycode(k, send) methods to class Keyboard_ to fix issues with the QWERTY-only press(k) and release(k) methods. Additionally rewrite press(k) and release(k) to actually use the new *Keycode(k, send) methods and add some KEYCODE_* defines.

This fixes the following issue: https://github.com/arduino/Arduino/issues/1386

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_STORE
 app/bin/
 app/pde.jar
 build/macosx/work/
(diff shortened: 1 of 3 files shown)


facchinm:
Close with #1803

PR #1399, opened 2013-05-04
Ide 1.5.x Cleaned up a few warnings and fixed strict-aliasing

penguin359:
Cleaned up warnings about unsigned types, unused variables, and
declarations.

Explicitly declare char as signed for cases when compiling with
-funsigned-char.

Fixed strict-aliasing issues in IPAddress.h.

Fixed array overrun where EXTERNAL_NUM_INTERRUPTS was defined too small.


matthijskooijman:
As stated above, most of the things in this PR are already fixed, others are detable. In its current form, this PR is no longer useful. If some of the warnings it fixes still occur and you want to fix them, creating a new PR would be the best approach. I'm closing this one.

PR #1435, opened 2013-05-24
Update EthernetClient.h

panchazo:
Added function "free()" that returns available space in output buffer (TX). Useful to avoid writing more bytes than available.


PaulStoffregen:
Fixed in 2.0.0 with client.availableForWrite()

PR #1442, opened 2013-05-28
Only disable timer interrupt instead of all interrutps in millis()/micros()

chaveiro:
In wiring lib, usualy a global interrupt disabled is issued when not realy needed.
It's very bad for timed interrupt sensitive designs
More info at: http://forum.arduino.cc/index.php?topic=168586.0


matthijskooijman:
Looking through old PR's, I think this one is actually worth merging. It does need to be rebased and needs a better commit message, though.

facchinm:
If someone wants to adopt this PR, please open a new one using this link https://github.com/arduino/ArduinoCore-avr/compare/pr_1442?expand=1 , thanks!

PR #1447, opened 2013-06-04
newer avr-libc debricated some typedefs, this enables compatabillity

maqifrnswa:
many examples/libraries will fail to compile with newer avr-libc that no longer has the typedef for prog_uchar and many other variables. This flag enables compatibility.


cmaglie:
Users of SAM3X will never use `prog_xxx` definition without a reason, I guess we can keep it there for increased compatibility.

For the rest, as already explained, I'm closing as wontifx.

Related to #2866

PR #1459, opened 2013-06-09
Use of eeprom_update_byte instead of eeprom_write_byte

leomil72:
Using eeprom_update_byte the code won't write on the EEPROM cell if it already contains the value that the user wants to write into it.


Chris--A:
The EEPROM library now provides an update method. `EEPROM.write()` still uses `eeprom_write_byte`.

`EEPROM.put()` uses the `upadte()` method to write its data.

facchinm:
Fixed with bed78fd

PR #1531, opened 2013-08-03
Added Multicast UDP support

aallan:
Patch to add multicast UDP support to the current Arduino Ethernet library.

### DIFF
--- a/libraries/Ethernet/EthernetUdp.cpp
+++ b/libraries/Ethernet/EthernetUdp.cpp
@@ -216,3 +216,41 @@ void EthernetUDP::flush()
   }
 }
 
+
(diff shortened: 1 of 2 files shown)


roeften:
This patch works for me. I've written an example here:
http://blog.roeften.com/2015/03/multicast-communications-between.html

subnoize:
Well, after two years, what is the hold up with this?

cmaglie:
Rebased and merged.
I've changed the method name from `beginMulti` to `beginMulticast` to be more Arduino-style, hope this won't be a big concern.

Thanks you and sorry for the long delay.

subnoize:
Cool! Thank you!

PR #1613, opened 2013-10-09
Take 2: Added support for automatically showing the Serial Monitor window upon successful sketch upload

rmangino:
This fix was originally attempted via the following pull request but the Arduino team didn't take it (even though users have been asking for it for years).
https://github.com/rmangino/Arduino/commit/77c869d6b770e66ad47733db7f473255f08c9e01

This new pull request fixes an issue which didn't exist when I made the original pull request (over a year ago).

This support is controlled via a checkbox menu item in the Tools menu ("Auto Show Serial Monitor"). The feature is *disabled by default* and is tracked by a boolean "serial.auto_show_monitor_window" in Preferences.txt. When "auto show" is disabled the Serial Monitor window behavior is unchanged (i.e. if you don't enable "auto show" the IDE will work as it always has).

The serial monitor window is hidden when the sketch fails to compile.

NOTE: The IDE's default behavior has *not* changed. The only difference is that a user may now *explicitly* request that the serial monitor window be shown upon automatically upon successful upload.
Signed-off-by: Reed Mangino <reed@themanginos.com>

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -539,6 +539,9 @@ public class Editor extends JFrame implements RunnerListener {
     item.addActionListener(new ActionListener() {
         public void actionPerformed(ActionEvent e) {
           handleExport(false);
+          // Hide the serial monitor window. It will reopen at the appropriate time
+          // if "auto show" is enabled
+          serialMonitor.setVisible(false);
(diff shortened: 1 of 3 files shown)


ffissore:
A similar feature implemented with #2742

PR #1678, opened 2013-11-13
delayMicroseconds(): added support for 1Mhz, 12Mhz and 24Mhz

cano64:
1Mhz is a default clock speed on Atmega328, many users run stand alone microcontroller on the
internal 1Mhz clock to save battery power. Up until now delayMicroseconds() function wasn't taking this frequencies into an account defaulting to 8Mhz resulting incorrect timing. Patch includes support for 12 and 24 Mhz clock as well

### DIFF
--- a/hardware/arduino/avr/cores/arduino/wiring.c
+++ b/hardware/arduino/avr/cores/arduino/wiring.c
@@ -123,118 +123,114 @@ void delay(unsigned long ms)
 void delayMicroseconds(unsigned int us)
 {
   // call = 4 cycles + 2 to 4 cycles to init us(2 for constant delay, 4 for variable)
+  
 	// calling avrlib's delay_us() function with low values (e.g. 1 or
 	// 2 microseconds) gives delays longer than desired.
 	//delay_us(us);
 #if F_CPU >= 24000000L
 	// for the 24 MHz clock for the aventurous ones, trying to overclock
 
-	// for a one-microsecond delay, simply wait 6 cycles and return. The overhead
-	// of the function call yields a delay of exactly one microsecond.
-	__asm__ __volatile__ (
-		"nop" "\n\t"
-		"nop" "\n\t"
-		"nop" "\n\t"
-		"nop" "\n\t"
-		"nop" "\n\t"
-		"nop"); //just waiting 6 cycles
-	if (--us == 0)
-		return;
+	// zero delay fix
+  if (!us) return; //  = 3 cycles, (4 when true)
 
 	// the following loop takes a 1/6 of a microsecond (4 cycles)
 	// per iteration, so execute it six times for each microsecond of
 	// delay requested.
-	us *= 6; // x6 us
+	us *= 6; // x6 us, = 7 cycles
 
 	// account for the time taken in the preceeding commands.
-	us -= 2;
+	// we just burned 22 (24) cycles above, remove 5, (5*4=20)
+  // us is at least 6 so we can substract 5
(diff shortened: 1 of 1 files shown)


cano64:
Can we please merge it already? It's been open for a year and half. I spent a lot of time writing and fine tuning it and it has been working fine over the past year and half.

cmaglie:
Rebased and merged the fix.
Thank you, and sorry for the long delay!

RobTillaart:
Just checked the 1.6.4 code and you still cannot call delayMicroSeconds(0)  as the --us statement wraps around. Apparently nobody tested it ...

```C++
void delayMicroseconds(unsigned int us)
{
	// calling avrlib's delay_us() function with low values (e.g. 1 or
	// 2 microseconds) gives delays longer than desired.
	//delay_us(us);
#if F_CPU >= 20000000L
	// for the 20 MHz clock on rare Arduino boards

	// for a one-microsecond delay, simply wait 2 cycle and return. The overhead
	// of the function call yields a delay of exactly a one microsecond.
	__asm__ __volatile__ (
		"nop" "\n\t"
		"nop"); //just waiting 2 cycle
	if (--us == 0)
		return;

	// the following loop takes a 1/5 of a microsecond (4 cycles)
	// per iteration, so execute it five times for each microsecond of
	// delay requested.
	us = (us<<2) + us; // x5 us

	// account for the time taken in the preceeding commands.
	us -= 2;

#elif F_CPU >= 16000000L
	// for the 16 MHz clock on most Arduino boards

	// for a one-microsecond delay, simply return.  the overhead
	// of the function call yields a delay of approximately 1 1/8 us.
	if (--us == 0)
		return;

	// the following loop takes a quarter of a microsecond (4 cycles)
	// per iteration, so execute it four times for each microsecond of
	// delay requested.
	us <<= 2;

	// account for the time taken in the preceeding commands.
	us -= 2;
#else
	// for the 8 MHz internal clock on the ATmega168

	// for a one- or two-microsecond delay, simply return.  the overhead of
	// the function calls takes more than two microseconds.  can't just
	// subtract two, since us is unsigned; we'd overflow.
	if (--us == 0)
		return;
	if (--us == 0)
		return;

	// the following loop takes half of a microsecond (4 cycles)
	// per iteration, so execute it twice for each microsecond of
	// delay requested.
	us <<= 1;
    
	// partially compensate for the time taken by the preceeding commands.
	// we can't subtract any more than this or we'd overflow w/ small delays.
	us--;
#endif

	// busy wait
	__asm__ __volatile__ (
		"1: sbiw %0,1" "\n\t" // 2 cycles
		"brne 1b" : "=w" (us) : "0" (us) // 2 cycles
	);
}
```

cmaglie:
@RobTillaart 
the code you quoted is not the one shipped on 1.6.4.
I'm pretty sure that `delayMicroseconds(0)` works as expected, I've tested it personally.

https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/wiring.c#L131

RobTillaart:
@cmaglie 
I just downloaded - https://www.arduino.cc/en/Main/OldSoftwareReleases#previous -
the - Windows ZIP file for non admin install - option

Please check the wiring.c in that zip

cmaglie:
Right, the patch has been merged after IDE 1.6.4.

BTW IDE 1.6.5 should ship the latest or, alternatively, you can update the AVR core through Cores Manager.

RobTillaart:
Thanks, I'll check the 1.6.5. instead.

PR #1699, opened 2013-11-25
Let EthernetServer::write return the number of bytes written to any client

matthijskooijman:
Before, it would return the sum of the number of bytes written to all
clients, which makes it hard for a caller to find out the length of an
object printed (e.g., for alignment purposes). On the flipside, the
return value is now slightly less useful to detect connection errors,
but if really needed, something different should probably be implemented
anyway.

This change was discussed here:
http://comments.gmane.org/gmane.comp.hardware.arduino.devel/1928
http://forum.arduino.cc/index.php?topic=185835.0


agdl:
This issue was moved to arduino-libraries/Ethernet#35

agdl:
Sorry closed by mistake

PR #1701, opened 2013-11-26
Update stepper library: High-speed stepping mod and timer rollover fix

ekozlenko:
When using the stepper library with a 1.8 degrees per step motor, and at high angular speeds, the current Stepper library leads to really loud and jittery rotation. This is due to the fact that the timing is calculated in milliseconds, and the delay length between steps is only 2.5 milliseconds when trying to spin at 120 rpm. Since only integer math is performed, you end up actually bouncing between different step delays, and thus speeds, from step to step instead of giving the motor a constant input.  Which causes the motor to freak out.

Changing the library to calculate the step delays in micros() solves that problem for any speed you can reasonably demand from your stepper motor. The down side is that the micros() counter rolls over every hour or so, and any move you perform after that point will hang your code. Easy fix for that is to add an || micros() - this->last_step_time < 0 to the while loop if statement in Stepper.cpp.


cmaglie:

PR #1702, opened 2013-11-26
Update comments to match change in stepper.cpp

ekozlenko:
See stepper.cpp


cmaglie:

PR #1783, opened 2014-01-03
Refresh the available list of libraries whenever the Import Library menu is expanded

remoun:
This way the Arduino IDE doesn't need to be restarted whenever a user drops a new library directory into Documents/Arduino/libraries or the sketchbook's libraries folder.

Note: I'm not sure how slow the `rebuildImportMenu` call can get if there are tons of libraries, but I expect it won't be so bad, since it's just a few `stat()` calls. Perhaps we could guard it with a setting (preferably defaulting to true [rebuild]) if it proves too slow.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -950,6 +950,9 @@ public class Base {
   }
 
 
+  // NOTE(remoun): This method is called whenever the Sketch > Import Library
+  // menu is expanded, so make sure not to break that completely.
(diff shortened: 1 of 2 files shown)


ffissore:
Sorry for the late reply. A side effect of introducing Boards/Lib managers with IDE 1.6.2 is that both cores and libraries get refreshed as soon as one of the Manager gets closed. Plus, filesystem monitoring and background library parsing are expensive, on windows especially because of antivirus software.

PR #1800, opened 2014-01-11
Toggle function

ionut97:
Hi!
I added a toggle function which I thought will be useful.
The function has a default parameter (true) and the first is passed by reference.
I thought it will be useful to change the toggle a boolean variable when a button is pressed.

Exemple:
bool state = false;
int button = 5;

void setup(){
pinMode(button, INPUT);
}

void loop(){
toggle(state, digitalRead(button)); // start/stop button
   if(state){
      //action ...
   }
}

Thank you!

### DIFF
--- a/hardware/arduino/cores/arduino/Arduino.h
+++ b/hardware/arduino/cores/arduino/Arduino.h
@@ -95,12 +95,15 @@ typedef uint8_t byte;
 
 void init(void);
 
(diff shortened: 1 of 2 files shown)


sandeepmistry:
I agree with @matthijskooijman's comments above. Please bring this up on the mailing list as he has suggested.

PR #1803, opened 2014-01-13
Overhaul USB HID for Leonardo and Micro

obra:
This pull request includes the features from pull request 1488 and pull request 1391. I've somewhat overhauled the code to better match the surrounding code and a few features.

I've also added code examples for the less discoverable features.

If this pull request isn't "right" yet, I'd love feedback so I can improve it.


NicoHood:
Okay I had the chance to test the wakeup now. It seems it was a PC HW limitation, that my fujitsu Laptops only wakeup when pressing the power button. My Desktop now works fine with the u4 and u2.

So I think I am mostly done with anything. The Gamepad isnt working with all the reports and the ABS mouse causes some problems with windows drivers and changing devices but in general its working now. Maybe the gamepad ist just too much. Windows probably cant handle that many devices all together in one. Linux also has problems with more than one Gamepad.

So it would be nice to get some Feedback for this new usb core:
https://github.com/NicoHood/HID/tree/Official-Integration

NicoHood:
This may be closed now as well.

facchinm:
I'd close this issue as, from 1.6.6, USB-related enhancements can be implemented by libraries. 
All the code from this PR was squashed in CompleteHID library which can be easily recovered from git history or using @NicoHood [HID library] (https://github.com/NicoHood/HID/tree/dev_2_4)
Thanks everyone for taking part in this great effort!

Issue #1734, opened 2013-12-13
Allow library to include a configuration file located in the project folder


GDV0:
Writing a generic library that can be fine tuned for several projects seems difficult (impossible) without adding extra code and having a bad impact on th project memory footprint.
The idea is to have a library configuration file located in the project folder to allow this fine tuning.

PR #1808, opened 2014-01-16
Library configuration file management (issue #1734)

GDV0:
When an application uses a library supporting configuration:
- the library configuration file is stored in the sketch folder as a
project ressource.
- a sketch tab is created in Arduino IDE to edit the library
configuration file

This commit manages following actions:
- create configuration file tab when opening a sketch which uses a
Library configuration file
- copy the configuration file in the sketch folder and create a
configuration file tab when importing a library which uses a library
configuration file
- add the library configuration file (from sketch) as include file when
compiling the library only

Added as an example, Morse library supporting configuration file

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -73,7 +73,10 @@ public class Sketch {
 
   /** code folder location for this sketch (may not exist yet) */
   private File codeFolder;
-
+  
+  /** configuration folder location for this sketch (may not exist yet) */
+  private File configFolder;
+    
   private SketchCode current;
   private int currentIndex;
   /**
@@ -160,15 +163,24 @@ public class Sketch {
   protected void load() throws IOException {
     codeFolder = new File(folder, "code");
     dataFolder = new File(folder, "data");
+    configFolder = new File(folder, "configuration");
 
     // get list of files in the sketch folder
     String list[] = folder.list();
+    // get list of files in the library configuration folder
+    String list1[] = configFolder.list();
 
     // reset these because load() may be called after an
     // external editor event. (fix for 0099)
     codeCount = 0;
-
-    code = new SketchCode[list.length];
+    
+    // Add files from library configuration folder if exists
+    if (configFolder.exists()) {
+      code = new SketchCode[list.length+list1.length];
+    }
+    else {
+      code = new SketchCode[list.length];    
+    }
 
     String[] extensions = getExtensions();
 
@@ -203,6 +215,38 @@ public class Sketch {
     if (codeCount == 0)
       throw new IOException(_("No valid code files found"));
 
+    //Check library configuration folder to create tabs for library configuration files
+    if (configFolder.exists()) {
+      for (String filename : list1) {
+        // Ignoring the dot prefix files is especially important to avoid files
+        // with the ._ prefix on Mac OS X. (You'll see this with Mac files on
+        // non-HFS drives, i.e. a thumb drive formatted FAT32.)
+        if (filename.startsWith(".")) continue;
+  
+        // Don't let some wacko name a directory blah.pde or bling.java.
+        if (new File(configFolder, filename).isDirectory()) continue;
+  
+        // figure out the name without any extension
+        String base = filename;
+        // now strip off the .pde and .java extensions
+        for (String extension : extensions) {
+          if (base.toLowerCase().endsWith("." + extension)) {
+            base = base.substring(0, base.length() - (extension.length() + 1));
+  
+            // Don't allow people to use files with invalid names, since on load,
+            // it would be otherwise possible to sneak in nasty filenames. [0116]
+            if (Sketch.isSanitaryName(base)) {
+              code[codeCount++] = new SketchCode(new File(configFolder, filename), extension);
+              // Don't forget to declare it as configuration file
+              code[codeCount-1].setConfig(true);
+            } else {
+              editor.console.message(I18n.format("File name {0} is invalid: ignored", filename), true, false);
+            }
+          }
+        }
+      }
+    }
+    
     // Remove any code that wasn't proper
     code = (SketchCode[]) PApplet.subset(code, 0, codeCount);
 
@@ -878,7 +922,14 @@ public class Sketch {
 
     // now make a fresh copy of the folder
     newFolder.mkdirs();
-
+    
+    // If exists, make a copy of the configuration folder
+    File newConfigFolder= new File(newFolder, "configuration");
+    if (configFolder.exists()) {
+      Base.copyDir(configFolder, newConfigFolder);
+    }
+    
+    
     // grab the contents of the current tab before saving
     // first get the contents of the editor text area
     if (current.isModified()) {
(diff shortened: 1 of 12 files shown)


ricardojlrufino:
+1

ffissore:
Sorry for the late reply. Would you please rebase your PR on top of current master and provide the example Morse library as a zip file or separate repo, without including it in the PR? Otherwise we had to remove it right after merge

ffissore:
On second thought, we probably don't want such a feature at the moment. Having a separate configuration file may be confusing and lead library developers rely too much on `#define`s instead of providing a clearer API. In case of the Morse library, you could provide a richer constructor or (better) several overloaded constructors. For example
```arduino
Morse(int pin); //would set variables as in https://github.com/arduino/Arduino/pull/1808/files#diff-5b136a0d870efacf6b332258c2258b1d
Morse(int pin, uint8_t morseUnit); //would set just morseUnit and use default multipliers for the other values
Morse(int pin, uint8_t morseUnit, uint8_t morseDuration, uint8_t dashDuration, etc...); //for a complete configuration
```

cmaglie:
Isn't the compiler able to remove the unused methods from the final output? In your Morse library I guess that you can remove the define `LIB_MORSE_ALPHABET_AVAILABLE` and leave the methods always defined.

tzapu:
Will this be added in 1.6.5? planned release date for 1.6.5?

Cheers

svatoun:
Well, the reasoning about clean API is good - but for example now, the lack of #defines made me to fork the standard Keypad library, since I'm memory constrained, and the Keypad does not use memory effectively at all. #define could allow to limit size of the object to the actually used number of switches (16 instead of 160), reducing memory consumption about ~60 bytes. Alternatives are quite bad: using new operator (= malloc library = +200bytes for sketch, fragmentation) or user-supplied buffers (error prone). Or templatized Keypad - very ugly and not understandable code.

siara-cc:
I strongly recommend re-opening this discussion, as I am facing this issue with my Sqlite arduino library, which is only a wrapper to the original sqlite code.

As a library developer, I wouldn't want to be the one to decide which sqlite feature should be included or omitted.  A `per sketch` configuration would allow the user of my library to turn on/off whatever feature he wants.

I am pasting below my `config_ext.h` (which I am having to supply with the library now) to illustrate the issue.

```
#define YYSTACKDEPTH                        20
#define SQLITE_SMALL_STACK                   1
#define SQLITE_DEFAULT_PAGE_SIZE          4096
#define SQLITE_MAX_EXPR_DEPTH                0

#undef SQLITE_OMIT_ALTERTABLE
#undef SQLITE_OMIT_ANALYZE
#undef SQLITE_OMIT_ATTACH
#define SQLITE_OMIT_AUTHORIZATION            1
#undef SQLITE_OMIT_AUTOINCREMENT
#define SQLITE_OMIT_AUTOINIT                 1
#define SQLITE_OMIT_AUTOMATIC_INDEX          1
#define SQLITE_OMIT_AUTORESET                1
#define SQLITE_OMIT_AUTOVACUUM               1
#undef SQLITE_OMIT_BETWEEN_OPTIMIZATION
#define SQLITE_OMIT_BLOB_LITERAL             1
#define SQLITE_OMIT_BTREECOUNT               1
#define SQLITE_OMIT_BUILTIN_TEST             1
#define SQLITE_OMIT_CAST                     1
#define SQLITE_OMIT_CHECK                    1
#define SQLITE_OMIT_COMPILEOPTION_DIAGS      1
#define SQLITE_OMIT_COMPOUND_SELECT          1
#define SQLITE_OMIT_CONFLICT_CLAUSE          1
#undef SQLITE_OMIT_CTE
#define SQLITE_OMIT_DECLTYPE                 1
#define SQLITE_OMIT_DEPRECATED               1
#undef SQLITE_OMIT_DISKIO
#define SQLITE_OMIT_EXPLAIN                  1
#define SQLITE_OMIT_FLAG_PRAGMAS             1
#define SQLITE_OMIT_FOREIGN_KEY              1
#define SQLITE_OMIT_GET_TABLE                1
#define SQLITE_OMIT_INCRBLOB                 1
#define SQLITE_OMIT_INTEGRITY_CHECK          1
#undef SQLITE_OMIT_LIKE_OPTIMIZATION
#undef SQLITE_OMIT_LOCALTIME
#define SQLITE_OMIT_LOOKASIDE                1
#undef SQLITE_OMIT_MEMORYDB
#undef SQLITE_OMIT_OR_OPTIMIZATION
#undef SQLITE_OMIT_PAGER_PRAGMAS
#define SQLITE_OMIT_PARSER_TRACE             1
#undef SQLITE_OMIT_PRAGMA
#define SQLITE_OMIT_PROGRESS_CALLBACK        1
#define SQLITE_OMIT_QUICKBALANCE             1
#undef SQLITE_OMIT_REINDEX
#define SQLITE_OMIT_SCHEMA_PRAGMAS           1
#define SQLITE_OMIT_SCHEMA_VERSION_PRAGMAS   1
#define SQLITE_OMIT_SHARED_CACHE             1
#define SQLITE_OMIT_TCL_VARIABLE             1
#define SQLITE_OMIT_TEMPDB                   1
#define SQLITE_OMIT_TRACE                    1
#undef SQLITE_OMIT_TRIGGER
#define SQLITE_OMIT_TRUNCATE_OPTIMIZATION    1
#define SQLITE_OMIT_UTF16                    1
#undef SQLITE_OMIT_VACUUM
#undef SQLITE_OMIT_VIEW
#undef SQLITE_OMIT_VIRTUALTABLE
#undef SQLITE_OMIT_WSD
#define SQLITE_OMIT_XFER_OPT                 1
#define SQLITE_PERFORMANCE_TRACE             1
#define SQLITE_OMIT_COMPLETE              1
#define SQLITE_OMIT_SUBQUERY              1
#define SQLITE_OMIT_DATETIME_FUNCS        1
#define SQLITE_OMIT_FLOATING_POINT        1
#define SQLITE_COUNTOFVIEW_OPTIMIZATION      0
```
@ffissore  @svatoun @cmaglie

ricardojlrufino:
I think it should be possible to create new header file (e.g. MyRunningMedian.h) in your project, containing just your changed definitions:

```
#ifndef MEDIAN_MAX_SIZE
#define MEDIAN_MAX_SIZE     30
#endif
```
Then, you need to pass this header file to preprocessor, to include it as a first #include for the RunningMedian. You can do this by using -include directive, see https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html.

E.g. in PlatformIO you can do it in platformio.ini this way:

```
[common]
build_flags = 
    -include "include/MyRunningMedian.h"
```


from: https://stackoverflow.com/questions/42151118/how-to-override-define-in-arduino-library-header

tobozo:
unfortunately this isn't a platformio problem and Arduino IDE has no such build_flags mechanism that can be used from the *project folder* itself

ricardojlrufino:
> unfortunately this isn't a platformio problem and Arduino IDE has no such build_flags mechanism that can be used from the project folder itself

sorry, i forgot to mention it was just for reference

Issue #1311, opened 2013-03-07
1.5.2: Due Wire.available() always returns number of bytes requested, not bytes successfully read


kevin-pololu:
From arduino-1.5.2\hardware\arduino\sam\libraries\Wire\Wire.cpp:

````c++
int TwoWire::available(void) {
	return rxBufferLength - rxBufferIndex;
}
````

`rxBufferLength` is set in `requestFrom()`:

````c++
uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint8_t sendStop) {
	if (quantity > BUFFER_LENGTH)
		quantity = BUFFER_LENGTH;

	// perform blocking read into buffer
	int readed = 0;
	TWI_StartRead(twi, address, 0, 0);
	do {
		// Stop condition must be set during the reception of last byte
		if (readed + 1 == quantity)
			TWI_SendSTOPCondition( twi);

		TWI_WaitByteReceived(twi, RECV_TIMEOUT); // *** Even if this times out...
		rxBuffer[readed++] = TWI_ReadByte(twi);  // *** this will still be incremented
	} while (readed < quantity);
	TWI_WaitTransferComplete(twi, RECV_TIMEOUT);

	// set rx buffer iterator vars
	rxBufferIndex = 0;
	rxBufferLength = readed;

	return readed;
}
````

Whether or not `TWI_WaitByteReceived()` times out, `readed` is still incremented. This means that `available()` will always return the number of bytes requested, whether or not the bytes were actually read.

facchinm:
should be solved with c922069

PR #1869, opened 2014-02-18
SAM Wire: fixed NACK/timeout behavior of requestFrom() and available()

kevin-pololu:
`readed` is no longer incremented in `requestFrom()` if `TWI_WaitByteReceived()` gets a NACK or times out. This corrects the behavior (return values) of `requestFrom()` and `available()` to match the Arduino reference; they should return the number of bytes actually available for `read()`, not the number of bytes requested, which was the previous behavior. Fixes arduino/Arduino#1311


facchinm:
Merged, thank you!

PR #1891, opened 2014-02-24
Allow the serial monitor to stay opened during upload, disabling it

avishorp:
Currently, the serial monitor window closes when starting program uploading. I find this behavior very annoying since the monitor window has to be opened over and over again after each upload. The attached patch fixes this issue. When upload starts, the window gets disabled. When the upload process ends the window gets re-enabled and become operational immediately (if it is visible). The user can also now open the serial monitor window while uploading is in progress.


ffissore:
Superseded by #2180

Issue #1918, opened 2014-03-09
Stream.h: Remove _startMillis


cider101:
The member _startMillis is unneccesarry and can be replace with a local variable in methods timedRead() and timedPeek()
This saves 4bytes per Stream instance!

```
timedRead() {
    int c = -1;
    unsigned long timout = millis()+_timeout;
    while(c < 0 && millis() < timout) {
      c = read();
    }
    return c;     // -1 indicates timeout
}
```

PR #1928, opened 2014-03-12
Replace_startMillis from Stream.h with local variables

cider101:
As discussed in Issue #1918 an implementation with a local variable in timedPeek()

I apologize for having new.cpp included in that pullrequest as well - i'm still strugeling with github ;) Nevertheless, just removed some unessarry semi-colons there to avoid compiler warnings


facchinm:
Closing the issue here; if someone wants to adopt it, please reopen on api repo using this link https://github.com/arduino/ArduinoCore-API/compare/pr_1928?expand=1

PR #1936, opened 2014-03-16
Strings without null-termination

matthijskooijman:
When working with the Arduino String class, I've found that I couldn't efficiently combine it with some external libraries that explicitely pass `char*` and length around, without nul-terminating their strings. This prompted me to modify  and expose the `concat (const char* cstr, unsigned int length)` method, add a new `String(const char* cstr, unsigned int length)` constructor. While I was going over the string class, I found some other minor improvements, which are included here.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -696,7 +696,7 @@ void String::remove(unsigned int index, unsigned int count){
 	if (count > len - index) { count = len - index; }
 	char *writeTo = buffer + index;
 	len = len - count;
-	strncpy(writeTo, buffer + index + count,len - index);
+	memmove(writeTo, buffer + index + count, len - index);
 	buffer[len] = 0;
 }
 
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -174,7 +174,7 @@ String & String::copy(const char *cstr, unsigned int length)
 		return *this;
 	}
 	len = length;
-	strcpy(buffer, cstr);
+	memcpy(buffer, cstr, length);
 	return *this;
 }
 
@@ -194,7 +194,7 @@ void String::move(String &rhs)
 {
 	if (buffer) {
 		if (capacity >= rhs.len) {
-			strcpy(buffer, rhs.buffer);
+			memcpy(buffer, rhs.buffer, rhs.len);
 			len = rhs.len;
 			rhs.len = 0;
 			return;
@@ -266,7 +266,7 @@ unsigned char String::concat(const char *cstr, unsigned int length)
 	if (!cstr) return 0;
 	if (length == 0) return 1;
 	if (!reserve(newlen)) return 0;
-	strcpy(buffer + len, cstr);
+	memcpy(buffer + len, cstr, length);
 	len = newlen;
 	return 1;
 }
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -289,49 +289,49 @@ unsigned char String::concat(unsigned char num)
 {
 	char buf[1 + 3 * sizeof(unsigned char)];
 	itoa(num, buf, 10);
-	return concat(buf, strlen(buf));
+	return concat(buf);
 }
 
 unsigned char String::concat(int num)
 {
 	char buf[2 + 3 * sizeof(int)];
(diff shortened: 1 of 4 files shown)


matthijskooijman:
@cmaglie, any chance you get merge this for the next release? I just ran into trouble again by not having a `concat(str, len)` function...

matthijskooijman:
I've just rebased this PR, so it is again current. I also realized this changed the AVR WString code, but not SAM, so I updated each commit to apply the same changes to both.

ffissore:
@ArduinoBot build this please

matthijskooijman:
I tested this PR against all examples in the repository, which all still work as before (78% even compiles to the same binary). Some sketches become bigger, some sketches became smaller (on average all sketches became 3 bytes bigger. Limiting to only the sketches that actually changed, so the sketches using String in some way, these are 14 bytes bigger on average).

All size increases seem to be caused because `String::concat(const char*, unsigned int)` now calls `memcpy` instead of `strcopy`, making it remember the `length` parameter, causing the compiler to store data on the stack instead of just in registers, which has a little overhead. All size decrease  is because `strcpy()` can be left out, because `String::concat(char)` is simplifed and because a bunch of `strlen()` calls are removed. Nothing out of the ordinary or unexpected stood out.

Chris--A:
The features here would combine well with my PR here: #3096

In the future, rather than relying on the len input ([diff](https://github.com/arduino/Arduino/pull/1936/files#diff-2001c85724849b6911ac5c96436d5e21R204)), `validate()` could be called to ensure the String internals reflect its contents as a null-terminated string, rather than what was copied directly in.

This would resolve problems if someone was to copy or concatenate a buffer which has trailing nulls.

```arduino
char buff[10] = {'d','a','t','a'};
String s;

//User could easily use:
s.concat( buff );

//But now has the option to use:
s.concat( buff, sizeof(buff) );
```

Whether it be a logical error (buffer isn't as full as expected), or intention (nulls mean something else), `validate()` will ensure `s.len` reflects:
- a null terminated string.
- what is output with Print members.
- and when used in other c-string function using `s.c_str();`, or memory functions requiring a length.

matthijskooijman:
> In the future, rather than relying on the len input (diff), validate() could be called to ensure the String internals reflect its contents as a null-terminated string, rather than what was copied directly in.

One of the goals of this PR is to allow a String to store embedded NUL characters as well (e.g. to store a binary packet in it). IIUC, your suggestion would exactly break that possibility.

Having a `validate()` function sounds reasonable, but I don't think the String class should be calling it automatically - the user should be in control when he wants to fix the length, I think.

matthijskooijman:
Rebased this PR again, and added one more commit to add a `String(const uint8_t *, unsigned int)` constructor.

Chris--A:
> Having a validate() function sounds reasonable, but I don't think the String class should be calling it automatically - the user should be in control when he wants to fix the length, I think.

I agree, it should be an optional.

With regards to this PR, as it doesn't seem to change any existing functionality, and only add features, this can have a benefit (saw the issue you referenced above).

> When working with the Arduino String class, I've found that I couldn't efficiently combine it with some external libraries that explicitely pass char* and length around

This seems to be a different problem to the issue. By combine, do you mean something like this:

```C++
String s = "Hello";
s += Obj.func();
```

I'm not sure I understand what difficulties you are having.

matthijskooijman:
> When working with the Arduino String class, I've found that I couldn't efficiently combine it with some external libraries that explicitely pass char* and length around

For example, consider a library that calls a callback function when data is received, passing in a (non-nul-terminated) `char*` buffer and a length. If I want to accumulate the data in a String object, now I have to allocate memory for a copy of the buffer with added nul-termination, and then concat that to the existing String. Instead, I would like to write something like:

```
String accum;
void onReceive(const char*buf, unsigned len) {
     accum.concat(buf, len);
}
```
Which should just copy the data into the String's buffer, without needing another intermediate copy.

facchinm:
@matthijskooijman do you think it still makes sense to port this PR to API repo?

matthijskooijman:
I think these changes are still useful. They probably need a bit of rebasing and cleaning up. I'm willing to have a look at resubmitting this, if you (or someone else) will review and merge it?

facchinm:
I can't assure anything about merging but for sure we'll review them (I think it's a good patch so the merge chances are high).

matthijskooijman:
> I can't assure anything about merging but for sure we'll review them (I think it's a good patch so the merge chances are high).

Yeah, that's what I meant. Thanks! Pullrequest created at https://github.com/arduino/ArduinoCore-API/pull/97, so I'm closing this one.

PR #1977, opened 2014-03-31
Fix for 32 bit processors.

Timmmm:
The time-outs do not ever time-out on 32 bit processors if t0 happens to be close to 65535. This means that if you try to open an SD card that isn't present your app will sometimes freeze entirely in card.init() rather than returning an error like it is supposed to.

### DIFF
--- a/libraries/SD/utility/Sd2Card.cpp
+++ b/libraries/SD/utility/Sd2Card.cpp
@@ -241,7 +241,7 @@ uint8_t Sd2Card::init(uint8_t sckRateID, uint8_t chipSelectPin) {
 
   // command to go idle in SPI mode
   while ((status_ = cardCommand(CMD0, 0)) != R1_IDLE_STATE) {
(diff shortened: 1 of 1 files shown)


facchinm:
Very interesting bug! 
AFAICS, the unexpected behaviour is in AVR side, not ARM.
If `t0` approaches 2^16, the `while` loop will last forever because we are checking if a number that can be at most 2^16 will ever become bigger than the max value it can assume +  a big constant (`SD_INIT_TIMEOUT` ).
So the strange thing is that this code works with 8bit AVR; I checked the generated ASM and no promotion is applied so when `(uint16_t)millis()` wraps the result of the check is TRUE (I believe this is a side effect and not the expected behaviour). 
However, on ARM with the patch applied only the intermediate value checked against `SD_INIT_TIMEOUT` is promoted, while the mainline version promotes both the operands before subtracting (leading to the wrong result).
We should take a look in all shared libraries to check if this happens somewhere else.
Merging soon

facchinm:
Merged as #2808

PR #1995, opened 2014-04-07
Enable immediate evaluation of a Serial rx

bluesign2k:
Added a call to a new Serial rx event function,
"serialEventImmediate()". It allows users the advanced option of
processing an incoming serial character at the time of the interrupt
rather than having to wait for the current loop() to finish.

While the new function should be used with caution as it's called within
an interrupt routine, it enables processing of high priority incoming
data. eg. an abort or e-stop command


sandeepmistry:
As discussed in https://github.com/arduino/Arduino/pull/3203#commitcomment-15000386 it has been proposed `serialEvent()` might be removed in the future. I don't think it's a good idea to introduce a new `serialEventImmediate()` feature.

If you think this is important still, please start a discussion on the [Arduino developer mailing list](https://groups.google.com/a/arduino.cc/forum/#!forum/developers).

PR #2006, opened 2014-04-13
Updated/added config options for all Due Serial

bluesign2k:
For USART based Serial (Serial1, Serial2, Serail3), expanded config
options to include 1.5 stop bits and mark & space parity modes. Also
added defines ready for 9-bit implementation - currently commented out
until serial read/write supports it.

For UART based Serial, added ability to set config. Due to UART module
limitations, mode is fixed with 8-bit, 1 stop but parity is selectable
(none, odd, even, mark, space).

In both classes, user configs are masked to prevent bad mode
configuration.


cmaglie:
Thanks!

Issue #1962, opened 2014-03-26
Serial.parseFloat doesn't parse leading decimal points


shardbearer:
If you send a string that starts with a decimal point to Serial.parseFloat, it does not include the decimal. For example, .034 becomes 34 while 0.034 works fine. Are there any plans to fix this?

sandeepmistry:
Resolved by #3337.

PR #2007, opened 2014-04-13
Fix for Issue #1962

iOSCowboy:
Pull request to solve issue #1962. 

Floating point numbers with a trailing **.** character would be rendered as integers. 
For example: .032 => will become 32. 

The new function ```peekNextDigitOrPoint``` takes into account this and is used in the function ```parseToFloat```. 

Let me know if any re-factor or modification is required.


Chris--A:
This PR is incomplete.

The changes require Stream.h to be updated also.

facchinm:
Superseded by #3337

PR #2010, opened 2014-04-14
Support the Atmel STK500 and JTAGICE3 for ISP

matthijskooijman:
This adds support for the STK500 development board.

### DIFF
--- a/hardware/arduino/avr/programmers.txt
+++ b/hardware/arduino/avr/programmers.txt
@@ -51,18 +51,6 @@ usbGemma.program.tool=avrdude
 usbGemma.program.extra_params=
 usbGemma.config.path={runtime.platform.path}/bootloaders/gemma/avrdude.conf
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I just added one more commit, that allows programming of slower parts (like brand new 328 chips).

cmaglie:
@matthijskooijman 
for JTAGICE3 it seems a bit overkill to have 3 entries just for that, I'd keep only the slow ISP mode that seems the most common for burning a bootloader, what do you think?

matthijskooijman:
Well, one of the reasons I bought a JTAGICE3, is to get faster uploads on my Pinoccio board (uploading and verifying a 128k sketch takes over a minute through the bootloader). ISP at max speed is fast, JTAG is even faster.

So, perhaps we need submenus for programmers? :-)

matthijskooijman:
I just rebased this PR so it applies cleanly again.

matthijskooijman:
Superseded by #5623.

PR #2031, opened 2014-04-24
One click compile all known boards

WesGilster:
Changes to compiler and boards.txt file to enable a single click compile. This is helpful for tool builders that need to distribute a simple program as a part of a larger distribution that supports the full range of Arduino devices.

### DIFF
--- a/app/src/processing/app/debug/Compiler.java
+++ b/app/src/processing/app/debug/Compiler.java
@@ -77,12 +77,12 @@ public class Compiler implements MessageConsumer {
 	  Map<String, String> boardPreferences = Base.getBoardPreferences();
 	  
 	  if (!Preferences.get("board").equals("buildall")) {
-		  return compileForBoard(sketch, buildPath, primaryClassName, verbose, boardPreferences);
+		  return compileForBoard(sketch, buildPath, primaryClassName, verbose, primaryClassName, boardPreferences);
 	  }
 	  
 	  boolean compileOk = true;
 	  for (Entry<String, Map<String, String>> board : Base.getTarget().getBoards().entrySet()) {
-		  if (!compileForBoard(sketch, buildPath, primaryClassName, verbose, board.getValue())) {
+		  if (!compileForBoard(sketch, buildPath, primaryClassName, verbose, board.getKey(), board.getValue())) {
 			  compileOk = false;
 		  }
 	  }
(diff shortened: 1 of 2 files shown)


ffissore:
Sorry for the late reply. If I got it right, this PR aims at easing integration with other tools, right? If so, does using the CLI suffice?

An additional "build all" board will make life harder to both students and teachers teaching Arduino. Proof is the confusion created to studends with osx, puzzled by the two ports listed under ports, one tty and one cu. We ended up filtering out one: see #2626 and a8b29d48a51fd633a9c1407581a8eb772141ac70

ffissore:
This the CLI man page: https://github.com/arduino/Arduino/blob/master/build/shared/manpage.adoc

WesGilster:
I don't really see how it complicates things given that everyone pretty 
much understands that you build for whatever board that is selected. 
Even if your dead set against the feature, you can always build and 
deploy the Arduino IDE binaries without the "build all board" entry in 
the boards.txt. Then all menu options look identical as they were before 
my patch and only those that want the feature can modify the boards.txt 
file manually to get it back.

To answer your other question, it does look like the CLI would work with 
a looping shell/batch script. On the other hand, if I went that 
direction, I'd probably just wrap the makefile script from within 
Eclipse and use that tool chain exclusively. It's pretty painful keeping 
projects in sync between the two IDEs anyway and the "build all" menu 
option was the only real efficiency that the Arduino IDE had over 
Eclipse(for me that is).

I haven't heard much interest on the "build all" topic; so, honestly I 
don't think you'd be affecting tons of people if you nixed it. It seems 
there aren't many developers looking to build their code for multiple 
boards, which I found kindof curious.

Thanks,
Wes G.

ffissore:
Sorry, I just don't think this feature fits into the IDE. By using the CLI you can get the same result and also finer control over which boards are actually selected as targets.

matthijskooijman:
@WesGilster, not sure if you saw this one already, but since my last message I've written [arduino-mass-builder](https://github.com/matthijskooijman/arduino-mass-builder), a python script that can build a sketch for a large number of boards using the IDE commandline interface. The primary goal of that script was to compare different builds and test code, but it might be useful for you too.

WesGilster:
Well yeah, that's pretty much perfect guys. That makes moving to Eclipse 
a little more difficult... The C++ code for my project has pretty much 
stabilized so I might not be able to test this for a few months, but it 
looks pretty good. It even looks like it'll keep each of the built hex 
files as well.

Thanks,
Wes G.

PR #2032, opened 2014-04-24
SoftwareSerial improvements, fixes and optimizations

matthijskooijman:
This pullrequest contains a bit of everything, for the SoftwareSerial library. The first couple of commits I still had lying around from last year, forgot to send a pullrequest for those earlier. They're mostly cleanups around the introduction of a `stopListening()` method, that can be made to disable RX of a SoftwareSerial instance.

The second half of the commits are a lot of optimizations and fixes, to make the SoftwareSerial code faster and more reliable, so it works again at higher baudrates. This is only partly succesful for RX, since at higher baudrates where bytes are sent back-to-back, the CPU spends most of the time in the interrupt handler. At higher baudrates, the interrupt overhead (cleanup and setup) becomes longer than than the available time (from halfway the stopbit to halfway the startbit) and reception starts to fail.

This problem is particularly evident when running at 8Mhz, or when using gcc 4.3.2 (which makes terrible register allocation choices). Using gcc 4.8.2 and running at 16Mhz, things work up to 115200 (though bytes are still regularly corrupted because of the millis() timer ISR delaying the SoftwareSerial interrupt). Using 1.5 or 2 stop bits on the other side will probably help here (though I haven't tested this, and there is no way to make SoftwareSerial do TX with longer stop bits.

There is probably some room for improvment still, but that would require the RX ISR / `recv` method to be handcoded in assembly. The upside of that would be that it's 100% deterministic, without depending on the compiler used, but it's a lot of work that I'm not going to invest right now (getting this far has cost me enough time already).

These commits also remove the delay lookup tables that were previously present, since the values can easily be calculated at runtime. This saves around 160 bytes of program memory.

In addition to making the code faster, it is also a lot smaller. Overall, a typical SoftwareSerial sketch on the Uno saves slightly over 400 bytes of program space (including the savings for the lookup table removal).

### DIFF
--- a/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
+++ b/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
@@ -178,6 +178,9 @@ inline void SoftwareSerial::tunedDelay(uint16_t delay) {
 // one and returns true if it replaces another 
 bool SoftwareSerial::listen()
 {
+  if (!_rx_delay_stopbit)
+    return false;
+
   if (active_object != this)
   {
     _buffer_overflow = false;
--- a/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
+++ b/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
@@ -242,15 +242,13 @@ void SoftwareSerial::recv()
     DebugPulse(_DEBUG_PIN2, 1);
 
     // Read each of the 8 bits
-    for (uint8_t i=0x1; i; i <<= 1)
+    for (uint8_t i=8; i > 0; --i)
     {
       tunedDelay(_rx_delay_intrabit);
+      d >>= 1;
       DebugPulse(_DEBUG_PIN2, 1);
-      uint8_t noti = ~i;
       if (rx_pin_read())
-        d |= i;
-      else // else clause added to ensure function timing is ~balanced
-        d &= noti;
+        d |= 0x80;
     }
 
     // skip the stop bit
--- a/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
+++ b/hardware/arduino/avr/libraries/SoftwareSerial/SoftwareSerial.cpp
@@ -259,11 +259,12 @@ void SoftwareSerial::recv()
       d = ~d;
 
     // if buffer full, set the overflow flag and return
-    if ((_receive_buffer_tail + 1) % _SS_MAX_RX_BUFF != _receive_buffer_head) 
+    uint8_t next = (_receive_buffer_tail + 1) % _SS_MAX_RX_BUFF;
+    if (next != _receive_buffer_head)
     {
       // save new data in buffer: tail points to where byte goes
       _receive_buffer[_receive_buffer_tail] = d; // save new byte
-      _receive_buffer_tail = (_receive_buffer_tail + 1) % _SS_MAX_RX_BUFF;
+      _receive_buffer_tail = next;
     } 
     else 
     {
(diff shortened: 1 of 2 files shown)


NicoHood:
It works, why does nobody merge this? Its a perfect patch!

matthijskooijman:
@NicoHood, seems I missed your comment on erronous initial bytes before. Can you explain your setup in a bit more detail so I can try to reproduce?

NicoHood:
Use an Arduino Uno, setup softserial and print the hello world in the setup() function. read from another inipendant arduino via hardware serial. Then reset the uno. on a reset it will always output the y. I dont know if this is a problem of the HW serial from the 2nd arduino or the softserial of the uno. It happens for 9600 and 115200 baud rate (at probably all others too).

matthijskooijman:
I had a go at reproducing this, with some success. Sometimes I received a 0x00 byte when pressing the reset button, sometimes 0xff (which is shown as ÿ in [ISO 8859-1](http://en.wikipedia.org/wiki/ISO/IEC_8859-1)) when releasing the reset button. Looking at the signal everything looks good, but during reset the serial pin is floating. Sometimes it would go low when pressing the reset button, which looks like a start bit followed by all zeroes, causing 0x00 to be received (which is technically incorrect on the receiving side, since the stop bit is missing). Sometimes it would very shortly go low when releasing the reset, which looks like a start bit followed by one bits, causing 0xff to be received.

So the bottom line is: During reset, pins are floating, so anything might happen. If this is a problem, add an external pullup to keep the pin high during reset (just tested this, works for me).

In other news, someone contacted me by mail asking if this PR was merged yet. He manually applied it, which solved his  softwareserial problems on his Yun.

matthijskooijman:
One more thing: the erratic reset stuff also happens without this PR applied, as can be expected.

NicoHood:
Does this also happens on HW Serial ports? As I remember it does, so its a general problem on any pin, not related to SoftSerial, more related to a general Serial/floating pin problem?

Yeah, merge it guys! He did a great job!

matthijskooijman:
Good point. I just tried hardware TX. I had to use an external pulldown to trigger the problem, since the pin stayed high. I did notice that I didn't see an 0x255 read, so no short low pulse when releasing reset. This got me thinking: The SoftwareSerial initialization actually switches to OUTPUT LOW for a short while before switching to OUTPUT HIGH, causing this 0x255 read. By switching initialization order, this can be prevented. I added one more commit which does this, so now you might be able to reset without any junk being read (but since floating pins are involved, you might still get junk of course)

I also rebased the PR on top of the current ide-1.5.x branch.

NicoHood:
Yeah that makes sense. Setting a pin to output results in a low signal, input in high (used that for 3.3v gamecube controller on 5v pins once).

I am really excited to merge this PR as you can see. It fixes all problems we had so far really good.

NicoHood:
Someone has to test it with 8mhz and then it can be merged, i'd say. Not sure if the original supported 8mhz though. I really want to see this in the official next update, since it is such a great fix.

NicoHood:
Results with 8MHZ:
Receiving works with up to 38400
Transmitting work up to 57600

Regarding to this, it is acceptable for 8MHZ
http://wormfood.net/avrbaudcalc.php

115200 (8mhz) for HW Serial works crapy, so why should SS do it at 8mhz :D

No other bugs found yet. As stated many times, this PR is ready to merge in my opinion.

facchinm:
My results with UNO:
up to 57600 full-duplex (with both tx and rx slightly buffered in software) with no error.

No bugs found so far

@matthijskooijman , these patches are outstanding and such a great speed improvement over the previous version!
Merging this PR asap

NicoHood:
Thanks!

quique123:
someone please help me get this patch so that I can test it on my Arduino IDE 1.6.6?  Im getting no data from a jpeg camera and Im thinking this might be the issue.

matthijskooijman:
@quique123, these changes are already included in 1.6.6, so they won't help you :-)

For getting help with your issue, I recommend posting on the forum, or asking on the #arduino IRC chat channel.

quique123:
Oh ok.  Thanks.

I have already posted in the forum.  It seems like nobody knows what is going on.  I've asked the manufacturer and the distributor and no one has been able to help me.

MV

PR #2063, opened 2014-05-09
Ide 1.5.x spi 16bit transfer

MadVikingGod:
Added functions to the Due SPI library to enable 16bit SPI transfers.

To use:

* Initialize as normal, then use SPI.setTransferWidth(pin, width).

```
void setup() {
    SPI.begin(4);
    SPI.setTransferWidth(4,SPI_WIDTH_16);
}
```
* Use SPI.transfer() as normal, but can use upto 16 bits.

```
SPI.transfer(4,0xAA55);
```

The LSBFIRST setting only works for 8 bits.
TODO: twiddle 9-16 bit words to be reversed for LSB transfers. This is made more difficult because of the right-justified contents of the transfer data format.

-MVG


cmaglie:
@MadVikingGod 
thanks for the patch but we are not going to merge this one because it's very SAM3X specific.
Implementing a function like `SPI.transfer16(..)` is more suited for this purpose, see #3961

PR #2080, opened 2014-05-16
Timer1c support and per-variant initialization function

matthijskooijman:
These are two fairly unrelated improvements, other than that I need them both for a sleep timing improvement in our Pinoccio project.

The first commit adds support for timer 1c, on hardware that supports it. The second commit allows a variant file to supply a custom initialization function, to allow for board-specific code to be included by third parties.

@cmaglie, I'd be thankful if you could have a look at these soon and merge them if they look ok to you? I need these changes merged before I can publish some of my own Pinoccio work. Thanks!

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Arduino.h
+++ b/hardware/arduino/avr/cores/arduino/Arduino.h
@@ -118,6 +118,7 @@ typedef uint8_t boolean;
 typedef uint8_t byte;
 
 void init(void);
+void initVariant(void);
 
 void pinMode(uint8_t, uint8_t);
 void digitalWrite(uint8_t, uint8_t);
--- a/hardware/arduino/avr/cores/arduino/main.cpp
+++ b/hardware/arduino/avr/cores/arduino/main.cpp
@@ -19,9 +19,16 @@
 
 #include <Arduino.h>
 
+// Weak empty variant initialization function.
+// May be redefined by variant files.
+void initVariant() __attribute__((weak));
+void initVariant() { }
+
 int main(void)
 {
 	init();
+
+	initVariant();
 
 #if defined(USBCON)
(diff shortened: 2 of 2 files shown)


NicoHood:
Is there any documentation on how to actually use it? How do I have to name the file and where to place it?

cmaglie:
Basically, if you define a function `void initVariant()` this function is called at init time before `void setup()`.
You can define this function anywhere but the intended place is inside your `variant/xxx` folder, typically in a file named `pins_arduino.c` that performs specific initialization required for the board (in addition to the `pins_arduino.h` that contains metadata about the board itself).

You can define ISR as usual, it's a C file that is compiled and linked like any other source file.

vicatcu:
Nico,

I don't know how much documentation there is or where to find it, but take
a look at the core for WildFire here:
http://github.com/wickeddevice/wildfire-arduino-core for a working example
of using this functionality. The variant files are in
/WickedDevice/avr/variants/wildfirev3

Cheers,
Vic

Victor Aprea // Wicked Device
victor.aprea@wickeddevice.com
T: @vicatcu <http://twitter.com/vicatcu/>

NicoHood:
Thank you so much!

Some results:
The cpp has to be placed in the same directory as the pins file (of course).
The cpp name doesnt matter.
You dont need a variants.h like in the example.
You dont need this initvariants function, you can also put other stuff in there, that gets compiled, like ISRs.

Thats my file for example: no_usb_isr.cpp
```cpp
#include <Arduino.h>
#include <pins_arduino.h>

// workaround for undefined USBCON has to be placed in every sketch
// otherwise the timings wont work correctly
ISR(USB_GEN_vect)
{
	UDINT = 0;
}
```

PR #2119, opened 2014-06-08
Fix the ATmega8U2 code to support 300 bps

supersat:
Fixes issue #542


NicoHood:
Still interested? you might want to try this:
https://github.com/NicoHood/HoodLoader2

facchinm:
I'm closing this PR due to the bootloader/firmware policy (no update if the generated binary is not going to be flash on production boards)

Issue #2147, opened 2014-06-28
boolean data type causes logic errors.


Chris--A:
It has come to my attention that the `boolean` type alias produces results differing to the C++ standard boolean type `bool`.

Consider this sketch:

``` C++
void setup() {
  Serial.begin(9600);
  Serial.println( true == (bool) 57 ? "true" : "false" );
  Serial.println( true == (boolean)57 ? "true" : "false" );
  Serial.println( true == bool( true ) ? "true" : "false" );
  Serial.println( bool( 1 | 2 | 4 | 8 ) == boolean( 1 | 2 | 4 | 8 ) ? "true" : "false" );
  Serial.println( true == false + 1 ? "true" : "false" );
  Serial.println( true + true == true ? "true" : "false" );
  Serial.println( bool( true + true ) == true ? "true" : "false" );
  Serial.println( boolean( true + true ) == true ? "true" : "false" );
}

void loop(){}
```

Using this sketch with `boolean` declared as a `typedef` to `uint8_t` ( which is what the current core does ) the output to the serial monitor is:

> true
> false
> true
> false
> true
> false
> true
> false

Whereas, after changing the core to allow a `boolean` to be declared as `bool`. The sketch produces this output:

> true
> true
> true
> true
> true
> false
> true
> true

It appears `boolean` has been implemented to provide boolean functionality with some C code in the Arduino core. I recommend changing the core so by the time `boolean` is seen by a sketch it is either a `typedef` or `#define` to the real `bool` type.

These are the modifications I propose be implemented in Arduino.h

``` C++
#ifdef __cplusplus
    typedef bool boolean;
#else
    typedef uint8_t boolean;
    #define false 0
    #define true !false
#endif
```

On a side note, the defines for `true` and `false` should be removed. Boolean's ( the real one ) are subject to integral promotion and convert to either 0 or 1. Also the standard forbids macros named the same as keywords. If the C code needs defines, use `TRUE` & `FALSE` instead.

I have created a pull request here:
https://github.com/arduino/Arduino/pull/2151

I also have a proof I'm in the middle of writing as to why the defines `true` and `false` need to be removed from the C++ side. I will add it soon.

PR #2151, opened 2014-06-30
Modified Arduino.h with changes to the boolean type

Chris--A:
As the Arduino specific boolean type is dangerous in C++, I have implemented a fix to solve the situation for both C and C++ There is also no need for stdbool, it also defines true and false, which is not needed or allowed in a C++ context.

This is in reference to the issue I posted here.

https://github.com/arduino/Arduino/issues/2147

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.h
+++ b/hardware/arduino/avr/cores/arduino/WString.h
@@ -62,6 +62,8 @@ public:
 	#ifdef __GXX_EXPERIMENTAL_CXX0X__
 	String(String &&rval);
 	String(StringSumHelper &&rval);
+	auto begin() -> const char* { return c_str(); }
+	auto end() -> const char* { return c_str() + length(); }	
 	#endif
 	explicit String(char c);
 	explicit String(unsigned char, unsigned char base=10);


cmaglie:
From:
http://www.cplusplus.com/reference/cstdbool/
http://www.c-faq.com/bool/booltype.html
http://stackoverflow.com/questions/1608318/is-bool-a-native-c-type

it seems that `_Bool` is not gcc specific but part of C99

> The purpose in C of this header (stdbool.h) is to add a bool type and the true and false values as macro definitions.
> In C++, which supports those directly, the header simply contains a macro that can be used to check if the type is supported:

So it's sufficient to include `stdbool.h` to have a proper `bool` type defined.
I'd like to leave all the boring definition of true/false etc. to stdbool.h and change only the following:

```diff
 #define bit(b) (1UL << (b))
 
-typedef uint8_t boolean;
+typedef bool boolean;
 typedef uint8_t byte;
 
 void init(void);
```

What do you think?

xxxajk:
That is exactly what I was saying :-)

On Tue, Jan 6, 2015 at 4:43 PM, Cristian Maglie <notifications@github.com>
wrote:

> From:
> http://www.cplusplus.com/reference/cstdbool/
> http://www.c-faq.com/bool/booltype.html
> http://stackoverflow.com/questions/1608318/is-bool-a-native-c-type
>
> it seems that _Bool is not gcc specific but part of C99
>
> The purpose in C of this header (stdbool.h) is to add a bool type and the
> true and false values as macro definitions.
> In C++, which supports those directly, the header simply contains a macro
> that can be used to check if the type is supported:
>
> So it's sufficient to include stdbool.h to have a proper bool type
> defined.
> I'd like to leave all the boring definition of true/false etc. to
> stdbool.h and change only the following:
>
>  #define bit(b) (1UL << (b))
> -typedef uint8_t boolean;+typedef bool boolean;
>  typedef uint8_t byte;
>
>  void init(void);
>
> What do you think?
>
> —
> <https://github.com/arduino/Arduino/pull/2151#issuecomment-68939162>.
>



-- 
Visit my github for awesome Arduino code @ https://github.com/xxxajk

Chris--A:
@cmaglie That is excellent. All that is important to me is correct semantics in C++.

PR #2175, opened 2014-07-10
Enable C++11 and C11

matthijskooijman:
Now that we have a branch with a more recent gcc version we can enable
C++11 support.

This enables some more powerful and useful C++ features, like variadic
templates, rvalue refs / move constructors (for more efficient code -
String class already supports these), constexpr (to allow e.g. function
calls as constant values), foreach-style loop, enumeration classes,
static assertions. See http://en.wikipedia.org/wiki/C%2B%2B11 for
details.

These features are mostly useful for more advanced users and library
authors, but there is a lot in there that can make a significant
difference to code efficiency or conciseness.

Enabling C++11 should be completely backwards compatible.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -43,7 +43,7 @@ String::String(const __FlashStringHelper *pstr)
 	*this = pstr;
 }
 
-#ifdef __GXX_EXPERIMENTAL_CXX0X__
+#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__)
 String::String(String &&rval)
 {
 	init();
@@ -189,7 +189,7 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length)
 	return *this;
 }
 
-#ifdef __GXX_EXPERIMENTAL_CXX0X__
+#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__)
 void String::move(String &rhs)
 {
 	if (buffer) {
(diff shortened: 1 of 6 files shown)


matthijskooijman:
@cmaglie, could you put this on the schedule for 1.6.1? Having C++11 support really opens up a lot of (optimization) opportunities, at little to know cost. It's a fairly trivial change, too.

ffissore:
@ArduinoBot build this please

matthijskooijman:
Seems this missed the 1.6.1 release, even though it was set to the 1.6.1 milestone. It seems there's two dozen of PR's and issues that also have 1.6.1 set but weren't fixed. I wonder if there should be a "Next couple of releases" milestone where things get assigned first and then moved to a version milestone when it's (as good as) certain that it will be in that release? Or perhaps a label "Ready to merge" or "Fix soon" or whatever?

matthijskooijman:
I've just rebased this PR and re-ran the tests. Now, more examples compile to identical results than before, 82.6%. Not sure what changed there, though. A quick glance over the resulting assembly shows the same changes as described before.

matthijskooijman:
See https://www.google.com/fusiontables/data?docid=1tX3nZ7onuoBGuDoxDBbr7B0KRw-uKcbDTsEQfDlh#rows:id=1 for the build results.

ffissore:
@ArduinoBot build this please

matthijskooijman:
On addition: Previously, this PR enabled C++11 and made the C99 default explicit. However, I've found out that the default for C is actually C89, so this PR actually changed the C standard used (which I think is a good thing). The latest version of this PR now also enables C11, the most recent version of the C standard (which, compared to C99, hasn't seen a lot of changes, just a few small additions of which `static_assert` is probably the most useful one).

matthijskooijman:
I just found another reason for wanting C++11: [delegating constructors](https://en.wikipedia.org/wiki/C++11#Object_construction_improvement), where a constructor can call other constructors in the same class.

Is there anything blocking this? It's fairly trivial change, and more and more code seems to benefit from it.

Chris--A:
There are numerous reasons I can vouch for activating C++11.
Personally I can see some great simplifications for newbies by using foreach constructs (#2179) and I think user defined literals have a lot of potential with embedded style code.

cmaglie:
Rebased and merged.
Thanks for the meaningful discussion, and sorry for delay!

PR #2177, opened 2014-07-11
Adding plotting functionality to the editor

henningpohl:
I found debugging Arduino projects with, e.g., time-varying capacitive sensor values rather tedious with the serial monitor. I added a graph view that makes it easier to observe such changes.


q2dg:
+1 !!

ffissore:
@henningpohl can you please rebase/merge your PR with the current master? It doesn't merge any more

henningpohl:
So, my PR now is quite old and a lot of things seem to have changed in the codebase. Also around SerialMonitor. If I update my code, should the plotter also extend AbstractMonitor? Conceptually that would make sense. However, AbstractMonitor isn't really that abstract. There are already a whole bunch of UI components in there which would not be needed for plotting. That could be moved to an AbstractTextualMonitor or something like that. But that would be a larger change. Any thoughts?

ffissore:
Well it's abstract as far as serial and network monitors are concerned. Sure a plotter monitor requires more abstraction, or a totally independent code. I would do the minimum for making it viable and still be, if not proud, at least convinced of the quality of the code.

henningpohl:
Alright, I got everything to compile and work again. It would be great if somebody more familiar with the codebase could have a look. I mostly tried to go by what was in place in SerialMonitor and AbstractMonitor, but might have misunderstood some intentions here or there.

ffissore:
@ArduinoBot build this please

facchinm:
:+1: !!!
Only one problem: it doesn't close the serial port when closing the plot window, so the serial monitor can't open anymore (pretty annoying) -> saw on Linux x64 with a Leonardo

henningpohl:
Hmm... I tested this on Ubuntu 14.04 (64 bit) in a VM and on Windows 8 (64 bit) and the serial port closed on both. I think I did find the problem though: if one tries to open the serial monitor/plotter while the other window is open, that fails but puts the application in a state where opening will also fail in the future even if the other window is closed. I added a check for that and switching between both windows should now work.

The better solution would be to have the serial connection somewhere central in the editor and allow any number of subscriptions to incoming data elsewhere. But that probably would be a larger change on its own.

ffissore:
@ArduinoBot build this please

PedroLopes:
Awesome pull request, definitively a feature I want to have in the IDE! 

(if possible in the future something in the lines of https://code.google.com/p/arduinoscope/ would be even better, but this is a fantastic start!)

Chris--A:
This is a really nice feature, there is many ways it can be expanded into a super useful tool while keeping the interface as basic as possible.

mixania:
Wish I had the opportunity to test this but I'm happy to see this feature come!! 👏👏👏

ffissore:
@mixania you still have it: code still needs to be merged

henningpohl:
@ArduinoBot build this please

ffissore:
@ArduinoBot build this please

ffissore:
@ArduinoBot build this please

q2dg:
Serial plotter is available in BUILD 317 but it is not in 1.6.5. Why?
Thanks

ffissore:
@henningpohl after some consideration, we would like to merge this feature. However we are puzzled by the added button next serial monitor one: we think either it won't get noticed or it will confuse users. Is removing the button and leaving the menu entry under `Tools` ok for you?
Restoring previous versions of EditorToolbar.java and buttons.git should suffice

henningpohl:
Great! I'd be fine with removing the button. Would it be an option to leave the image in buttons.gif though? While adding a button to the interface is an easy change, recreating an icon is a bit more work. There might be a button in the future, or the icon might be used in the window title. I could alternatively also make removal a dedicated commit, keeping the larger buttons.gif in the history. 

I'm also out of office right now, so can't properly test changes. I will update this and get it ready early next week.

ffissore:
The changed image is already in the history. If you prefer, it's ok to edit buttons.gif in a dedicated commit

ffissore:
Thanks!

q2dg:
Could it be possible to add to Serial Plotter the ability of recognize several serial streams to be able to paint several graphs accordingly?? I mean...I'd like to see on Serial Plotter the evolution of data got by two different analog sensors at the same time, in two different coloured graphs.
Thanks.

PR #2179, opened 2014-07-13
Updated String library to use C++11 iterators.

Chris--A:
This will allow using the String library in a ranged for loop. Also the additions expose the first and last character pointers, which may be of use somewhere.

```C++
void setup() {
  
  Serial.begin( 9600 );
  String s = "Hi, this is a test of C++11 iterators.";
  
  for( char c : s ){
    Serial.print( "Character in string: " );
    Serial.println( c );
  }
}
void loop(){}
```

This modification is placed inside the check for C++11 which is also due for update here: https://github.com/matthijskooijman/Arduino/commit/6624d614fa1f95c26c471ae494012923e03a2e78

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.h
+++ b/hardware/arduino/avr/cores/arduino/WString.h
@@ -89,8 +89,6 @@ public:
 	#ifdef __GXX_EXPERIMENTAL_CXX0X__
 	String & operator = (String &&rval);
 	String & operator = (StringSumHelper &&rval);
-	auto begin() -> const char* { return c_str(); }
-	auto end() -> const char* { return c_str() + length(); }	
 	#endif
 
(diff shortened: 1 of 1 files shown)


Chris--A:
This update does not need to wait until C++11 is turned on.

It can be of use now, just like STL containers:

```C++
for( char *ptr = s.begin() ; ptr != s.end() ; *ptr++ = /* modify each character in string */ );
```

It might be worth noting if documented, that end() returns the null pointer, so the loop above and a C++11 ranged loop do not include the null in the iteration.

Chris--A:
Now that C++11 has been enabled, can this be merged too!

cmaglie:
Rebased and merged, thanks @Chris--A !

PR #2200, opened 2014-07-20
Update RobotIR library to play nice with other versions of IR-Remote

joshuajnoble:
This leaves all the functionality of RobotIR in place but updates the underlying IRRemote library that I'm maintaining now. The AVR_ROBOT_CONTROL and AVR_ROBOT_MOTOR #defines are used to indicate that the board is the Arduino Robot and that the pin defines should be left in.


PaoloP74:
Can you change the name of library from IrRemote to RobotIRremote?
This would avoid conflict with the original IrRemote.

PaulStoffregen:
I have been working on improved header to library name matching code.  It's still a work-in-progress, but already solves this problem and others.  I was planning to add several more cases before submitting a pull request.  But maybe it should be considered now, and more cases added later?

ffissore:
Closing this as it has been fixed with #2681

PR #2237, opened 2014-08-13
Resolved a few compile warnings.

Chris--A:
This patch will resolve a handful of warnings when compiling an empty sketch. It would be great to have the core warning free so the `-w` compile flag can be removed.

Sorry for all the commits, still working on the GitHub clone issue.


Chris--A:
Things have changed since this pr. I can look at redoing this patch later.

PR #2240, opened 2014-08-14
Added idle mode to the delay function

Zorph:
The controller will be sent to idle mode if the remaining delay time allows this. It will be woken up by any interrupt, which will be latest by the timer0 interrupt.

### DIFF
--- a/hardware/arduino/cores/arduino/wiring.c
+++ b/hardware/arduino/cores/arduino/wiring.c
@@ -23,6 +23,7 @@
 */
 
 #include "wiring_private.h"
+#include <avr/sleep.h>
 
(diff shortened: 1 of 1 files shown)


Zorph:
Hi,
Did this ever make it into the built? Just thought about this again.
Sorry, I can't really test the compatibility of it by myself, since I don't know the Arduino system so much. I just copied some functions for my own AtMega programming, and put this change back since I think this is needed to run Arduinos on battery.
Maybe creating a second delay function with sleep functionality would be better to avoid conflicts.

facchinm:
I'm closing the issue here; if someone wants to adopt it, please follow this link https://github.com/arduino/ArduinoCore-avr/compare/master...pr_2240?expand=1 to reopen it in the proper repo

PR #2280, opened 2014-09-01
Less surprising parseFloat() handling of multiple '.'

AbstractBeliefs:
When handling decimal points in incoming strings, break on any 2nd point.

If sending "1.2.3" with base, parseFloat() will return 1.23, which is probably not the intention.
With this update, it will return 1.20.

Here's a quick sketch to compare the difference: https://gist.github.com/AbstractBeliefs/3352f6fbd378375b70db


Chris--A:
This can be resolved with: #3337

PR #2291, opened 2014-09-08
9-bit frame support for HardwareSerial 1.5.X

Bouni:
I've added support for sending and receiving 9-bit frames for AVR as well as SAM based boards.
The changes are tested with a test sketch [1] and also with a logic analyzer (Saleae Logic 8).

On AVR based boards i use two bytes in the buffer for storing a 9 bit frame. So i don't have to change the buffer type to uint16_t which means that all other modes then 9 bit don't waste RAM. The only downside for 9 bit users is that the buffer has only half the size. 

[1] [Test Sketch](https://gist.github.com/Bouni/7087db3915bc530b41c2)


Bouni:
I'll close this as 7 years have gone by, so I guess nobody cares about this anymore 🤷🏽

PR #2292, opened 2014-09-08
9-bit frame support to Hardware Serial 1.0.X

Bouni:
I've added support for sending and receiving 9-bit frames for AVR based boards.
The changes are tested with a test sketch [1] and also with a logic analyzer (Saleae Logic 8).

I use two bytes in the buffer for storing a 9 bit frame. So i don't have to change the buffer type to uint16_t which means that all other modes then 9 bit don't waste RAM. The only downside for 9 bit users is that the buffer has only half the size.

[1] [Test Sketch](https://gist.github.com/Bouni/7087db3915bc530b41c2)

### DIFF
--- a/hardware/arduino/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/cores/arduino/HardwareSerial.cpp
@@ -19,6 +19,7 @@
   Modified 23 November 2006 by David A. Mellis
   Modified 28 September 2010 by Mark Sproul
   Modified 14 August 2012 by Alarus
+  Modified 12 April 2014 by Bouni
 */
 
 #include <stdlib.h>
@@ -121,14 +122,26 @@ inline void store_char(unsigned char c, ring_buffer *buffer)
   {
   #if defined(UDR0)
     if (bit_is_clear(UCSR0A, UPE0)) {
-      unsigned char c = UDR0;
+      unsigned char c;
+      // if 9 bit mode, read the ninth bit
+      if(bit_is_set(UCSR0B, UCSZ02)) {
+        c = (UCSR0B >> TXB80) & 0x01;
+        store_char(c, &rx_buffer);
+      }
+      c = UDR0;
       store_char(c, &rx_buffer);
     } else {
       unsigned char c = UDR0;
     };
   #elif defined(UDR)
     if (bit_is_clear(UCSRA, PE)) {
-      unsigned char c = UDR;
+      unsigned char c;
+      // if 9 bit mode, read the ninth bit
+      if(bit_is_set(UCSRB, UCSZ2)) {
+        c = (UCSRB >> TXB8) & 0x01;
+        store_char(c, &rx_buffer);
+      }
+      c = UDR;
       store_char(c, &rx_buffer);
     } else {
       unsigned char c = UDR;
@@ -147,7 +160,13 @@ inline void store_char(unsigned char c, ring_buffer *buffer)
   ISR(USART1_RX_vect)
   {
     if (bit_is_clear(UCSR1A, UPE1)) {
-      unsigned char c = UDR1;
+      unsigned char c;
+      // if 9 bit mode, read the ninth bit
+      if(bit_is_set(UCSR1B, UCSZ12)) {
+        c = (UCSR1B >> TXB81) & 0x01;
+        store_char(c, &rx_buffer1);
+      }
+      c = UDR1; 
       store_char(c, &rx_buffer1);
     } else {
       unsigned char c = UDR1;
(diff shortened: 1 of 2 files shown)


NicoHood:
I like the idea but would rather use templates for this. Because as it is right now it adds more flash to the serial. And its not likely that people will use this 9 bit more, even though its really cool.

So in my opinion the whole Serial should be reworked as template, but they wont accept this anyways I guess.

Bouni:
Unfortunately i have no idea how to use templates to achieve this :-/

NicoHood:
FYI: I also found this:
https://github.com/arduino/Arduino/pull/1221

Bouni:
@NicoHood 
I was in touch with @agalliazzo before i started my work on the 9-bit support.
He said that his work was made such a long time ago that it will be a pain to merge it with the state of the Arduino code at that time. 
Also i wanted an approach that doesn't require to change the entire buffers to 16-bit.

NicoHood:
Well this now also applies to your code, as I've checked it.
the Arduino core now uses 3 different .cpp files for each Serial instance. So you have to migrate it anyways.

I am currently working on a basic setup to send 9N1 bytes. I got it working with 8N1. with 9N1 as sending setting and 8N1 as receiving setting I still get the lower 8 byte of data. Is this a normal behaviour? the extra bit is just thrown away as far as I can see.

Bouni:
@NicoHood 
Have you seen my other Pull request #2291 ?

NicoHood:
I think I've missed it. Thanks.
I currently have no time to test it though.

Bouni:
@NicoHood you mean compare the standard Arduino IDE to my extended 9-bit version?

NicoHood:
Yes. But with the 1.6.x release. So we can see if it adds much overhead if this would be merged.
Compare a standard 8N1 setting with the standard IDE and your code and also post how big a 9N1 setting (not available in the IDE) compiles. So we have 3 setting to compare:

IDE 8N1
PR 8N1
PR 9N1

with ram and flash usage.

Bouni:
@NicoHood Can you explain how I can get the RAM and Flash usage?

NicoHood:
Its displayed after compiling.

Bouni:
@NicoHood Ok, i did a few tests:

i compiled SerialCallResponse.ino with 4 different Arduino setups using the CLI (except for 1.0.5)

>./arduino --verify --board:avr:uno examples/04.Communication/SerialCallResponse/SerialCallResponse.ino

**Arduino 1.0.5** (master branch at the time i did my modifications)

>Binary sketch size: 3,060 bytes (of a 32,256 byte maximum)

**Arduino ide-1.5.x** (for comparison to 1.0.5)

>Sketch uses 2,986 bytes (9%) of program storage space. Maximum is 32,256 bytes.
>Global variables use 192 bytes (9%) of dynamic memory, leaving 1,856 bytes for local variables. >Maximum is 2,048 bytes.

**Arduino ide-1.5.x-hardware-serial-9-bit**

Sketch uses 3,470 bytes (10%) of program storage space. Maximum is 32,256 bytes.
Global variables use 194 bytes (9%) of dynamic memory, leaving 1,854 bytes for local variables. Maximum is 2,048 bytes.

**Arduino 1.6.4** (current version installed from arch linux AUR repo)

>Sketch uses 2,878 bytes (8%) of program storage space. Maximum is 32,256 bytes.
>Global variables use 196 bytes (9%) of dynamic memory, leaving 1,852 bytes for local variables. >Maximum is 2,048 bytes.

NicoHood:
Those results are unclear to me. What is 1.5.x? What exact version did you use here? And what serial setting? 8N1 or 9N1? You have to tell me the Serial setting, otherwise it doesnt make any sense. I personally also do not care about 1.0.5 at all.

You need to test all 3 that I listed above with 1.6.4 (or 1.6.5).

IDE 8N1 vs PR 8N1 tells us how much overhead this PR gives us when its just implemented. If nobody uses those 9N1 function he also has the delta in his sketch. If this delta is too big its not worth to merge the PR.

PR 8N1 vs PR 9N1 tells us if using 9N1 with the PR already enabled gives us even more overhead. But my guess is that there is no more overhead.

Bouni:
@NicoHood post code snippet(s) you want to have compiled and i'll do it for 1.6.4, and the Pull request!

> What is 1.5.x?
its the branch ide-1.5.x which was the development branch for the 1.5 version of Arduino at that time,
1.5 was not released back then.

NicoHood:
A simple Serial.begin(9600, 8N1) in setup() and a Serial.println("Hello World"); delay(1000); in the loop() should do it.

facchinm:
Closing as 1.0.x branch is in maintenance mode only.

Issue #2314, opened 2014-09-19
ArduinoISP example sketch does not work with Mega or Due


championswimmer:
The ArduinoISP included example works fine with Uno or Leonardo

But it fails to work for Mega (this works good for Mega https://raw.githubusercontent.com/rsbohn/ArduinoISP/master/ArduinoISP/ArduinoISP.ino)

 and it cannot compile at all for Due as arm doesn't support spi_* functions like avr gcc (this works good for Due https://raw.githubusercontent.com/PeterVH/ArduinoISP/due/ArduinoISP/ArduinoISP.ino)

facchinm:
Solved by #3500

PR #2315, opened 2014-09-19
add mega and due compatible ArduinoISP sketches 

championswimmer:
To fix https://github.com/arduino/Arduino/issues/2314

  The example present does not work with either
   Arduino Mega 2560 or Arduino Due

  Tested this 
   https://raw.githubusercontent.com/rsbohn/ArduinoISP/master/ArduinoISP/ArduinoISP.ino
   to work perfectly to turn Mega into a stk500v1 programmer

   And tested this
    https://raw.githubusercontent.com/PeterVH/ArduinoISP/due/ArduinoISP/ArduinoISP.ino
   to be compatible with Due to turn it into a stk500v1 programmer. (using the native port)

Signed-off-by: Arnav Gupta <championswimmer@gmail.com>


NicoHood:
Both versions doesnt work with a Leonardo to me. I also found this tutorial, but that also doesnt work. Using IDE 1.6 rc1
https://petervanhoyweghen.wordpress.com/2012/09/16/arduinoisp-on-the-leonardo/

Does anyone know a working ISP sketch for the 32u4?

PeterVH:
I recently picked up the effort to make ArduinoISP more reliable and better working on more (hopefully all) arduino models. I started again from rsbohn's last work and try to fix only things that help achieving these two goals.

This effort is in my [master branch] (https://github.com/PeterVH/ArduinoISP/tree/master).

**Status**

This is still a work in progress, but here is a summary of the status:
* Rsbohn's last work introduced some regressions, but these are fixed.

* On the Uno it works much more reliable than the version in the IDE, mainly by simply removing the delay() call from the heartbeat() method.

* On the Due. There is no bit bang spi yet: I think this would better be handled by a SPI.setSpeed() method in the SPI lib (if the selected speed is too slow, the implementation could automatically switch to bit banged SPI).

Therefore on the Due it is already working (untill now I only tested reading signatures) but not for slow targets like the attiny85...

**Testing:** I am composing a test suite bundling a set of about 10 typical/important use cases for ArduinoISP. The suite needs to be fleshed out, but here is a preview:
1. No target connected.  (note this does not work as it should with the current version in the IDE)
2. Burn UNO bootloader
3. Burn mega bootloader (requires programming above 128KB)
4. Program an attiny85
5. Program an attiny841
6. Program an at89s52
7. Chaucer115 to 1284p: Program a big (115KB) sketch into an atmega 1284p
8. Chaucer115 to mega: Program a big (115KB) sketch into an arduino mega
9. LongStoryShort: Program above 128KB.

(the name Chaucer comes from the optiboot test suite that burns big sketches that output part of the "knight's tale" by Chaucer on the console)

**Avrdude:** I think it would be better to upgrade to avrdude 6.1 because this version implements some clever fixes that make it possible to program high addresses (over 128KB) with stk500v1 (which is what ArduinoISP uses). This makes it possible to reliably burn the bootloader of an atmega2560 using ArduinoISP.

facchinm:
Hi @PeterVH , I recently took a look at your patches over ArduinoISP and it works much better then the one upstream. Would you mind opening a PR (with the complete history, not only a squash of the patches) on the top of the current sketch? It will of course supersede this PR.

Test suite is also a great idea! We'd be glad to include it once it's ready.

About avrdude 6.1, I'm :+1: for the update; if you want to speedup the process please add a comment on arduino/toolchain-avr/issues/17

PeterVH:
>Would you mind opening a PR...

Done : #3321
> ...with the complete history, 

I also kept the link with rs-bohn's work: in my ArduinoISP fork, I created a branch at commit db0fd332 on which I applied the IDE's later patches (basically an auto format and a baudrate/heartbeat delay dance). Then I rebased rsbohn and my patches on that branch. I forced (!) that branch to become my master branch. (My previous master branch is still there, it is called called rs-bohn).

The same history is also in [branch issue-3321 of my Arduino fork] (https://github.com/PeterVH/Arduino/tree/issue-3321)

There is no pull request yet as I still need to address the slow spi clock issue on the due.

facchinm:
Hi @PeterVH ,
great job! 
Feel free to open the pull request even if the patchset is not complete yet. 
Github will always let you add more commits afterwards by only pushing them to the PR's topic branch (so it's better to change the rebased HEAD on your repo from `master` to something like `ArduinoISP_PR`)

facchinm:
superseded by #3321

Issue #2304, opened 2014-09-13
Arduino IDE preprocessor causes unknown type references from prototype creation


JeffBobbo:
Functions using/returned an object, structs or using typedefs cause bad code to be fed to the compiler:

Source code:
http://0bin.net/paste/Xz5MZdbmuafYUQ2Q#567rrlhZy8-xCaiI7geiRXEZ+DFo92e4dpZ/cAJvJ3j

Arduino preprocessed code (fed to avr-g++):
http://0bin.net/paste/JgtLxMb23ZOaGJMu#RYZv9sCkcSXeapn0XXRCKphshFH38YLSHmGmn2Iy4EL

ffissore:
We've just posted a new preprocessor version at https://groups.google.com/a/arduino.cc/d/msg/developers/4X2T3rCxXWM/WbQ2_dGzuhsJ
While it does not solve your issue, at least it allows you to work around it by specifying the prototype your self. Try adding a `Foo makeFoo();` at line 21: the sketch will compile fine.

ffissore:
Fixed by #3779

Issue #992, opened 2012-11-15
Function prototypes not generated for variable argument functions [imported]


cmaglie:
This is [Issue 992](http://code.google.com/p/arduino/issues/detail?id=992) moved from a Google Code project.
Added by 2012-07-27T17:08:51.000Z by [janar.s...@gmail.com](http://code.google.com/u/115724867442263714249/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium, Component-IDE

### Original description

**What steps will reproduce the problem?**
1. Make new Arduino project having a main file and additional ".ino" file.

2. Into additional ".ino" file insert a function definition having C style variable arguments.
void nice_func(const char *fmt, ...) 
{
  // do nothing
}

3. Now insert call to that function into main file. For example inside
loop function:
void loop()
{  
  nice_func("test", 1);
}

During compilation of the whole project Arduino gives error that
'nice_func' was not declared in this scope.

When we peek into temporary build directory then we can see that Arduino is not inserting that function prototype definition into final .cpp file.

For example if you change inside additional ".ino" file this function definition to:
void nice_func(const char *fmt, int i)
{
  // do nothing
}
and now it'll compile without errors.

It seems that file building functionality (part that is inserting prototype definitions)
inside Arduino is skipping functions having "..." inside arguments area.

I'm using Arduino 1.0.1 on WinXP x64.

ffissore:
Please checkout this thread [1] on devs mailing list and try one of the linked IDEs. It should fix the problems reported in this issue
[1] https://groups.google.com/a/arduino.cc/forum/#!topic/developers/4X2T3rCxXWM

ffissore:
New preprocessor tracked at #2636. Builds for testing it are available

ffissore:
Fixed by #3779

Issue #386, opened 2012-11-15
Prototypes for methods with default arguments are not generated [imported]


cmaglie:
This is [Issue 386](http://code.google.com/p/arduino/issues/detail?id=386) moved from a Google Code project.
Added by 2010-10-23T13:23:50.000Z by [andre...@gmail.com](http://code.google.com/u/101138032102755792048/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium, Component-PreProcessor

### Original description

Steps to reproduce the issue:

STEP 1. Insert the following code into a blank sketch:

```C++
void setup() {
  test(42);
  test2(42);
}

void loop() {}
void test(int arg) {};
void test2(int arg = 0) {};
```

STEP 2: Click the verify button and observe the error message:

```
DefaultArgBugDemo.cpp: In function 'void setup()':
DefaultArgBugDemo:2: error: 'test2' was not declared in this scope
```

STEP 3:
Examine the generated *.cpp file and observe that a prototype has been generated for test1 but not test2:

```C++
#include "WProgram.h"
void setup();
void loop();
void test(int arg);
void setup() {
  test(42);
  test2(42);
}

void loop() {}
void test(int arg) {};
void test2(int arg = 0) {};
```

**What is the expected output? What do you see instead?**

You would expect a prototype to be generated for test2.

**What version of the Arduino software are you using? On what operating**
**system?  Which Arduino board are you using?**

Tested in 0018 and 0021. Running on Win7 Home Premium (though the issue likely exists on all platforms)

**Please provide any additional information below.**

Of course the workaround is relatively simple (just define the prototype yourself) but it would be nice for all methods to be treated equally.

ffissore:
Please checkout this thread [1] on devs mailing list and try one of the linked IDEs. It should fix the problems reported in this issue
[1] https://groups.google.com/a/arduino.cc/forum/#!topic/developers/4X2T3rCxXWM

ffissore:
With #3779 the preprocessed code is now 
```
#include <Arduino.h>
#line 1
#line 1 "sketch_sep11c.ino"
void setup();
void loop();
void test(int arg);
void test2(int arg = 0);
#line 1
void setup() {
  test(42);
  test2(42);
}

void loop() {}
void test(int arg) {};
void test2(int arg = 0) {};
```

which now fails with 
```
sketch_sep11c.ino: In function 'void test2(int)':
[...]

PR #2324, opened 2014-09-24
Preference for toggling preprocessor prototype generation

JeffBobbo:
Added preproc.generate.prototypes preference setting to disable the generation of function prototypes (also done some cleanup) - Issues: #2304, #992 and #386 (workaround)


ffissore:
Closing as wontfix. Preprocessing has much improved since the introduction of https://github.com/arduino/arduino-builder, available with the latest hourly build http://www.arduino.cc/en/Main/Software#hourly

Issue #2028, opened 2014-04-23
Bridge: understand if and when Yun linux side is ready


ffissore:
This gist https://gist.github.com/wayoda/db3c023417757f726088 provides a proof of concept boot monitor on the Yun. Using the handshake line on pin 7, the sketch is able to tell when the linux side is ready. Consider using this technique as an alternative to the current "read the output and wait for a prompt" algorithm.
/cc @cmaglie

ffissore:
Requires a hardware fix and it's not convenient

PR #2353, opened 2014-10-09
Add check for bridge

arve0:
This will allow restarting bridge connection. Hack for #2028.

```
if (Bridge.isOk() != true) {
  Serial.println("waiting..");
  delay(60000); // hope for ar9331 booting
}
// restart bridge
Bridge.started = false;
Bridge.begin();
```

### DIFF
--- a/libraries/Bridge/src/Bridge.cpp
+++ b/libraries/Bridge/src/Bridge.cpp
@@ -88,9 +88,7 @@ bool BridgeClass::isOk() {
   // check if bridge connections is OK
   uint8_t cmd[] = {'X','X'};
 
-  uint8_t rxbuff[2]; // length 2 is needed because transfer will truncate to rxlen
-  rxbuff[0] = 0; // clear buffer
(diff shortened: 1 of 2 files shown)


facchinm:
As in #2781, I did not merge this PR because it's more like an hack than a real fix.
In fact, the "XX" command you are sending triggers this function (https://github.com/arduino/YunBridge/blob/master/bridge/packet.py#L98) which launches `usr/bin/blink-start` (unwanted) and returns `2`.
A real "ping" function would be a good improvement but must be implemented by both side of the Bridge; on the python side it will be something like 
```python 
self.processor.register('Q', PING_Command(self))
```

I compiled a list of commands already in use https://gist.github.com/facchinm/569c1c8d112d28d8ff73 so if you want to update your merge request dividing it in two parts (one here and one for YunBridge) while keeping the same functionality it would be great!

arve0:
Thanks for the feedback. I might look into creating a correct ping, if I find some time.

PR #2356, opened 2014-10-11
Enable Keyboard library to send key scancodes.

jclopes:
This gives more flexibility to the library and allows the development of keyboard driver.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/HID.cpp
+++ b/hardware/arduino/avr/cores/arduino/HID.cpp
@@ -494,6 +494,61 @@ size_t Keyboard_::release(uint8_t k)
 	return 1;
 }
 
+// press() adds the specified key scancode
+// to the persistent key report and sends the report.  Because of the way
+// USB HID works, the host acts like the key remains pressed until we
+// call release(), releaseAll(), or otherwise clear the report and resend.
+size_t Keyboard_::press_sc(uint8_t k)
+{
+	uint8_t i;
+
+	if (k >= 224 && k <= 231) {
+		_keyReport.modifiers |= (1 << (k - 224));
+	}
+	// Add k to the key report only if it's not already present
+	// and if there is an empty slot.
+	else if (_keyReport.keys[0] != k && _keyReport.keys[1] != k &&
+		_keyReport.keys[2] != k && _keyReport.keys[3] != k &&
+		_keyReport.keys[4] != k && _keyReport.keys[5] != k) {
+
+		for (i=0; i<6; i++) {
+			if (_keyReport.keys[i] == 0x00) {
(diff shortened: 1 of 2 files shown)


RasmusB:
I just tried this patch out for my project. The ability to send raw scancodes removes a lot of unnecessary overhead from the Arduino. It works just as expected, thank you!

ffissore:
@jclopes after 1.6.0 release, your PR does not merge any more. Can you please update it with the current master?

PaulStoffregen:
I implemented someone very similar on Teensy, without adding new functions to the API.  I simply increased the input from uint8_t to uint16_t, and defined symbols for the raw codes with the MSB (bit 15) set.

Maybe you like this approach better, but it might be work considering reusing the existing function.  Conceptually, it does exactly the same thing.

jclopes:
@ffissore I rebased my master into the current arduino/master. It should work now.

@PaulStoffregen I prefer to have an explicit API then to use a convention for the functions input. IMHO it makes the code easier to read.

PaulStoffregen:
@jclopes You are correct, separate functions for ascii & scan codes does indeed make Arduino's source code easier to read.  But does it also make Arduino easier to use for novices, who have little or no understanding of ascii vs scan codes?

ffissore:
FWIW, I too would like a generic function. Less code is usually better.
I'll make the bot build this PR.
@jclopes what about having public `press` dispatch to `press_sc` or (a new) `press_key`, both private or protected? Code would still be easy to read and the api won't change much

ffissore:
@ArduinoBot build this please

jclopes:
Regarding the novice user being confused - Since the old API is still there
and there are plenty of example code, I don't think that is a real problem.
I think is more confusing to have only one function that requires you to
pass some odd argument when you just want to send a char.

I don't see how having a wrapper function will help having less code, help
the experienced or the novice user.
But that could be done in a separate commit if there is demand for it.

ffissore:
It won't. I was just suggesting a mediation between paul's advice and your implementation

NicoHood:
Implemented in the current dev branch. All keys are now possible to press. This was solved via an enum. Have a look at the KeyboardAPI files. New keys will be added soon.

https://github.com/NicoHood/HID/tree/dev_2_4

We could adapt this API to the Arduino core or just leave it as it is right now. But I think @cmaglie and @facchinm tended to keep the api simple and leave such things to my library. I have no problems if this is also added to the IDE though.

But if you decide on not adding it, you may close this issue then.

facchinm:
Thanks @NicoHood for the explanation, indeed we'll keep Keyboard and Mouse library (now that they are outside the core) as simple as possible, while additional functionalities will be added in expanded libraries.

PR #2376, opened 2014-10-18
Ide 1.5.x AVR and sam (ARM) SPI atomicity fixes

xxxajk:
These are the fixes for interacting with SPI methods without accidentally stepping on them from some other place, such as a scheduler, or other non-pin interrupt source. It also prevents updates to the variables from being interrupted by a pin interrupt, which can corrupt the mask/stored mask. and other states.

AVR is 100% tested, and solid.
sam is untested because I lack the hardware to do so.

### DIFF
--- a/hardware/arduino/sam/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/sam/libraries/SPI/SPI.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2010 by Cristian Maglie <c.maglie@bug.st>
  * Copyright (c) 2014 by Paul Stoffregen <paul@pjrc.com> (Transaction API)
+ * Copyright (c) 2014 by Andrew J. Kroll <xxxajk@gmail.com> (atomicity fixes)
  * SPI Master library for arduino.
  *
  * This file is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


xxxajk:
I see that half the patch made it into 1.6.0 (the avr half).
Due, not yet.

facchinm:
Closing as already merged in master (please reopen if I'm wrong)

matthijskooijman:
I actually think this is only merged for AVR, not [SAM](https://github.com/arduino/ArduinoCore-sam/blob/master/libraries/SPI/src/SPI.cpp#L106) or [SAMD](https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/SPI/SPI.cpp#L115).

I was going to write that I'm not entirely convinced of the necessity of this atomicity, but looking through the comments it seems making `beginTransaction()` atomic is indeed relevant (as explained [here](https://github.com/arduino/Arduino/pull/2381#issuecomment-61119011) and [here](https://github.com/arduino/Arduino/pull/2449/commits/a9735bf91fda36eb9cd97274f01b86f1fb234155)). I'm not [convinced yet](https://github.com/arduino/Arduino/pull/2449#issuecomment-63607275) about the atomicity of `begin()`/`end()` and `usingInterrupts()`/`notUsingInterrupts()` but these won't really hurt either.

One other change included here that is still missing from SAM/SAMD is the addition of `notUsingInterrupts()`.

I'm not sure if we should reopen this, or just create new issues on the SAM and SAMD repositories to track this?

facchinm:
I'd go for keeping it closed and create new PRs on sam/samd

cmaglie:
> I was going to write that I'm not entirely convinced of the necessity of this atomicity

About the atomicity fix for AVR I remember your comment [here](https://github.com/arduino/Arduino/pull/2381#issuecomment-61292161) and by re-reading it now I see that at the times we didn't realize that the operation:

> Main: `SPI_AVR_EIMSK &= ~interruptMask;`

is probably not compiled as a single-instruction but as a read+set+write. My feeling is that taking this into consideration we may optimize the code and remove the guards. Anyway, the SPI lib is now well tested and untouched for a very long time, changing it again will require a lot of validation test I don't have the time to look again into this.

About the atomicity fix for SAM, since we hadn't any report about issues on SPI, my feeling is that the guard is not needed but, again, this is totally empirical.

matthijskooijman:
> My feeling is that taking this into consideration we may optimize the code and remove the guards.

I'm not sure why that would be? Wouldn't this only be more reason to do this atomically? Consider a case where *another* pin ISR runs (e.g. one that is not in `interruptMask`) that disables itself. If that happens between the read and write, that other interrupt could be *re-enabled* by this code, which is clearly not intended.

> About the atomicity fix for SAM, since we hadn't any report about issues on SPI, my feeling is that the guard is not needed but, again, this is totally empirical.

I'm pretty sure it is needed, for the same reasons on AVR, but it perhaps it doesn't show up so much due to the limited number of users, or because they don't often do tricky things in ISRs? In any case, it would be good to at least create an issue in both repositories to track this, probably with a link to my comment above (which again links to various other relevant comments). I don't have time to create the tickets right now, though.

cmaglie:
> Wouldn't this only be more reason to do this atomically? Consider a case where another pin ISR runs (e.g. one that is not in interruptMask) that disables itself. If that happens between the read and write, that other interrupt could be re-enabled by this code, which is clearly not intended.

oh, you're right :-)

> I'm pretty sure it is needed, for the same reasons on AVR, but it perhaps it doesn't show up so much due to the limited number of users

No, my thought is that on ARM disabling the ISR is already atomic (-> is done in a single instruction), but this is just a thought, all to be verified :-)

matthijskooijman:
> No, my thought is that on ARM disabling the ISR is already atomic (-> is done in a single instruction), but this is just a thought, all to be verified :-)

Ah, but that still leaves the issue with the static `interruptSave` variable that you pointed out [here](https://github.com/arduino/Arduino/pull/2381#issuecomment-61119011) and expanded by me a bit more [here](https://github.com/arduino/Arduino/pull/2381#issuecomment-61292161). OTOH, in that last comment I note that this might indeed by AVR-specific, so if it's timed differently on ARM, it might indeed be safe already (needs to be checked, though).

I also just realized that the race condition I described here:

> Consider a case where another pin ISR runs (e.g. one that is not in interruptMask) that disables itself. If that happens between the read and write, that other interrupt could be re-enabled by this code, which is clearly not intended.

is a lot broader then I thought. If, at any point during an SPI transaction, a pin interrupt (that is or is *not* in `interruptMask` is disabled), it will be re-enabled by `endTransaction()`. For interrupts that are in `interruptMask`, this is hard to fix (but also less of a problem: The ISR itself cannot run to disable itself, so it must be the code that's in control of the begin/endTransaction that does it, which should then just be written differently). For interrupts that are *not* in `interruptMask` we should probably fix this. I think fix would be to replace [this line](https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/SPI/src/SPI.h#L276) by:

    SPI_AVR_EIMSK |= (interruptSave & `interruptMask`);

(or possibly do the masking with `interruptMask` in `beginTransaction` already to more resilient to changes in `interruptMask` during a transaction, even though we don't really support that)

PR #2381, opened 2014-10-21
[AVR] SPI atomicity fix avr (part 1/2 of #2376)

cmaglie:
This pull request broke into 6 simpler commits the big patch in #2376.

/cc @xxxajk @PaulStoffregen

### DIFF
--- a/hardware/arduino/avr/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/SPI.cpp
@@ -12,7 +12,6 @@
  */
 
 #include "SPI.h"
-#include "pins_arduino.h"
 
 SPIClass SPI;
 SPIflags_t SPIClass::modeFlags = {false, false, 0};


nickgammon:
Would you please document these functions on the reference page: http://arduino.cc/en/Reference/SPI

* beginTransaction
* usingInterrupt
* notUsingInterrupt
* transfer16
* endTransaction

I have code that is now failing under IDE 1.6.0 due to SPI behaving differently to previous versions. I had to find a reference to this page in the source, but no official documentation on the web site.

Specifically, SPI.transfer (0) is hanging indefinitely. I am presuming that an interrupt is (unexpectedly) being generated due to legacy code (ie. code that used to work in all previous versions of the IDE) not setting up beginTransaction / endTransaction or an interrupt handler.

nickgammon:
It turns out that I was inadvertently calling SPI.begin() twice and SPI.end() once. This meant that after calling SPI.end() and then calling SPI.begin() again the second SPI.begin() was effectively a NO-OP because of the reference counter. It might be worth pointing out in the documentation that SPI.begin() and SPI.end() must be balanced in the number of calls, or you get unexpected behaviour.

nickgammon:
See: https://github.com/arduino/Arduino/issues/2654

PR #2387, opened 2014-10-24
Move the flush method from Stream to Print

matthijskooijman:
This method originally flushed pending input bytes, which makes sense in
Stream. At some point it was changed to flush output bytes instead, but
it was never moved to Print to reflect this.

Since Stream inherits from Print, this should not really affect any
users of the Stream or Print classes. However to prevent problems with
existing implementations of the Print class that do not provide a
flush() implementation, a default implementation is provided. We should
probably remove this at some point in the future, though.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.h
+++ b/hardware/arduino/avr/cores/arduino/Print.h
@@ -79,6 +79,8 @@ class Print
     size_t println(double, int = 2);
     size_t println(const Printable&);
     size_t println(void);
+
+    virtual void flush() { /* Empty implementation for backward compatibility */ }
 };
 
 #endif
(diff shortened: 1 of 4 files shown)


matthijskooijman:
I just rebased this PR. Anything keeping it from being merged?

Chris--A:
+1, this should be merged.

I need this feature available from a `Print` only derived class.

eric-wieser:
I imagine this needs a rebase, since the sam files are now gone

PaulStoffregen:
This looks good to me.  I'd recommend merging.

eric-wieser:
@PaulStoffregen: I don't think merging through the online interface will do the right thing, since this modifies files that were deleted in master. I'm surprised this doesn't produce a merge conflict

cmaglie:
it's still mergeable becuase it targets the `ide-1.5.x` branch instead of `master`, I'll manually merge this commit and edit the conflicts.

cmaglie:
Manually merged here: https://github.com/arduino/Arduino/commit/ed7007c01076038f48033ffe535f2abecb728846

Issue #2419, opened 2014-11-06
FindReplace dialog layout is inconsistent and (sometimes) broken after i18n


wayoda:
### My runtime environment

I run  the Arduino IDE on Linux Ubuntu with the(default)  javax,swing.GTK theme.
### Dialog is broken in German translation

Here is a screenshot of the Find-Replace window as it appears in a German locale 
*_Linux GTK theme_
![searchandreplace-gtk](https://cloud.githubusercontent.com/assets/418786/4935228/eb3080d2-65a9-11e4-9f3e-c442c7f3ccd2.jpeg)
The _Ignore case_ checkbox is unreachable because the dialog window is not wide enough and not resizable. 

Looking at the code in `src/processing/app/FindReplace.java`reveals` an improper use of FlowLayout. The author throws all the components into a panel with the simple FlowLayoutManager and then  adds some pixel-calc voodoo  to break the components into lines.

In an (java-)application that runs on different platforms and which is translated into many languages this is a very bad idea. In Swing you must never mess around with window sizes or component positions/sizes. That is strictly the task of LayoutManagers. 
### Design is inconsistent across platforms

Here are three more versions of the Dialog as is appears on different platforms
**Linux Metal Theme**
![searchandreplace-metal](https://cloud.githubusercontent.com/assets/418786/4935418/8aefe700-65ac-11e4-879b-8fc08e60a7bc.jpeg)
**Linux Nimbus Theme**
![searchandreplace-nimbus](https://cloud.githubusercontent.com/assets/418786/4935423/94b75dae-65ac-11e4-90db-39de96bc58c6.jpeg)
**Windows 7 Theme**
![searchandreplace-windows](https://cloud.githubusercontent.com/assets/418786/4935428/b0b3d334-65ac-11e4-8cc9-f44d27796a82.JPG)

As you can see the result is quit different :
On the Metal-Theme the buttons blend into the bottom frame and on windows the input fields are suddenly side by side, 
### Fix it ?

Using a more versatile Layoutmanager will fix this problem. If you need  some help with this, just send me a note...

cmaglie:
Sorry, just figured out that only #2424 is a pull request, I'll check that.

PR #2424, opened 2014-11-07
Fix layout for Find-Replace dialog

wayoda:
This code fixes issue  #2419

Here is some code that fixes the layout problem in the Find-Replace Dialog.
I was able to test it on Linux and Windows 7.

fixed-searchandreplace-gtk
This  is an image of the Linux GTK version, but the layout is consistent across various javax.swing Themes (Metal, Nimbus, GTK, Windows)

Someone else has to do the Mac-Test I don't have access to one


cmaglie:
An updated PR is on #2540, any follow up there.

PR #2434, opened 2014-11-10
Made swap a function to prevent awful error messages from macros.

Chobbes:
This macro causes a lot of grief for poor students who aren't yet familiar with gcc's dreadful error messages. Unfortunately "swap" is a common name for functions, and if people aren't careful and declare their own functions when including Adafruit_GFX.h you get a misleading error, as opposed to a duplicate declaration one.

### DIFF
--- a/libraries/Robot_Control/utility/Adafruit_GFX.h
+++ b/libraries/Robot_Control/utility/Adafruit_GFX.h
@@ -25,7 +25,7 @@
 
 //#include "PImage.h"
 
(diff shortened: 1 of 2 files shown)


facchinm:
@cmaglie should we apply this to Robot_control and TFT repos even if they are in maintenance mode?

cmaglie:
I'd say yes.

Chobbes:
:+1:

facchinm:
Applied to the given repos, closing here

PR #2458, opened 2014-11-19
Add save hex option and update Sketch menu 

championswimmer:
* Have "Save Hex" option (with Ctrl + Shift + R as well)
 * Move Upload and Upload w/ Programmer into Sketch menu
   as they are functions of sketch and not file technically


championswimmer:
@ffissore , @cmaglie 

refactored on new 1.6 code and pushed here 
https://github.com/arduino/Arduino/pull/2567

Also uses java to copy file, instead of pattern

cmaglie:
Closing in favor of #2567.

PR #2469, opened 2014-11-24
Add USB Serial Port Callbacks

embmicro:
These edits allow the functions lineCodingEvent and lineStateEvent to be used in sketches to know when the virtual serial port's control lines or baud rate changes.

They are needed for creating sketches that act as a USB->Serial Port interface or interfacing with software that uses these signals as controls.

They were originally authored by Peter Van Hoyweghen http://petervanhoyweghen.wordpress.com/2012/11/08/using-the-leonardo-as-usb-to-serial-converter/


NicoHood:
A similar implementation can be found here:
https://github.com/NicoHood/HID

PaulStoffregen:
I would advise caution when considering interrupt callbacks.

Such callbacks are seductive, since they promise very immediate response, and because they're so easy to implement.

But exposing end users to interrupt context should be avoided, if possible.  Interrupt context is a nightmare of thorny programming, requiring very careful design to properly share any data with other parts of the code that aren't interrupt based.  Even experts often get it wrong.  Bugs tend to be subtle and almost impossible to diagnose.

A good portion of the Arduino API also isn't safe to use from interrupt context.

NicoHood:
This may be closed with the new API now.

cmaglie:
Fixed with #3343.

Issue #2198, opened 2014-07-20
Due bug: Cannot use pin as digital after analogRead


trickedj:
Once a pin has been used by analogRead it cannot be returned to digital use with pinMode(xxxx, OUTPUT) because analogRead only frees up the pin when another channel is read. pinMode needs to check if the pin is currently selected for the ADC and if so free it up.

A work around is to free up the pin by selecting a spare with analogRead

I discovered this while porting (from a Uno to Due) a TFT+touchpanel screen which shares digital TFT functions with raw touchpanel resistors.

facchinm:
I've added some thoughts and code on https://groups.google.com/a/arduino.cc/forum/#!topic/developers/IdhTIT0V5GY ,
I'd be very glad to have some discussion before committing any fix :wink:

facchinm:
Solved by #3524

PR #2486, opened 2014-12-05
this closes issue #2198

agdl:
In this way the problem doesn't subsist any more because  pinMode checks if pin is digital and if so disable ADC when pin is declared as OUTPUT and enable it if declared as INPUT


ffissore:
@ArduinoBot build this please

facchinm:
some thoughts and tests here https://groups.google.com/a/arduino.cc/forum/#!topic/developers/IdhTIT0V5GY

facchinm:
Closing because the solution has the side effect to break digitalRead() on analog pins.
Discussion on the fix continues in the mailing list

PR #2490, opened 2014-12-09
Fixed duplicate definitions of File when using SD and Bridge libraries together

cmaglie:
The two `File` classes have been enclosed into different namespaces.

An additional:

```C++
using namespace xxxxx;
```

has been added to guarantee compatibility with sketches that uses only one of
the two libraries.

`BridgeLib::File` and `SDLib::File` classes have been aliased to `YunFile` and `SDFile` respectively,
users are encouraged to use that instead of `File`.


ffissore:
@ArduinoBot build this please

cmaglie:
Manually merged.

PR #2500, opened 2014-12-15
Bugfix DNSClient::getHostByName()

jhorck:
I found a bug in DNSClient::getHostByName() and fixed it. 

I'm trying to connect to a local server, called 'htpc'. The hostname is resolved by DNSClient::getHostByName(). The first thing this function does is "See if it's a numeric IP address" by calling inet_aton(). It returns falsely 1 (=yes, it's a numeric IP address) for hostname 'htpc'

For most (longer) hostnames it will accidentally return 0. This is probably why it hasn't been discovered before. 

- valid IP address digits have to be >= 0 AND <= 9
- valid IP addresses must have 4 segments

### DIFF
--- a/libraries/Ethernet/src/Dns.cpp
+++ b/libraries/Ethernet/src/Dns.cpp
@@ -60,7 +60,7 @@ int DNSClient::inet_aton(const char* aIPAddrString, IPAddress& aResult)
     // See if we've been given a valid IP address
     const char* p =aIPAddrString;
     while (*p &&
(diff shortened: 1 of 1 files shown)


cmaglie:
@jhorck 
thanks for the patch!

This PR is a good fix for your use case, but there are other parsing problems.
I come with another solution in #3780, If you're still interested please follow up there.

(closing as superseded by #3780)

PR #2511, opened 2014-12-24
Transmit interrupt for UART and USART. Merge UART/USART

collin80:
Increased ringbuffer size to 128 for the Due. 

Merged UARTClass and USARTClass as much as possible to keep duplicate code to a minimum.

Implemented interrupt driven transmission over all hardware serial ports.

### DIFF
--- a/hardware/arduino/sam/cores/arduino/RingBuffer.cpp
+++ b/hardware/arduino/sam/cores/arduino/RingBuffer.cpp
@@ -21,7 +21,7 @@
 
 RingBuffer::RingBuffer( void )
 {
-    memset( _aucBuffer, 0, SERIAL_BUFFER_SIZE ) ;
+    memset( (void *)_aucBuffer, 0, SERIAL_BUFFER_SIZE ) ;
     _iHead=0 ;
     _iTail=0 ;
 }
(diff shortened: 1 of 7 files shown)


MorganSandercock:
Is UARTClass::flush() correct? This looks like it waits until the hardware 1-byte buffer is empty. It doesn't look at the ringbufffer. Can it see this empty before the interrupt or will the interrupt always be able to fill the ringbuffer into the transmit register without making it look empty?

mikaelpatel:
Great job extracting the ring buffer from the hardware serial driver. In Cosa I implemented something similar and added a twist. 

First Cosa IOBuffer as I call the ring buffer (https://github.com/mikaelpatel/Cosa/blob/master/cores/cosa/Cosa/IOBuffer.hh) is a template class which allows a more flexible configuration. Second the class implements the IOStream::Device interface which allow it to be used as a stream as well. The UART code becomes really easy. https://github.com/mikaelpatel/Cosa/blob/master/cores/cosa/Cosa/IOStream/Driver/UART.hh. This actually hides a lot of variability and complexity. 

Cheers!

collin80:
It's not worth the risk. I think that flush does need to explicitly ensure
that the TX buffer is fully flushed. I've moved around code in the end
method and placed the TX buffer check in flush. This should make things
work properly.

cmaglie:
Thanks!

I see that you defined some USART methods that simply calls the corresponding method of the base UART class, so I've removed them:
https://github.com/arduino/Arduino/commit/37ea166e1999e57eee6835786d3310c7ef8f2ec3

The class hierarchy should do exactly the same right?
(I've left only the begin method because I need it in another pull request that is coming...)

C

PR #2513, opened 2014-12-25
trivial: actually include the empty yield placeholder

pmjdebruijn:
fixing the following build failure:

In function `init':
/usr/share/gnoduino/hardware/arduino/cores/arduino/wiring.c:228: undefined reference to `yield'
collect2: error: ld returned 1 exit status

original commit introducing yield in master:
6ecb174c40fb880ac555182f918048b0050b3c98

### DIFF
--- a/hardware/arduino/cores/arduino/Arduino.h
+++ b/hardware/arduino/cores/arduino/Arduino.h
@@ -34,7 +34,7 @@
 extern "C"{
 #endif
 
-void yield(void);
+#include "hooks.c"
 
 #define HIGH 0x1
 #define LOW  0x0


cmaglie:
@pmjdebruijn 
I'm not able to reproduce the error, which version of the IDE are you using?
IMHO this patch is not correct and will lead to duplicate definitions of yield.

pmjdebruijn:
Right, as my error message betrays, I'm using a third party IDE, called Gnoduino, which is a lightweight PyGTK based IDE.

Which seemed fine until we pulled in the following commit into our local SDK:
6ecb174c40fb880ac555182f918048b0050b3c98

The reason why I came to the above conclusion, is that I couldn't find hooks.c included anywhere...

If this issue doesn't happen with the original Arduino SDK, I guess you can close this as INVALID. However I would appreciate it, if you could share any thoughts on what might be the issue?

cmaglie:
My best guess is that `hooks.c` for some reason is not compiled and linked in the final binary.

PR #2522, opened 2014-12-30
WS0010 OLED controller support

orangkucing:
1. WS0010 supported (4- and 8-bit transfers and JP font as well as w/
or w/o _rw_pin, and EU-I, Russia and EU-II fonts)
2. busy flag is checked if _rw_pin is specified

### DIFF
--- a/libraries/LiquidCrystal/LiquidCrystal.cpp
+++ b/libraries/LiquidCrystal/LiquidCrystal.cpp
@@ -5,6 +5,10 @@
 #include <inttypes.h>
 #include "Arduino.h"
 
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @orangkucing,
thanks for your work, but it's not possible to merge your request. 

The LiquidCrystal library is only targeted to HD44780 fully compatible displays, so support for something slightly different should be contained in a standalone library (which, by the way, already exists -> https://github.com/ladyada/Adafruit_CharacterOLED.git) .

Also, for the next pull request, make sure that the code can be used by anyone as-is, without the need to `#define` something in system libraries

PR #2536, opened 2015-01-05
Using LZMA as compression algorithm on linux

ffissore:
By making a tar.xz instead of tar.gz, the IDE dist file size goes down to ~60M from the original ~116M


ffissore:
Sorry, wrong target branch

PR #2566, opened 2015-01-22
windows binary product version within exe file

apla:
more info:
http://launch4j.sourceforge.net/docs.html#Ant_task
http://msdn.microsoft.com/en-us/library/windows/desktop/aa381058(v=vs.85).aspx

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -502,8 +502,8 @@
       <fileset dir="windows/launcher"
 	       includes="about.bmp, application.ico, config.xml, config_debug.xml"/>
     </copy>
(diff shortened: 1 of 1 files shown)


ffissore:
@ArduinoBot build this please

ffissore:
Tested and it fails to build because both config.xml and config_debug.xml are missing a `<versionInfo>` section.
I've pushed a preliminary fix at https://github.com/ffissore/Arduino/commit/006d56fc9fc7e32d8d54c354e630886a002d85d0

ffissore:
Closing because of reported issues. Reopen as they get fixed

PR #2597, opened 2015-02-02
Correct undefined Serial Buffer Size

NicoHood:
I split the definition to change the buffers individual. This also applies to the CDC Buffer.


cmaglie:
Thanks!

Issue #2571, opened 2015-01-23
Ethernet Library: DHCP Broadcast Spam


Elijahg:
This has been reported in the forum but I'd like to 'officially' report it here since it's seemingly gone ignored for 3 years, and the problem still exists.

When using Ethernet.maintain(), there are a few bugs that cause the Arduino to constantly spam the network with DHCP broadcast requests - not great for battery powered devices like phones that have to awaken to process all this data (around 50k/sec I was getting).

One bug was reported on the old tracker as issue #985. That's not fixed yet, but the new one was on the forum here: http://forum.arduino.cc/index.php?topic=195422.0 . Essentially half the DHCP reply is missing, and the Library relies on the DHCP server _not_ following the RFC and thus dishing out an existing address. DHCP servers that do follow the RFC hand out a new address, confusing things.

Both are fairly important as not to break people's networks, and fixes have been provided. It'd be a 2 minute job to fix.

I've checked the source for 1.5.x and it's not fixed there, nor in the live 1.0.6 release.

![screen shot 2015-01-23 at 19 20 04](https://cloud.githubusercontent.com/assets/1248059/5880996/e2ae04c0-a334-11e4-9dc7-58704c5adc97.png)

chaveiro:
See my comments on this before merging : https://github.com/arduino/Arduino/commit/bde3d715bd25a2ce16f84c43d93e4acb77c16f23#comments

cmaglie:
A proposed fix is in #3287 please follow up there.

Issue #985, opened 2012-11-15
DHCP infinite lease + ethernet.maintain() [imported]


cmaglie:
This is [Issue 985](http://code.google.com/p/arduino/issues/detail?id=985) moved from a Google Code project.
Added by 2012-07-16T13:09:48.000Z by [mik...@gmail.com](http://code.google.com/u/107003327038578600787/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium, Component-Ethernet

### Original description

**What steps will reproduce the problem?**
Ethernet shield + DHCP. If the DHCP server gives you lease with "infinite"  lease time (0xFFFFFFFF) the method calculating rebind time (_dhcpT2 = _dhcpT1 << 1;) gives you wrong time forcing Ethernet.maintain() to rebind every time it's called.

**What is the expected output? What do you see instead?**
PROBLEM ROW #104:_dhcpT2 = _dhcpT1 << 1;
I'm expecting positive integer but I'm seeing -2. Calculation method gives 2147483647 for renew timer(_dhcpT1) and -2 for rebind timer(_dhcpT2). Ethernet.maintain() sees this like it would be always 2 seconds "late" and forces it to rebind. This creates huge amount of traffic inside LAN as DHCP uses broadcast messages for request etc.

If I'm correct the calculation method should be something else for example the one I used in my scenario (see below).

**What version of the Arduino software are you using? On what operating**
**system?  Which Arduino board are you using?**
IDE 1.0.1, Windows 7 SP1, Arduino Mega2560+ Ethernet shield with SD

**Please provide any additional information below.**
In my own scenario I fixed the problem by changing the row used to calculate rebind time to not use bitwise left shift. Instead the row is now like this: "_dhcpT2 = _dhcpLeaseTime*7/8;".

cmaglie:
A proposed fix is in #3287 please follow up there.

PR #2601, opened 2015-02-04
Fix long DHCP lease times 

chaveiro:
Fix for lease times > 7FFFFFFF seconds.
Forever lease time is unsigned long with value FFFFFFFF.
If lease time > 7FFFFFFF, calling Ethernet.maintain() would force a renew every time. That is wrong, overloads dhcp server and generates unnecessary network traffic.
Issue was described here: https://code.google.com/p/arduino/issues/detail?id=985


chaveiro:
This is related to https://github.com/arduino/Arduino/issues/985 and https://github.com/arduino/Arduino/issues/2571

Elijahg:
This does *not* fix #2571 entirely. It needs a fix in Dhcp.cpp too.

chaveiro:
Please be specific what you mean entirely. My fix is for DHCP lease time.

cmaglie:
This has been included in #3287 please follow up there.

PR #2605, opened 2015-02-06
GSM Library fix for Galileo

zorzano:
Adapted so the same library runs for Galileo. Tested also on 1.0.6 and 1.5.8.
NONE state is renamed GSM_NONE
When in Intel, the SoftwareSerial is not used, but Serial1 is called.
checkModemInput is included and executed to look for modem comms.
manageResponse gets a void implementation, so it compiles.
GSM3ShieldV1DataNetworkProvider::inet_aton uses an intermediate value to generate the IP address
A small change in WebClient that makes for us easier the automatic testing.

### DIFF
--- a/libraries/GSM/src/GSM3SoftSerial.cpp
+++ b/libraries/GSM/src/GSM3SoftSerial.cpp
@@ -31,19 +31,23 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 The latest version of this library can always be found at
 https://github.com/BlueVia/Official-Arduino
 */
-#include "GSM3SoftSerial.h"
-#include "GSM3IO.h"
-#include <avr/interrupt.h>
-#include <avr/pgmspace.h>
-#include "pins_arduino.h"
+#include <GSM3SoftSerial.h>
+#include <GSM3IO.h>
+#include <pins_arduino.h>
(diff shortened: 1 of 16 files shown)


libraries/GSM/src/GSM3SoftSerial.cpp
```diff
@@ -327,6 +357,62 @@ void GSM3SoftSerial::recv()
...
+	// If we find a line feed, we are at the end of a paragraph
+	// check!
+	
+	if (fullbuffer)
+	{
+		// And... go handle it!
+		if(mgr)
+			mgr->manageMsg(thisHead, cb.getTail());
+	}
+	else if(d==10)
```
facchinm: why is this needed?

facchinm:
Hi @zorzano , first of all thanks for the submission; however there are some issues to address before merging this patch.
* please don't remove public functions (eg. `getLocation`) because it will break existing sketches
* divide the patch into atomic commits with proper description, so the rationale behind every change can be understood by anyone reading the git log
* make sure no useless or debug code is committed (see note on `GSM3SoftSerial::recv()`)

If you could rebase your patch and push force over this branch (or create another PR) I'll be happy to review and eventually merge it!

facchinm:
Closing for lack of feedback. Reopen if needed

PR #2614, opened 2015-02-09
rename hex -> output

championswimmer:
using generic term 'output', because for samx32,
output is .bin format and not .hex

Signed-off-by: Arnav Gupta <championswimmer@gmail.com>


championswimmer:
@cmaglie 

as requested, on a separate pull req

PR #2618, opened 2015-02-10
EthernetClient == and != operators

### LINKED ISSUE #2611: Official EthernetClient Sample Not Compiling - Ambiguous overload error with latest Arduino 1.0.6
The code sample below - the official Arduino sample code.
will not compile:

        http://arduino.cc/en/Reference/ServerAvailable

My public code libraries use the EthernetClient library like this - and same method I used is mentioned in several books and several ethernet sources - but it will not compile anymore on 1.0.6

(I have tried official 1.0.6 and the nightly build ARDUINO 1.6.0rc3 - 2015.02.03)

The problem is EthernetClient cannot be tested for true/false anymore (it used to have a overloading that allowed you to query it being ready by using boolean comparitor). 

The sample code has not worked since 1.0.5

Official Sample Code: 

  EthernetClient client = server.available();
  if (client == true) {  //compiler error here
  }

What is the correct new way to do this test now - and will the documentation be changed or the IDE/Compiler be changed back (best option)?

Thanks very much.

ffissore:
Added `virtual bool operator==(const bool value)` and `virtual bool operator!=(const bool value)`.
Fixes #2611
Bug introduced with #1700
@cmaglie @matthijskooijman you're far more expert than me with c++

### DIFF
--- a/libraries/Ethernet/src/EthernetClient.h
+++ b/libraries/Ethernet/src/EthernetClient.h
@@ -24,6 +24,8 @@ public:
   virtual void stop();
   virtual uint8_t connected();
   virtual operator bool();
+  virtual bool operator==(const bool value) { return bool() == value; }
+  virtual bool operator!=(const bool value) { return bool() != value; }
   virtual bool operator==(const EthernetClient&);
   virtual bool operator!=(const EthernetClient& rhs) { return !this->operator==(rhs); };
 


ffissore:
@ArduinoBot build this please

OzmoOzmo:
Thanks very much - that does indeed sort that problem. Can close.

Sorry not getting back to you sooner - but I was trying to get the whole project compiling to fully test it- - having issues with lots of memory related functions now depreciated that I'm still working through
const prog_uchar allmonths[] = {"JANFEB.."} etc. 

Thanks for getting EtherClient working anyway :)

PR #2623, opened 2015-02-11
update chinese translation

coloz:
Previous is funny,for example, programmer is a tool，but ‘程序员’ is a people


ffissore:
Thanks for the PR @coloz . Arduino translations are managed with transifex: https://www.transifex.com/projects/p/arduino-ide-15/ We regularly poll from transifex and update translations. Please join the translators group on transifex

Issue #2576, opened 2015-01-25
cannot find YUN  in 1.6.0   


jl-p:
Hi,

I have a Yun connected to my network and running 1.5.8 and 1.6 on the same laptop
when i start the 1.5.8 version it detects my yun. 
when i start the 1.6.0 version it won't 

Johan

TrippyLighting:
Did you mean to say 1.0.6 ?
That IDE does not support the Yún. The Due and the Yun are t main reasons behind the 1.5.x IDE

ffissore:
What operating system are you using?

jl-p:
windows 8.
And no 1.6.0 nightly build.

ffissore:
nightly builds are not signed. Could this be the source of the problem?

jl-p:
I Know.
But if no one reports it. it will be in the final version also.

ffissore:
If the source of the problem is the missing signature, then no: the final release will be signed

NicoHood:
I played with my yun shield a few days ago and noticed that the board appeared and disappeard sometimes very bad. Not sure if this is a 1.6 bug since ive never tested it with 1.5.8. But I also noticed some problems with the connection.

ffissore:
@jl-p 1.6.0 has been released. Have you already tried it?

roadfun:
I was also having several problems with Yun discovery (I have 3 Yuns and my issues started prior to 1.6). I did some debugging in the IDE and found some problematic code. I submitted a code change that has resolved these issues for me and now the Discovery is very reliable. Hopefully the fix is a good one and will then help others. I've also seen posts about this problem on the Arduino Yun forum.

agdl:
@jl-p & @ffissore reopen it if needed

ffissore:
Indeed fixed (or at least mitigated) in 1.6.1

PR #2646, opened 2015-02-16
Fix for #2576

roadfun:
I’ve had a couple issues with unreliable Yun discovery. This change has
fixed those issue for me and would appear to address concerns raised in
#2576.

### DIFF
--- a/arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
+++ b/arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
@@ -62,12 +62,19 @@ public class NetworkDiscovery implements Discovery, ServiceListener, cc.arduino.
     Iterator<BoardPort> iterator = ports.iterator();
     while (iterator.hasNext()) {
       try {
-        BoardPort board = iterator.next();
-        if (!NetUtils.isReachable(InetAddress.getByName(board.getAddress()), Integer.parseInt(board.getPrefs().get("port")))) {
-          iterator.remove();
-        }
+          BoardPort board = iterator.next();
+          InetAddress address= InetAddress.getByName(board.getAddress());
+          boolean reachable = false;
+          try {
+              reachable = address.isReachable(100);
+          } catch (IOException e) {
+              // ignore (reachable is false by default)
+          }
+          if (!reachable) {
+              iterator.remove();
+          }
       } catch (UnknownHostException e) {
-        iterator.remove();
+          iterator.remove();
       }
     }
     return ports;


ffissore:
Hi @roadfun 
I see you're using `InetAddress.isReachable()`. Unfortunately, that won't work. IsReachable does not behave the same way on all platforms and in general is not reliable.
Google for `inetaddress.isreachable problem` to see some interesting reports. Here is one http://bordet.blogspot.it/2006/07/icmp-and-inetaddressisreachable.html

Still, I admit using advertised `port` 80 is not be a smart move, because it's not the one used by the IDE. We should add port 22 to the advertised ports

ffissore:
Or we can just use the `ping` command and let the OS deal with ICMP. Ping command line is not equal across different OSs, so we'll have to define a method for each `Platform` object

roadfun:
Hi Federico. I did also consider those two alternatives. A couple of observations.

First, there is something wrong with the NetUtils implementation beyond the port being used which is why I didn't change it to use the port 22 for example. If I use NetUtils.isReachable() Yun discovery remains unreliable. At times the IDE will not find any Yuns at all (as reported by the OP and experienced by myself and others). With the implementation I submitted it works perfectly.  

I did consider doing a cross-platform call to ping but I rejected that for two reasons. First, like your claim about InetAddress.isReachable() the implementation on each platform is different. Windows does not support a true ping but instead uses TCP on a specific port. If you set up the Yun to be more secure it will not respond to echo. And in fact when I lock down a system it won't respond to ping either (could be an extreme case). Even the article you linked to on Blogspot says ping will often fail.

I think unreliable checks for isReachable() are worse than no check at all (false positives). I'd much rather have false positives than have none of my Yuns showing up in the Ports menu or have them disappearing randomly. If checks are deemed essential then I think a reliable implementation mimicking an upload connection is the only reasonable implementation. 

Since I experience both of these problems constantly using the official IDE I will spend a little more time to see if I can find an approach to checking the port availability that won't suffer the issue of the current method. Feel free to share any other thoughts or advice.

ffissore:
Thanks for all your feedback @roadfun. I agree with you: ping is not reliable, not even when you run the one provided by the OS, especially if you've hardened your yun. And I also agree that attempting an upload is the best way to know if the actual upload will work.

The yun uses SSH, other future boards may use something else (who knows). We should add some metadata to yun's bonjour message to tell the IDE which test to perform.

I'll do some tests with a yun tomorrow morning, but I'm eager to share some code: please take a look at https://github.com/ffissore/Arduino/commit/abc376bd941da0b6f3265024967b5763613be06d

Variables naming apart, it attempts to establish a connection on both port 80 and port 22. Timeout is raised to 300ms: that will slow down the Ports menu, but the yun may be slow at responding.

I'll also try to make that check asynchronous, so that we could raise timeout even to a second

roadfun:
Hi Federico. I left a comment on your code. With one change I think it works. I think there is an additional change which could be an improvement and that is to refactor to stop on the first isReachable success since you are checking multiple ports. In that case there is no need to keep checking is there?

ffissore:
Thanks @roadfun. Fixed and opened PR #2658

ffissore:
Code merged into #2658. Closing

PR #2648, opened 2015-02-16
"objcopy" recipes are no more tied to "hex" extension

cmaglie:
This allow to define specific recipes for every platform but keeps backward compatibility.
Related to #2614

/cc @ffissore @championswimmer


ffissore:
:+1: looks good

championswimmer:
Makes sense

I'll need to modify the save output patch a little to comply with this.
I'll refresh that pull req

Arnav Gupta
( http://championswimmer.in )
Sent from Android.
Pardon brevity and/or autocorrect errors.

PR #2658, opened 2015-02-17
An attempt to improve Yun's discovery

ffissore:
Thanks @roadfun. See #2576

/cc @cmaglie

### DIFF
--- a/arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
+++ b/arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
@@ -58,19 +58,35 @@ public class NetworkDiscovery implements Discovery, ServiceListener, cc.arduino.
 
   @Override
   public List<BoardPort> discovery() {
-    List<BoardPort> ports = clonePortsList();
-    Iterator<BoardPort> iterator = ports.iterator();
-    while (iterator.hasNext()) {
+    List<BoardPort> boardPorts = clonePortsList();
+    Iterator<BoardPort> boardPortIterator = boardPorts.iterator();
+    while (boardPortIterator.hasNext()) {
       try {
-        BoardPort board = iterator.next();
-        if (!NetUtils.isReachable(InetAddress.getByName(board.getAddress()), Integer.parseInt(board.getPrefs().get("port")))) {
-          iterator.remove();
+        BoardPort board = boardPortIterator.next();
+
+        InetAddress inetAddress = InetAddress.getByName(board.getAddress());
+        int broadcastedPort = Integer.valueOf(board.getPrefs().get("port"));
+
+        List<Integer> ports = new LinkedList<Integer>();
+        ports.add(broadcastedPort);
+
+        //dirty code: allows non up to date yuns to be discovered. Newer yuns will broadcast port 22
+        if (broadcastedPort == 80) {
+          ports.add(0, 22);
+        }
+
+        boolean reachable = false;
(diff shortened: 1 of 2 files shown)


arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
```diff
@@ -58,19 +58,35 @@ public NetworkDiscovery() {
...
+        ports.add(broadcastedPort);
+
+        //dirty code: allows non up to date yuns to be discovered. Newer yuns will broadcast port 22
+        if (broadcastedPort == 80) {
+          ports.add(0, 22);
+        }
+
+        boolean reachable = false;
+        for (Integer port : ports) {
+          reachable = reachable || NetUtils.isReachable(inetAddress, port);
```
cmaglie: This loop always check for 80 and 22, it may be rewritten as:

```java
reachable = false;
for (Integer port : ports) {
  if (NetUtils.isReachable(inetAddress, port)) {
    reachable = true;
    break;
  }
}
```

so it stops at the first successful check

roadfun:
Tried both "macosx" download links. When I unzip the files in both cases I get a mess that Arduino has been damaged and should be moved to the trash. So I can't test this build.

arduino-core/src/cc/arduino/packages/discoverers/NetworkDiscovery.java
ffissore: After @roadfun comment, I've changed `&&` to a `||`. This way the loop will spin but it won't run `isReachable` any more as soon as one returns `true`

ffissore:
I can download and unpack it successfully but I'll trigger another build

ffissore:
@ArduinoBot build this please

ffissore:
Since InetAddress.isReachable is reported to be reliable on mac, I've merged both ways into NetUtils (see https://github.com/ffissore/Arduino/commit/0990f98b1430bdd79022d48981f91ddf354bae78)
Now, first attempt is done by `isReachableByEcho`, second one by opening ports with `isPortOpen`

roadfun:
Sorry Federico but this doesn't work for me either. I checked the Console log and launching the app gives an error I've never seen before, and a quick web search didn't turn up any clue. I'd have to do some digging to figure out why it doesn't work if it works for you. Only thing I can guess is some sort of security/sandbox issue. But I don't have any problem downloading and running the official build.

2/19/15 8:41:30.753 AM CoreServicesUIAgent[2653]: Error -60005 creating authorization
2/19/15 8:41:30.817 AM CoreServicesUIAgent[2653]: Error SecAssessmentCreate: The operation couldn’t be completed. (OSStatus error -67030.)

ffissore:
Ah maybe it's because osx treat the app as unsecure because you've downloaded and it's not signed (automated builds are not signed, only releases are)
You should find some way to allow this app to run: I've once seen a temporary permission, but I can't point you to the right button, sorry

roadfun:
My usual trick for opening unsigned apps didn't work. I'll do some more searching.

roadfun:
Figured out the work around - in this case it requires changing the system setting to allow any app to run. The single-app exceptions doesn't work. Having solved that problem.

I think this build works as well as the version I created. I still some issues (e.g. sometimes no Yuns are shown at all) but those are common between your build and mine. Definitely this fixes the "port 80 not open" problem.

ffissore:
Good! Thanks for all your work and feedback :)

PR #2659, opened 2015-02-18
Refactor SPI begin to not impact SS pin un-necessarily

vicatcu:
Submitting this per Christian's request to be tested

### DIFF
--- a/hardware/arduino/avr/cores/arduino/wiring.c
+++ b/hardware/arduino/avr/cores/arduino/wiring.c
@@ -322,4 +322,8 @@ void init()
 #elif defined(UCSR0B)
 	UCSR0B = 0;
 #endif
+
+  // this line is for backwards compatibility with legacy examples
+  // that expect this to have been done in SPI.begin() prior to PR2659
+  digitalWrite(SS, HIGH);  
 }
--- a/hardware/arduino/avr/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/SPI.cpp
@@ -42,7 +42,7 @@ void SPIClass::begin()
     // When the SS pin is set as OUTPUT, it can be used as
     // a general purpose output port (it doesn't influence
     // SPI operations).
-    // pinMode(SS, OUTPUT); //an input with internal pullup enabled suffices too
+    pinMode(SS, OUTPUT); //an input with internal pullup enabled suffices too
 
     // Warning: if the SS pin ever becomes a LOW INPUT then SPI
     // automatically switches to Slave, so the data direction of
--- a/hardware/arduino/avr/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/SPI.cpp
@@ -42,7 +42,7 @@ void SPIClass::begin()
     // When the SS pin is set as OUTPUT, it can be used as
     // a general purpose output port (it doesn't influence
     // SPI operations).
-    pinMode(SS, OUTPUT); //an input with internal pullup enabled suffices too
+    pinMode(SS, OUTPUT);
 
     // Warning: if the SS pin ever becomes a LOW INPUT then SPI
     // automatically switches to Slave, so the data direction of
--- a/hardware/arduino/avr/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/SPI.cpp
@@ -34,7 +34,7 @@ void SPIClass::begin()
     volatile uint8_t *reg = portModeRegister(port);
 
     // if the SS pin is not already configured as an output
-    // then set it high (to enable the internall pull-up resistor)
+    // then set it high (to enable the internal pull-up resistor)
     if(!(*reg & bit)){
       digitalWrite(SS, HIGH);
     }
--- a/hardware/arduino/avr/libraries/SPI/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/SPI.cpp
@@ -34,7 +34,7 @@ void SPIClass::begin()
     volatile uint8_t *reg = portModeRegister(port);
 
     // if the SS pin is not already configured as an output
-    // then set it high
+    // then set it high (to enable the internall pull-up resistor)
     if(!(*reg & bit)){
       digitalWrite(SS, HIGH);
     }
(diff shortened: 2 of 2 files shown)


cmaglie:
@ArduinoBot build this please

facchinm:
Hi @vicatcu , I read the thread on the forum and tested your patchset on some hardware. 

At the end, I must say that I'm not convinced at all. If the switch between slave and master mode could only occur during `init` it would be the perfect solution.

However, the switch can happen at any time and, if the pin is left untouched by the user, sooner or later the SPI peripheral could "disappear" without any hint of what happened (internal pullups are not enough strong to beat handling sometimes)

Also, in all examples which don't use hardware SS pin (SD card etc) it is clearly stated that the pin must be an OUTPUT, so I really don't see the point in shipping a (possibly) broken configuration as the default one. 

So, thumbs up for the removal of the unnecessary `digitalWrite(SS, HIGH)` if the pin is already an output, but I'd prefer to keep the original behaviour if the pin is left untouched.

vicatcu:
I didn't have strong feelings about commenting out pinMode(SS, OUTPUT). I only did it based on feedback from @cmaglie and others on the dev list. That being said I think you're juts speculating. Can you actually demonstrate a reproducible problem with the proposed implementation?

facchinm:
I had this feeling and I spent some hours this morning to test it and, indeed, I experienced some hangs with an SPI DAC. 
I must admit that I had to force myself to perform all the worst practices I could imagine :grin:, but at the end handling the board without paying much attention is what most hobbyist do.

vicatcu:
@facchinm, just checking... if you uncomment the digitalWrite(SS, HIGH) and use the same kind of testing your SPI DAC doesn't hang?

facchinm:
Yep, it never hung during all my tests (whether the pin was forced low or high)

vicatcu:
@facchinm I've reverted the pinMode change per your testing feedback. Thanks, I appreciate the scrutiny!

facchinm:
Perfect, thank you very much.
I'm merging your commits squashed into one (to not spam git history)

facchinm:
Reopening the PR to better handle issue #2868
Unfortunately the provided examples for the SD library set pin 10 as output (implicitly `LOW`) before `SD.init()`, so an Ethernet shield will have both peripherals' chip select low and the SD will not respond.

The bug is not visible
- if no peripheral has its CS on pin 10 or
- if the pin 10 is also set `HIGH` before `SD.init()` or
- if the pin 10 is not set as `OUTPUT` before `SD.init()`

However, we can't know how many sketches have been based on the examples, so we have to consider it a regression and not only fix the examples but also revert the commit (if no one has a better idea :smile: )

I'll leave it open for a couple of days to gather ideas, if nothing happens we'll revert 0e898d5e0 and fix the examples.

vicatcu:
@facchinm I apologize for having inadvertently caused you regression grief. I still think this issue is a real issue (i.e. not just cosmetic). Multiple calls to SPI.begin() in a sketch have unwanted/unexpected side effects (impacting the state of the SS pin even if it's not used). 

I'll offer an alternative. Put the `digitalWrite(SS, HIGH)` in the `wiring.c` at the end of the `init()` function (which gets called before `setup()` in `main.cpp`) with a note as to why it's being done in comments. That would make this PR compatible, as is, with issue #2868. The impact would be:

1. Any sketch that does nothing with the SS pin would have that pin configured as an input with internal pull-up enabled, whereas previously it would have been configured as a floating input.
2.  Any sketch that uses the SS pin would over-ride the default settings anyway.
3. The CardInfo example, anti-pattern though it might be, would still work, because of the default pin state applied by the proposed change to `init()`.

No matter what we do here, it feels like we're sweeping some dirt under the carpet, but this seems like a plausible low-impact approach. I've made the change and pushed it to my fork for your consideration.

vicatcu:
On an unrelated note, it might not be a terrible idea to fix the examples *anyway*...

facchinm:
Hi @vicatcu , I spent some time thinking about the possible consequences of all the approaches. 

I like your patch but if we apply it we are going to change the behaviour _again_, this time  for sketches which uses pin10 but without SPI (probably no one will notice it but it's full of strange setups out there)

At the end, probably the best thing to do is simply fixing the examples;
we need to understand if doing it by hiding the pin's peculiar nature (as it is already handled by the lower init code) or by forcing it high.

By the way, thanks a lot for your efforts!

matthijskooijman:
I'm a bit confused by the current code, @facchinm, perhaps you could clarify? AFAICS the intention of this PR was to enable the pullup on  SS instead of switching it to output, since that should be sufficient to keep SPI at master mode. However, IIUC, you have tested and found that a pullup on a floating pin is not sufficient and sometimes causes SPI slave mode to trigger, right?

So, the pin is always configured as an output like before. But what's the point of the "enable pullup" code now? Its comment is certainly confusing and AFAICS, the behaviour is now that:
 - if the pin is OUTPUT, it is unchanged
 - if the pin is INPUT, it is configured as OUTPUT HIGH

Is that really the intention?

cmaglie:
The original thread in devlist is here:
https://groups.google.com/a/arduino.cc/d/msg/developers/ZA7emDC8GrA/ySzdGyDq_sEJ

The main reason behind this change is to preserve the state of pin 10 when it's used as general purpose output (i.e. it's not used as CS but to control something else).

facchinm:
@matthijskooijman , as you perfectly stated the original intention was to leave the pin as-is except if it was configured as INPUT (FLOATING). But pullups are not enough strong to keep HIGH value while touching the board and we decided change the patch.

The order of the events (`digitalWrite` before `pinMode`) prevents unexpected changes of state (calling pinMode(OUTPUT) directly set pin LOW)

So yes, it was fully intentional :smile:

vicatcu:
@matthijskooijman The principle idea here was that if someone already turned the SS pin into an output, the SS pin should be left alone, because as an output it doesn't interact with the internal SPI peripheral hardware, and the pin can be used for general purpose output. The remainder of the code (order of events) was implemented for backwards compatibility basically, with the intent being that the SS pin should be pulled high if SPI is enabled and SS is an previously configured as an input (otherwise you can end up getting unexpected interrupts as a SPI slave).

matthijskooijman:
Ok, that seems to make sense to me. Though I wonder why the pin is configured as output high and not output low, but I guess that's the backward compatibility part?

I'm surprised that enabling the pullup doesn't provide enough pull to keep the pin high, though if it doesn't, the pin should indeed be an output.

Would it make sense to change the code to make the above intentions a bit more clear? The current code and comments are confusing. AFAICS the pinMode can just be moved inside the if, and the comments should be updated (which should not change functionality, but make the code significantly more clear).

barsznica:
Related to [this](https://github.com/arduino/Arduino/commit/d101bf51a2c9ebe08be737dd9b5db8b32d64e44a)?

PR #2666, opened 2015-02-19
Undo sets the sketch to modified, even if it's not

### LINKED ISSUE #2653: 1.6.0 IDE sets "file modified" flag whenever a tab is clicked for viewing, instead of when the content is actually changed
Context:
* MacOS
* Arduino IDE 1.6.0

To reproduce:
* Start with a multi-tab sketch
* Save it (ensures all "file modified" state/markings are cleared for this example...)
* Click on another tab but do not modify its content
* Notice the tab now shows the "file modified" squiggle even though nothing was done to modify the content of the tab

IMO, the "modified" state should track the changelog/undo buffer, such that "undo"ing all edits clears the modified flag and the modified state only is set after a true content modification.  The current implementation seems to be much less nuanced :-)

ffissore:
Fixes #2653
/cc @plocher @cmaglie

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1365,6 +1365,7 @@ public class Editor extends JFrame implements RunnerListener {
     public void actionPerformed(ActionEvent e) {
       try {
         undo.undo();
+        sketch.setModified(true);
       } catch (CannotUndoException ex) {
         //System.out.println("Unable to undo: " + ex);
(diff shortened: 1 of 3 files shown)


ffissore:
Wow, a test fails. Need to fix it

plocher:
Verified on original multi-pane sketch, this build does not exhibit the
original problem.

Thanks!

Issue #2591, opened 2015-02-01
bug in Stream find


MauroMombelli:
IF the buffer contains "1211121110" and you are looking for "121110" with actual findUntil() implementation (all all function using it, like find() ) you'll get a false!

Xuth:
I just found this bug myself.  To be able to handle this the find function would need to either allocate a buffer which is problematic to do behind the scenes in a low memory environment or to require the user to pass in a buffer (which contradicts the "simple" aspect of the arduino libraries).  But without this, the functions genuinely are broken.  I ended up writing my own find routine that is more functional and requires the user pass in a buffer but it's more complicated to call than the arduino libraries.

matthijskooijman:
Can you clarify why it needs a buffer? It would just need to backtrack to the start of a failed find attempt, right?

Xuth:
Well, to backtrack it would need as long a buffer as the failed find attempt.  However as I think about it more, we already effectively have that buffer since, if we've gone N characters into the target word, we know what the last N characters were and can use that to "look back into the buffer" to find a new starting point.

After looking through the Stream library (since the documentation on the Arduino site was insufficient to actually write code doing things like describing target of the find functions as a char rather than a char * and _lots_ of other details) and seeing this as the most egregious bug and a couple more minor bugs while I was at it, I was thinking about refactoring the module and maybe some of the documentation and figuring out how to get someone to look at it for inclusion here.

MauroMombelli:
well a solution may be to use one variable with is increased until you find the first letter and the index of the number of letter matched is > 0 (in my example the first value should be the index 2). That way we keep track of where it COULD start the next recursive sequence. if the match fail, instead of continuing from the current position, as you do now, you'll have to fall back to that stored value.

do you like it?i'm planning to write down some code when i'll have time

Xuth:
You can actually do better than that and you can do so with minimal variables on the stack.

peabo7:
> On February 22, 2015 at 11:49 AM Xuth <notifications@github.com> wrote:
> 
> 
> Well, to backtrack it would need as long a buffer as the failed find attempt.
>  However as I think about it more, we already effectively have that buffer
> since, if we've gone N characters into the target word, we know what the last
> N characters were and can use that to "look back into the buffer" to find a
> new starting point.

You might like to look at the KMP algorithm:

http://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm

[...]

PR #2696, opened 2015-02-26
Fix findUntil and general timeouts in Stream library

Xuth:
This is my proposed fix to issue 2591 https://github.com/arduino/Arduino/issues/2591 .  Primarily this fixes two things:  The most important is that searching for something like "1112" in the stream "11112" now works.  The second thing is that timeouts on stream reads are properly respected.  In the previous version of the code, if you set a timeout of 1000 millisecs and your stream produced a single character every 500 milliseconds, a read would never time out since the clock would be reset every time the program entered timed read.  I had to add a handful of private methods to keep all of the other methods intact but I believe that all of the public and protected methods are maintained as before minus the two bugs I described.  I also added a low level find routine that would support an arbitrary number of target strings since it's not uncommon to be searching for more than one good string and one error string indistinguishable in the interface from a timeout.  This last routine, findMulti() is not very Arduino library like (requiring you to pass in an array of structs) but you need one integer of scratch space per find target* and I felt that dynamically allocating that in the function was a greater sin with only 2k of ram.  Everything else uses a constant amount of stack space.

* Alternately a different algorithm could be used that requires a buffer space at least as large as the largest search target.  When searching for lots of different short strings this other algorithm would be more space efficient.  It's possible to write an routine that will do this in constant space and I'm sure I could write it but it is ugly and more inefficient than what I've written here to get away with only using one extra int per search target.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Stream.cpp
+++ b/hardware/arduino/avr/cores/arduino/Stream.cpp
@@ -18,6 +18,8 @@
 
  Created July 2011
  parsing functions based on TextFinder library by Michael Margolis
+
+ findMulti/findUntil routines written by Jim Leonard/Xuth
  */
 
 #include "Arduino.h"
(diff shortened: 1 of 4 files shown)


onmicrosoftclub:
Fix

damellis:
I think the original idea was in fact for the timeout to be per-character, not total. (Regardless, the current behavior should probably be better explained in the documentation: http://arduino.cc/en/Reference/StreamSetTimeout). An overall timeout is not a bad idea, but it would be a significant change from the current behavior. Maybe it could be an additional setting?

Xuth:
If that was genuinely the intent, that is a) not at all conveyed well in the documentation and b) would seem to have fairly limited utility.  Making it so that the basic interface provides no means of detecting a device that is misbehaving (ie generating spurious characters because the rx line is floating) seems like awful design.  I would be shocked if this were genuinely the original design goal, just that the current implementation was "good enough" to make things mostly work and nobody came back to make them better.

I agree that this is a significant change, but based on the documentation and the usage I found online trying to figure out how to use the library, I'd be genuinely surprised if many people thought that the timeouts work the way they currently do.

If I'm genuinely wrong about the original intent, the documentation needs to be made very clear of this.  That said, the documentation needs to be made much clearer about lots of things.  I only looked at the code in the first place because the docs were insufficient and in places gave incorrect information about the parameters of the various methods (treating strings/char *'s interchangeably with chars).  Thus once I get this bug fix settled I want to go over the stream documentation and suggest a bunch of fixes but I want to know specifically what the methods are all supposed to do and make sure that they're actually doing them first.

facchinm:
Hi @Xuth ,
first of all thank you for the patch, very clever implementation and so small overhead.

I'd like to merge it as-is but the "global timer" modifications are not directly related to the topic, and will need a lot more discussion (also on the mailing list).

So I rebased only the findUntil patch here facchinm/Arduino@f43a7a698 , tested it and it works like a charm so, if it's ok for you, I'd merge only that part and close this PR

Xuth:
Thanks, this looks good.  (I don't have that much pride tied up my code.  i'm mostly coming at this as a user wanting to be able to use this for my own needs)  I need to write up some comments for the mailing list on the timeout issue and also (after using the library a bit more) realizing that at times I wanted access to the timedRead() routine and see no reason why that and the timedPeek() routines shouldn't be made public rather than protected.

facchinm:
Perfect, waiting for you in the mailing list :wink:

Issue #2655, opened 2015-02-16
"Save as" should default to parent of current sketch folder.


damellis:
On Mac at least, when you select "Save as..." from the File menu, the dialog box has the sketch's folder selected by default. That means if you simply type a new name and press enter, you'll get an error about trying to save a copy of the sketch inside itself ("How very Borges of you").

It would probably make more sense for the dialog box to default to the parent of the current sketch folder instead, i.e. default to saving the copy next to the original instead of inside it.

This is with 1.6 (the recommended version) on Mac OS X 10.10.2.

ivovdgrift:
It seems to only do this on mac, not on windows. I'll try to create a fix for macs

ivovdgrift:
Offered a solution [here](https://github.com/arduino/Arduino/pull/2716)

ffissore:
Fixed with #2716

PR #2716, opened 2015-02-27
Fix Saveas method to prevent saving sketch into itself error

ivovdgrift:
Simple fix that should solve issue [#2655](https://github.com/arduino/Arduino/issues/2655)
Added .getParentFile() to the saveas method, such that it actually uses the parent folder of the sketch, and no longer the sketch itself. Only does this for macs, as the problem does not occur on windows computers.
Prevents the saving sketch into itself error


ffissore:
@ArduinoBot build this please

ffissore:
Thanks. I've added a commit since the problem affects all supported OSs, not just mac

PR #2726, opened 2015-03-02
TemperatureWebPanel.ino Print "Yún" and not "Y�n", "degrees C" to "ºC"

hashashin:
Hi, just got a Yún few days ago and noticed that when testing stuff, just cosmetic.

Thanks.


ffissore:
Thank you!

PR #2636, opened 2015-02-12
New preprocessor

ffissore:
This PR introduces a new preprocessor, based on two binary tools: [coan](http://coan2.sourceforge.net/) and (a modified version of) [ctags](https://github.com/ffissore/ctags)

coan is use to resolve #define #ifdef et al, producing a clean sketch.
If your sketch has
```
#define DEBUG 0
#if DEBUG
#include "debug.h"
myDebugFunction("hello debug");
#endif
```
the resulting sketch will be
```
#define DEBUG 0




```

ctags is used to collect function definitions and to generate their prototypes. If a prototype is already defined, it will not be generated.
If your sketch has
```
void firstFunction() {
  // code here
}
void secondFunction();
void secondFunction() {
  //code here
}
```
the resulting sketch will be
```
void firstFunction();
void firstFunction() {
  // code here
}
void secondFunction();
void secondFunction() {
  //code here
}
```

coan and ctags are defined in a parent platform.txt (currently located in IDE_FOLDER/hardware)

A new mechanism for loading platform.txt allows to group key value pairs common to multiple platforms or packages into separate files. Those pairs will be merged into children platform.txt. The key value pair nearest to boards.txt wins and overwrites pairs with the same key.
An example is IDE_FOLDER/hardware/arduino/sam/platform.txt, where coan command line is customized with sam specific parameters.
This keep users that have defined a 3rd party platform based on the avr one from having to update it adding the new tools definitions, and lowers the burden of those that started from the sam one.

This PR fixes lots of issues: see https://github.com/arduino/Arduino/labels/Component%3A%20Preprocessor

PR #2729, opened 2015-03-04
New preprocessor

ffissore:
This PR introduces a new preprocessor, based on two binary tools: [coan](http://coan2.sourceforge.net/) and (a modified version of) [ctags](https://github.com/ffissore/ctags)

coan is use to resolve #define #ifdef et al, producing a clean sketch.
If your sketch has
```
#define DEBUG 0
#if DEBUG
#include "debug.h"
myDebugFunction("hello debug");
#endif
```
the resulting sketch will be
```
#define DEBUG 0




```

ctags is used to collect function definitions and to generate their prototypes. If a prototype is already defined, it will not be generated.
If your sketch has
```
void firstFunction() {
  // code here
}
void secondFunction();
void secondFunction() {
  //code here
}
```
the resulting sketch will be
```
void firstFunction();
void firstFunction() {
  // code here
}
void secondFunction();
void secondFunction() {
  //code here
}
```

coan and ctags are defined in a parent platform.txt (currently located in IDE_FOLDER/hardware)

A new mechanism for loading platform.txt allows to group key value pairs common to multiple platforms or packages into separate files. Those pairs will be merged into children platform.txt. The key value pair nearest to boards.txt wins and overwrites pairs with the same key.
An example is IDE_FOLDER/hardware/arduino/sam/platform.txt, where coan command line is customized with sam specific parameters.
This keep users that have defined a 3rd party platform based on the avr one from having to update it adding the new tools definitions, and lowers the burden of those that started from the sam one.

This PR fixes lots of issues: see https://github.com/arduino/Arduino/labels/Component%3A%20Preprocessor

Closes #2636


Wackerbarth:
There is a problem with the function prototype generation.
```
byte program (const byte b1, const byte b2 = 0, const byte b3 = 0, const byte b4 = 0)
  {
   /* ... */
  return b4;
  } // end of program
```
generates the prototype
```
byte program(const byte b1, const byte b2 = 0, const byte b3 = 0, const byte b4 = 0);
```
however that causes the compiler to fail
```
Atmega_Board_Detector.ino: In function 'byte program(byte, byte, byte, byte)':
Atmega_Board_Detector.ino:275:85: error: default argument given for parameter 2 of 'byte program(byte, byte, byte, byte)' [-fpermissive]
Atmega_Board_Detector.ino:265:6: error: after previous specification in 'byte program(byte, byte, byte, byte)' [-fpermissive]
```

ffissore:
What is the correct prototype?

mastrolinux:
`byte program(byte, byte, byte, byte)`

Is the correct prototype.

mastrolinux:
or with "const" if you like but without the default arguments

ffissore:
Please ignore previous (deleted) build message. A new build with a fix is being backed

ffissore:
Ok I've fixed the issue. Given function
```
byte program (const byte b1, const byte b2 = 0, const byte b3 = 0, const byte b4 = 0)
  {
   /* ... */
  return b4;
  } // end of program
```

the correct preprocessor result should have the prototype specify the default values and the implemention just mention the var names, like

```
byte program (const byte b1, const byte b2 = 0, const byte b3 = 0, const byte b4 = 0);

byte program (const byte b1, const byte b2, const byte b3, const byte b4)
  {
   /* ... */
  return b4;
  } // end of program
```

However this would mean modify your code and we will **NOT** do that.

Instead, we now generate this code

```
byte program (const byte b1, const byte b2, const byte b3, const byte b4)

byte program (const byte b1, const byte b2 = 0, const byte b3 = 0, const byte b4 = 0)
  {
   /* ... */
  return b4;
  } // end of program
```
Where default values are in the implementation, not in the prototype. This is not correct, but compiles and works fine.

If you want to have the correct version, you need to type the prototype yourself: this new preprocessor will **not** generate prototypes when they are already specified.

ffissore:
I've just found out it doesn't work if you have a configuration #define into a file included by the sketch and your sketch checks it with #ifdef. Coan won't read the included file and hence the #ifdef will result to false. I'm going to fix this in the upcoming days

supuflounder:
OK, I downloaded the PR-2729-BUILD-219-macosx.zip, which contained only the Arduino executable.  I moved this to the applications folder and did a bit of renaming so it is now "Arduino" and the previous one is "Arduino-1.5-r2".  But now, it doesn't even have all the boards the previous version had.  I need to support the Adafruit Trinket Pro 5V (FTDI), and the other support, for the plain Trinket, the Trinket Pro 3V, and the Gemma would be nice.  I've got a question out in the Adafruit forum regarding this, but I find that there is a lot of guesswork involved in trying to understand the platform.txt file and the boards.txt file.  Is there a document anywhere that defines these files and how they all cohabit my development space?

supuflounder:
Would it help at all if Arduino had an option to either write a log file as it is starting up, or produce a file which is its representation for boards, boot loaders, and platforms?  Then I could post the file, and somebody could say "OH, the reason your definitions are not being seen is that..." instead of my having to do "black box" experiments trying to figure out what is going on.  The document in the wiki is only of marginal help.  I have spent three days trying to figure out how to program my Trinket Pro, with no success.  In fact, I'm further behind now than I was; under the 1.5-r2 release, it could see the board types and compile, but the upload failed.  Now, I can't even find the boards.  The Adafruit support keeps putting me into the middle of circular link sequences, where each node has either incomplete or apparently conflicting information.  I have tried every variant described in every link, and get NOTHING!

What I have done is take a zip file which contains boards.txt, bootloaders.txt, and platforms.txt, and installed it in

mylogin/documents/Arduino/hardware/avr

with no happiness.  I tried moving them up to be just under "hardware" (siblings to the avr folder) and nothing changed.  I have six projects waiting for me to be able to download code, and for the last six months, I could not compile some of the code (I'm hoping 1.6.2 really does fix these problems), and while I could download the code that did compile to an Arduino Uno for prototyping, I can't get it downloaded into the Trinket Pro in the "final product" implementation.  This is extremely frustrating.

supuflounder:
Using what I thought was the most recent version (1.6.2), I got the following error compiling my sketch:

[code]
/Applications/Arduino.app/Contents/Resources/Java/hardware/tools/coan source -m -E -P -kb -DF_CPU=16000000L -DARDUINO=10602 -DARDUINO_AVR_UNO -DARDUINO_ARCH_AVR /var/folders/qk/48tgmmgj3bq6x7_r2cxw_20w0000gn/T/sketch4171742237898923971.cpp
Using library Adafruit_NeoPixel in folder: /Users/flounder/Documents/Arduino/libraries/Adafruit_NeoPixel (legacy)
Using library colorref4 in folder: /Users/flounder/Documents/Arduino/libraries/colorref4 (legacy)
/Applications/Arduino.app/Contents/Resources/Java/hardware/tools/ctags -u --language-force=c++ -f - --c++-kinds=svpf --fields=KSTtzn /var/folders/qk/48tgmmgj3bq6x7_r2cxw_20w0000gn/T/sketch5036631053244141579.cpp
processing.app.debug.RunnerException
	at processing.app.debug.Compiler.compile(Compiler.java:359)
	at processing.app.debug.Compiler.build(Compiler.java:98)
	at processing.app.Sketch.build(Sketch.java:1169)
	at processing.app.Sketch.build(Sketch.java:1142)
	at processing.app.Editor$BuildHandler.run(Editor.java:1979)
	at java.lang.Thread.run(Thread.java:695)
Caused by: java.lang.NullPointerException
	at processing.app.preproc.CTagsParser.addPrototypes(CTagsParser.java:94)
	at processing.app.preproc.CTagsParser.preprocess(CTagsParser.java:38)
	at processing.app.preproc.CTagsBakedPreprocessor.parseCTagsOutput(CTagsBakedPreprocessor.java:113)
	at processing.app.preproc.CTagsBakedPreprocessor.preprocess(CTagsBakedPreprocessor.java:52)
	at processing.app.preproc.CTagsBakedPreprocessor.preprocess(CTagsBakedPreprocessor.java:42)
	at processing.app.debug.Compiler.compile(Compiler.java:357)
	... 5 more
[/code]
If you need the actual sketches to test this, I can send them.
     joe

ffissore:
@supuflounder yes please, send your sketch and the libraries it uses to f.fissore@arduino.cc

NicoHood:
With a custom core I get:
`Cannot run program "{runtime.tools.avr-gcc.path}/bin/avr-g++": CreateProcess error=2, The system cannot find the file specified`

May someone update this PR?

Without a custom core the bug, that drove me crazy is now gone. I really look forward to see this in the new IDE and working. An example where I need it:
My Infrared decoding library can use PinInterrupts and PinChangeinterrupts. 2nd thing needs a special library. If you include the PCINT library before the Infrared library (which is purely .h inline) it can detect the inclusion and a) include the library as well and dynamically add those new pins that are available to decode the signals. But right now the definition is not passed to the Infrared library because it somehow gets reordered wrong.

This PR fixes it, but it doesnt like my HID-Project core. You might want to try it yourself.

RobAtticus:
Hi,

This issue seems to have gone stale a bit, but I was wondering what the outlook for this is. These changes seem greatly useful to me and wanted to know if I can expect to see them hit the mainline sometime soon.

Jeppedy:
Would love to see the preprocessor fixes merged.  It has taken a good deal of my time before I figured out what was odd about the Arduino IDE.  I imagine others would love being saved from this confusion about IDE behavior...

NicoHood:
Now there are conflicts in the branch...
I aggree to jeppedy. Please rebuild the IDE with the bot then to test it again.

ffissore:
Superseded by #3779

PR #2734, opened 2015-03-06
Update Japanese sentence in preference dialog.

Shinichi-Ohki:


cmaglie:
@Shinichi-Ohki 
we use Transifex for the translations, please register here if you want to contribute to translations:

https://www.transifex.com/projects/p/arduino-ide-15/

Thanks!

PR #2771, opened 2015-03-14
Update EthernetServer.cpp

qbahn21:
Sometimes WIZ5100 reports status ESTABLISHED without any data available. In such a case, whole socket is permanently blocked. Moreover there is no possibilities to implement time out methods on higher level of application because EthernetServer returns no client.


CaptCoonoor1:
Didn't get you?

agdl:
This issue was moved to arduino-libraries/Ethernet#34

agdl:
Sorry closed by mistake

q2dg:
Shouldn't the open issue be the other one instead of this ??

qbahn21:
Better late than never :)
I had to modify local libraries, but it's nothing wrong with that. Meantime I noticed that this lib uses only one socket. It may be good enought in case of small Arduino but unsufficient for bigger or faster systems.

PaulStoffregen:
Ethernet 2.0.0 adds EthernetServer::accept(), which solves this issue.

I wrote some explanation about it on this blog article.  Scroll down to "Server Functions".
https://www.pjrc.com/arduino-ethernet-library-2-0-0/

The AdvancedChatServer example has been updated to show how to use this new function.

Closing this old issue.

PR #2783, opened 2015-03-17
Attempt to fix glitches introduced by java8 on windows. Fixes #2761

### LINKED ISSUE #2761: GUI glitch
My arduino GUI is glitching. It works, but it is really anoying when text is disappearing. I dont know how to explain. So the pictures should give more info: 
In first pic we see almost normal GUI, but these white dots doesnt seem right...

Then the fun begins, I sweep my mouse over text (i mean "file, edit, sketch...") and its gone. Well, almost gone... Few symbols stays sometimes.

Then even i'm getting even more fun: i press "File" in the corner and move my mouse around options, and they are disappearing too:

So please, help me. I can't work like this... Everything else works fine i gues. I  mean code writing, compiling, uploading etc. But when I need to open something from sketchbook, I'm almost guesing what I'm gonna open, cause file names are becoming invisible whan I move my mouse over them (and they are not comming back until i press something else).
It goes the same way with all other options (edit, sketch, etc.).

All started after i updated my arduino to 1.6.1 yesterday

What I have tried: reinstalling arduino, downloading zip file, updating java (idk if that has anything to do with arduino...), restarting computer. 
Any suggestions?

ffissore:
/cc @Nerijuslt4 @jcarlosgarciauah

### DIFF
--- a/build/windows/launcher/config.xml
+++ b/build/windows/launcher/config.xml
@@ -33,6 +33,7 @@
     <maxVersion></maxVersion>
     <jdkPreference>preferJre</jdkPreference>
     <opt>-splash:./lib/splash.png</opt>
+    <opt>-Dsun.java2d.d3d=false</opt>
   </jre>
   <messages>
     <startupErr>An error occurred while starting the application.</startupErr>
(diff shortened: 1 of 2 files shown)


Nerijuslt4:
Had only few free minutes and tried 
http://arduino.cc/download.php?f=/javaide/pull_requests/arduino-PR-2783-BUILD-224-windows.zip 
Those issues seems like to be gone at first look. Ill test it more later.

PR #2793, opened 2015-03-19
'Uncertified' warning should be removed

sgk:
The warning message just confuses users and have no other benefits.

**Who are you all working for?** I would like to work for users and community. **DO NOT INVOLVE INNOCENT USERS WITH YOUR INTERNAL CONFLICT.** Additionally, do you recognize that the trademark is effective for the LLC only in the US? Do you know it is **illegal** in certain countries to claim any products with legal registered trademark as "fake" or something? @ffissore 

Arduino IDE is now world wide popular software product. You have to remove the message or release such version of IDE only in the US.

### DIFF
--- a/hardware/arduino/avr/boards.txt
+++ b/hardware/arduino/avr/boards.txt
@@ -16,7 +16,7 @@ yun.pid.2=0x0041
 yun.vid.3=0x2A03
 yun.pid.3=0x8041
 
-yun.vid.0x2A03.warning=Uncertified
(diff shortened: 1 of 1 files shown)


MauroMombelli:
+1 if the warning message has to be kept, then it should apply for ANY "VID" not officially owned/recognized by arduino LLC

ffissore:
Fixed in 90026f3764c0607cc885e274bcb4cfbad9fba5d2. Available with avr core 1.6.6 and sam core 1.6.4

PR #2806, opened 2015-03-22
Correcting Value Declaration For ACCEL_Y_PIN

SpiderMatt3825U:
The following definition of the accelerometer's Y axis is incorrect based on the preceding comments:

const byte ACCEL_Y_PIN = A11;

The following would be the correct definition based on the preceding comments:

const byte ACCEL_Y_PIN = A7;

### DIFF
--- a/libraries/Esplora/src/Esplora.cpp
+++ b/libraries/Esplora/src/Esplora.cpp
@@ -51,7 +51,7 @@ const byte BUZZER_PIN = 6;
 // Led 13: D13
 
 const byte ACCEL_X_PIN = A5;
-const byte ACCEL_Y_PIN = A11;
+const byte ACCEL_Y_PIN = A7;
 const byte ACCEL_Z_PIN = A6;
 
 const byte LED_PIN     = 13;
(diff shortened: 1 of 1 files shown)


facchinm:
Thank you @SpiderMatt3825U , nice finding. 
However, your fix breaks all existing Esplora projects that use the accelerometer :smile: 

In fact the comment is wrong while the code is right.
Change it and I'll be glad to accept the PR!

SpiderMatt3825U:
This possibility had also crossed my mind (my apologies, I wasn't setup in my current environment at the time to compile this although the forward effect should have been considered on my part). I will get it tweaked and submitted. Thanks @facchinm!

facchinm:
That comment almost all broken, in fact from the schematics (and the code) the right assignment is
```C
// Accelerometer: x-A5, y-A11, z-A6
// External outputs: D3, D11
// Buzzer: D6
// RGB Led: red-D5, green-D10, blue-D9
// Led 13: D13
```
I'll commit the patch squashing yours and fixing the remaining wrong definitions
Thanks!

facchinm:
Fixed with ae8549c
Thank you @SpiderMatt3825U !

PR #2814, opened 2015-03-24
Fix findUntil in Stream library

facchinm:
PR #2696 without timeout-related changes


facchinm:
closes #2696 and #2591

PR #2820, opened 2015-03-26
Update to EEPROM, added bit access.

Chris--A:
This commit adds `EEPROM.writeBit()` and `EEPROM.readBit()` functions.

The way it has been implemented is similar to my [BitBool](http://forum.arduino.cc/index.php?topic=128407.0) library, Basically, the whole EEPROM library is using the same object based methods.

I originally wasn't going to add this, however after speaking to some people, it seems this functionality may be quite useful. In fact with this change, the library will encapsulate every feature found in all custom EEPROM libraries.

#### **`EEPROM.readBit( address, bit )`**

This function allows you to read a single bit value from the EEPROM.
It requires two inputs: an `int` addressing the EEPROM cell, and an `unsigned char` specifiying the bit to read (in a range 0 - 7).

The function returns a `bool` containing the value of the bit read.

#### **`EEPROM.writeBit( address, bit, value )`**

The `writeBit()` method allows you to write a single bit of data to the EEPROM.
Three parameters are needed. The first is an `int` containing the address that is to be written, and the second is the index of the bit to access. The Last parameter is the data to be written (`unsigned char`).

This function does not return any value.
 
---

The system maintains its optimal performance, and through the use of objects, code becomes very neat. The three examples below all produce the same compiled code.

```C++
// 1. Standard method
unsigned char data = eeprom_read_byte( (uint8_t*) 5 );
eeprom_write_byte( (uint8_t*) 5,  data |= (1 << 4) );

// 2. Using writeBit()
EEPROM.writeBit( 5, 4, true );

// 3. Using subscript operators.
EEPROM[5][4] = true;
```

I have also added a `ready()` function.

#### **`EEPROM.ready()`**

EEPROM access takes longer than a single cycle. This function will return false if there is an operation still pending, if so, any future operations will have to wait until its completed. This may be useful for time critical applications, as all EEPROM methods wait until ready.

This function returns a `bool`, with the value `true` if the EEPROM is accessible without delay.

Cheers, Chris Andrews.

### DIFF
--- a/hardware/arduino/avr/libraries/EEPROM/EEPROM.h
+++ b/hardware/arduino/avr/libraries/EEPROM/EEPROM.h
@@ -1,7 +1,7 @@
 /*
   EEPROM.h - EEPROM library
-  Original Copyright (c) 2006 David A. Mellis.  All right reserved.
-  New version by Christopher Andrews 2015.
+  Version 1.0 Original Copyright (c) 2006 David A. Mellis.  All right reserved.
+  Version 2.0 Copyright (c) 2015 Christopher Andrews.
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
@@ -33,6 +33,8 @@
     This class has an overhead of two bytes, similar to storing a pointer to an EEPROM cell.
 ***/
 
+struct EEBit; //Forward declaration for EERef::opreator[]
+
 struct EERef{
 
     EERef( const int index )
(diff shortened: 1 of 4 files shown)


facchinm:
Hi @Chris--A , I believe the best thing to do is opening a discussion on the mailing list (like you did for EEPROM v2) before accepting this PR.

I'm closing it in the meanwhile, will reopen when the discussion on the ML is over

Chris--A:
@facchinm 

Will do. I'm creating a support thread for the forums first. I'll ask the community what they think of the additions.

Also, they might bring to light an important addition, so I will post in the ML after a discussion.

PR #2823, opened 2015-03-27
Due: modify ADC enable/disable sequence

facchinm:
to avoid the bug #2198 simply reconfigure the pin -> no additional overhead if pinMode configuration is performed at the beginning of the sketch, 4 to 25% overhead on all analogRead() due to the additional check


facchinm:
Superseded by #3524

PR #2848, opened 2015-03-30
Bridge: fix transfers which ignored host response

facchinm:
Bridge.put() was broken by #2781 because it used transfer() 2-parameters overloaded version, which imply that rxlen == 0.
But the Linux Bridge responded, so the check (i >= rxlen) was true and the function timed out after retrying 50 times.

Every bridge command (python side) has been checked and now SHOULD strictly follow this rule and never ignore silently the received data

### DIFF
--- a/libraries/Bridge/library.properties
+++ b/libraries/Bridge/library.properties
@@ -1,5 +1,5 @@
 name=Bridge
-version=1.0.0
+version=1.0.1
 author=Arduino
 maintainer=Arduino <info@arduino.cc>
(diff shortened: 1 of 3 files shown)


ffissore:
Can you also increase version in library.properties?

ffissore:
@ArduinoBot build this please

cmaglie:
Please check also the `Bridge::get()` command: the host sends back an "unbounded" response, so it may exceed the reserved buffer.

ffissore:
@ArduinoBot build this please

ffissore:
Let's make another build since we pushed lots of fixes in the last two days

Issue #2847, opened 2015-03-30
Advise users of duplicate libraries


PaulStoffregen:
Sometimes Arduino users will install a duplicate copy of a library, or another library that appears to be a duplicate due to conflicting header file names.  Sometimes this actually works, but breaks months or years later when other code depending on those libraries updates, long after the user as forgotten they installed the library.

When this problem strikes, the user gets incomprehensible compiler errors.

Arduino should detect duplicate or ambiguous library matches and print a friendly warning after compiling, even if no error resulted, to give users clear visibility that it had to guess which library to build into their program.

PaulStoffregen:
@ffissore - I'm working on a patch with the new 1.6.2 code.

The difficult part is the new LibrariesIndexer class.  As it rescans libraries, it uses LibraryList addOrReplace(lib).  That produces a nice list, but a lot of information about possible duplicates is lost.  I have the rest working, and now I'm trying to extend LibraryList to also create a list with duplicates preserved.  Does that sound ok?  Or would you prefer this done some other way?

ffissore:
Sounds ok to me

ffissore:
#2850 merged, closing

PR #2850, opened 2015-03-30
Advise of duplicate libraries after compiling

PaulStoffregen:
When 2 or more libraries match to the same #include in the sketch, this prints a helpful message after compiling.  As discussed on issue #2847 and the mail list...

Internally, this change extends LibrariesIndexer to keep another list of libraries which includes duplicates.  In BaseNoGui, importToLibraryTable is extended to map header names to LibraryList objects, instead of only a single UserLibrary object.  The building of importToLibraryTable is improved to also add the duplicates found by LibrariesIndexer.  Compiler is extended to use the new importToLibraryTable, and take note of any LibraryList is uses with more than 1 library.  After compiling, those are used to print the helpful message.

### DIFF
--- a/arduino-core/src/cc/arduino/contributions/libraries/LibrariesIndexer.java
+++ b/arduino-core/src/cc/arduino/contributions/libraries/LibrariesIndexer.java
@@ -52,6 +52,7 @@ public class LibrariesIndexer {
 
   private LibrariesIndex index;
   private final LibraryList installedLibraries = new LibraryList();
+  private final LibraryList installedLibrariesWithDuplicates = new LibraryList();
   private List<File> librariesFolders;
   private final File indexFile;
   private final File stagingFolder;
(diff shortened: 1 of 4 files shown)


PaulStoffregen:
Here's an example of the message, when a copy of SPI is placed in the user's sketchbook libraries folder:

duplicate_lib_message

PaulStoffregen:
Here's another example, when the copy of SPI in the sketchbook libraries is an old copy from Arduino 1.0.6.  For users who long-ago made copies of libraries that later conflict or need to be updated, this pull request adds a nice message where they would have only seen difficult compile errors from deep within the libraries.

duplicate_lib_message2

PaulStoffregen:
Please consider merging this before the 1.6.3 release.  It's very low risk.  The only practical effect is printing a message, so in a worst case scenario, it prints a confusing message.  No actual build behavior is changed.

I'd like to move forward on redoing the library dependency detection (1.6.2 broke that pull request).  I want to build more pull requests for more features on top of this work, so I'm really hoping you can find a few moments to review this low risk change and merge it.

ffissore:
Thanks @PaulStoffregen

PaulStoffregen:
thanks :)

PaulStoffregen:
Just want to say thanks again for adding this.  I know it was tight timing to squeeze it in between 1.6.2 and 1.6.3 and I really appreciate your effort.

This patch is **really** helping with end user support on forums!  Here's a perfect example:
https://forums.adafruit.com/viewtopic.php?f=47&t=79059&p=405524#p405503

ffissore:
:8ball:

PR #2862, opened 2015-03-31
Add charset selection for output code

NeiroNx:
Some times need compiled code with charset different from UTF-8


matthijskooijman:
It seems your commit changes _every_ line in the file. Presumably your editor automatically converted from DOS to UNIX line endings, or vice versa. You should probably configure your editor to just keep the existing line ending type, if possible.

For this PR, could you restore the line endings so we can actually review the commit? You can use `git commit  --amend` to modify the commit, and a forced push to update the commit and this PR.

NeiroNx:
Eclipse ...

PR #2903, opened 2015-04-04
Char now has the right KEYWORD

mixania:
Before, char was assigned as a structure to KEYWORD3. Since it is a constant, I assigned it to LITERAL2 instead. 

I also renamed the DecrementCompound to its appropriate name.


ffissore:
Thanks!

mixania:
It's a pleasure!

PR #2907, opened 2015-04-04
Show sub-menu settings in Tools menu

PaulStoffregen:
This patch adds settings display in the Tools menu, for each sub-menu.  It's visually easy to see both the selected board and port, and any other custom menu settings, without having to position the mouse over each one, one at a time.


PaulStoffregen:
With this patch:
after

PaulStoffregen:
Before this patch:
before

Chris--A:
This looks like a useful change.

00alis:
I really like it as well!

ffissore:
Thanks!

PR #2938, opened 2015-04-08
Added definitions for USB MIDI interface

gurbrinder:
Added USB midi interface to Aruduino Due


matthijskooijman:
I just noticed half a dozen of pullrequests from you, each containing one commit modifying just one file. I haven't looked in detail, but I don't think this is the right approach. Guidelines for commits and PR's are:
 - One commit per logical change, regardless of how many files are changed.
 - Each commit should be small, but also self-contained (normally you should have working code after each commit).
 - One PR should contain one or more related commits

Perhaps you could restructure your changes, taking into account the above?

PR #2943, opened 2015-04-08
USB MIDI for Arduino

gurbrinder:
Added USB MIDI device capability top Arduino Due


facchinm:
Hi @gurbrinder , 
I think that something horrible happened to your commit ed21dbd , as the full cpp file is replaced by its header.

Could you recover the correct one (with your modifications, of course) and push force over it?

gurbrinder:
Fixed the file.

gurbrinder:
Hi,
I fixed the file.


Gary

On Tue, Apr 14, 2015 at 12:44 AM, Martino Facchin <notifications@github.com>
wrote:

> Hi @gurbrinder <https://github.com/gurbrinder> ,
> I think that something horrible happened to your commit ed21dbd
> <https://github.com/arduino/Arduino/commit/ed21dbd5df50b00182ac8efda3d637bf9d5e03af>
> , as the full cpp file is replaced by its header.
>
> Could you recover the correct one (with your modifications, of course) and
> push force over it?
>
> —
> <https://github.com/arduino/Arduino/pull/2943#issuecomment-92678244>.
>



-- 
Regards,
Gurbrinder Grewal

facchinm:
@ArduinoBot build this please

facchinm:
I finally found some time to test the patch and it works great (setting up my Linux machine to hear something was a PITA instead :smile: )

The example must be moved (for example to `libraries/Audio/USBMIDI/` folder) because it will open as a project if placed in the same folder with `SimpleAudioPlayer`

Then it would be great to have some smart functions wrappers like `USBMIDI.noteOn` to mimic `Mouse` and `Keyboard` classes, which hide the complexity in a very smart way

facchinm:
Completely ported to PluggableUSB [in this repo] (https://github.com/arduino-libraries/MIDIUSB). 
Thanks @gurbrinder for the initial code, it was extremely useful.

odbol:
Hi @gurbrinder ! I've been further testing the USB MIDI implementation you made, and have come across an issue that I can't figure out: the whole Arduino sketch will freeze if you try using it after the USB is connected, but before an app opens the MIDI port and starts listening to the MIDI sent from the Arduino. Was hoping you could help me figure it out or at least point me in the right direction.

I've narrowed it down to this:

1.    When you plug the USB into a computer, the SerialUSB port initializes.
2.    The sketch calls MidiUSB.sendMIDI(), which freezes the whole sketch.
3.    The reason, is that the USBD_Send() function blocks until there is a connection.
4.    Once you open a MIDI app on the PC (such as MIDI Monitor or Garageband), it opens the MIDI port, and the sketch unfreezes ( USBD_Send() stops blocking).
5.    Sending over SerialUSB works: the SerialUSB.write() function does not block if there is nothing connected. See the "if (_usbLineInfo.lineState > 0)" line in CDC.cpp for how they do this.
6.    We need to figure out a way to tell if USBD_Send() will block (i.e. the MIDI port is opened on the other end), and avoid calling it if it will block (just drop the messages if there's nothing listening).

I've attempted to duplicate the "_usbLineInfo.lineState" technique in CDC.cpp, but couldn't get it to work:

https://github.com/odbol/Arduino/tree/midi-fixserial (differences from this version: https://github.com/odbol/Arduino/compare/midi...odbol:midi-fixserial)

Any advice or help would be much appreciated! If you'd like me to write a test sketch to show the problem, let me know.

Thanks so much for your help!

Tyler

odbol:
I've crafted a very simple test sketch to illustrate the problem:

https://github.com/odbol/Arduino-MIDI-Test-Sketch

When you run it on a Due, the LED should blink consistently. If MIDI is enabled in your Arduino Core, the LED stops blinking (until you launch a MIDI app that opens the MIDI port of the Arduino—then it works as  expected).

odbol:
Just opened this as an issue on the new repo where this code is housed, since that's where we'd apply the fix: https://github.com/arduino-libraries/MIDIUSB/issues/4

odbol:
I think I may have fixed this now: https://github.com/arduino-libraries/MIDIUSB/pull/5

gurbrinder:
Tyler,

I am sorry I have not been responsive. I am currently in China for work.
Also I have switched my job to Apple and I don't work on Arduino anymore so
don't have hardware to test with.

Gary

PR #2947, opened 2015-04-09
Allow library manager to install multiple libraries at once.

Lauszus:
I just selected comma as the delimiter, but if you prefer something else I can change it.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -368,28 +368,30 @@ public class Base {
       indexer.setLibrariesFolders(BaseNoGui.getLibrariesPath());
       installer.updateIndex();
 
-      String[] libraryToInstallParts = parser.getLibraryToInstall().split(":");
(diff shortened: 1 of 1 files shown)


Lauszus:
Btw you might want to see the diff while ignoring whitespace: https://github.com/arduino/Arduino/pull/2947/files?w=1

ffissore:
super fast :)

ffissore:
I'm just changing error messages so that they mention, for example, which library is already installed

ffissore:
Ah sorry you did that already in #2948

ffissore:
Thanks!

Lauszus:
You are welcome :)

Btw the reason why I wanted this was for the Balanduino: https://github.com/tkjelectronics/Balanduino, as this will allow me to write an install script that installs all required libraries and the hardware add-on. This should make it much easier for newcomers to the Arduino scene.

PR #2948, opened 2015-04-09
Library debug

Lauszus:
Print library name when installing via library manager and don't throw exception if library is already installed.

### DIFF
--- a/arduino-core/src/cc/arduino/contributions/libraries/LibraryInstaller.java
+++ b/arduino-core/src/cc/arduino/contributions/libraries/LibraryInstaller.java
@@ -106,7 +106,7 @@ public class LibraryInstaller {
 
     // Step 1: Download library
     try {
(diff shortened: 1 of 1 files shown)


Lauszus:
This especially makes sense when multiple libraries are installed at once: https://github.com/arduino/Arduino/pull/2947.

ffissore:
Thanks! I'll change error messages a bit so that they play more nicely with i18n

Lauszus:
Okay thanks. I was wondering about that as well, but just submitted it anyway ;)

ffissore:
Thank you :) This is the commit f0cc644d8f60c65a95f6e5bc0e74c68feb1a2fd7 . It plays nicer because only string `"Library is already installed: \"{0}\""` is extracted and translated. This also allows translations to move `{0}` around, to fit a particular grammar

Issue #292, opened 2012-11-15
tone command does not work correctly on ATmega8 [imported]


cmaglie:
This is [Issue 292](http://code.google.com/p/arduino/issues/detail?id=292) moved from a Google Code project.
Added by 2010-07-15T07:34:09.000Z by [p.reckte...@gmail.com](http://code.google.com/u/100614490999857774768/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium

### Original description

**What steps will reproduce the problem?**
1. load and run a tone example script, f.e. toneMelody.pde

**What is the expected output? What do you see instead?**

You can hear the melody correctly on a Arduino Duemilanove with ATmega328.
On a board with ATmega8 you can only hear one tone (NOTE_B3 I think??). 

**What version of the Arduino software are you using? On what operating**
**system?  Which Arduino board are you using?**

Arduino018 on Windows 7. myAVR USB MKII board with ATmega8 3.6864MHz Quarz.

**Please provide any additional information below.**
Only the following frequencies give a correct signal: 61Hz, 244Hz, 976Hz, 3.9KHz. Any other frequency gives no signal (PWM output signal measured with oscilloscope)

jan-r:
A fix for this issue has been proposed, see pull request #2961.

facchinm:
PR #2961 merged

PR #2961, opened 2015-04-10
Fix for issue #292

jan-r:
This bugfix makes the Tone library work with ATmega8-based boards. See #292.

Rebased the bugfix from the original [Google Code issue 292](https://code.google.com/p/arduino/issues/detail?id=292) to work with Arduino 1.6.x

Description of original fix provided by Pete62:
The later 8 bit AVR's use two registers (TCCRxA, TCCRxB) whereas the ATmega8 only uses a single register (TCCR2) to house the control bits for Timer 2.  Bits were inadvertently being cleared.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Tone.cpp
+++ b/hardware/arduino/avr/cores/arduino/Tone.cpp
@@ -30,6 +30,7 @@ Version Modified By Date     Comments
 0006    D Mellis    09/12/29 Replaced objects with functions
 0007    M Sproul    10/08/29 Changed #ifdefs from cpu to register
 0008    S Kanemoto  12/06/22 Fixed for Leonardo by @maris_HY
+0009    J Reucker   15/04/10 Issue #292 Fixed problems with ATmega8 (thanks to Pete62)
 *************************************************/
 
 #include <avr/interrupt.h>
@@ -296,13 +297,13 @@ void tone(uint8_t _pin, unsigned int frequency, unsigned long duration)
 #if defined(TCCR0B)
       if (_timer == 0)
       {
-        TCCR0B = prescalarbits;
+        TCCR0B = (TCCR0B & 0b11111000) | prescalarbits;
       }
       else
 #endif
 #if defined(TCCR2B)
       {
-        TCCR2B = prescalarbits;
+        TCCR2B = (TCCR2B & 0b11111000) | prescalarbits;
       }
 #else
       {
@@ -456,19 +457,19 @@ void disableTimer(uint8_t _timer)
 
 #if defined(TIMSK3)
     case 3:
-      TIMSK3 = 0;
+      TIMSK3 &= ~(1 << OCIE3A);
       break;
 #endif
 
 #if defined(TIMSK4)
     case 4:
-      TIMSK4 = 0;
+      TIMSK4 &= ~(1 << OCIE4A);
       break;
 #endif
 
 #if defined(TIMSK5)
     case 5:
-      TIMSK5 = 0;
+      TIMSK5 &= ~(1 << OCIE5A);
       break;
 #endif
   }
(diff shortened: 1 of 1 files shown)


hardware/arduino/avr/cores/arduino/Tone.cpp
```diff
@@ -456,19 +457,19 @@ void disableTimer(uint8_t _timer)
 
 #if defined(TIMSK3)
     case 3:
-      TIMSK3 = 0;
+      TIMSK3 &= ~(1 << OCIE3A);
```
facchinm: Did you test it with a Leonardo? 
If so, it would be more coherent to use the notation `bitWrite(TIMSK3, OCIE3A, 0);` as in the previous case

hardware/arduino/avr/cores/arduino/Tone.cpp
```diff
@@ -296,13 +297,13 @@ void tone(uint8_t _pin, unsigned int frequency, unsigned long duration)
 #if defined(TCCR0B)
       if (_timer == 0)
       {
-        TCCR0B = prescalarbits;
+        TCCR0B = (TCCR0B & 0b11111000) | prescalarbits;
```
facchinm: seems ok to me, although I've only read the datasheets and did't test it live

hardware/arduino/avr/cores/arduino/Tone.cpp
jan-r: Actually the problem seems to be TIMSK5... Just tried to compile for the Leonardo:

```
Arduino: 1.6.1 (Linux), Platine: "Arduino Leonardo"

/home/jan/arduino-1.6.1/hardware/arduino/avr/cores/arduino/Tone.cpp: In function 'void disableTimer(uint8_t)':
/home/jan/arduino-1.6.1/hardware/arduino/avr/cores/arduino/Tone.cpp:472:24: error: 'OCIE5A' was not declared in this scope
       TIMSK5 &= ~(1 << OCIE5A);
                        ^
Fehler beim Kompilieren.
```

I'll try to find a fix for that.

jan-r: According to the ATmega32u4 datasheet, the Leonardo doesn't have a timer 5 and therefore does not have TIMSK5. Nevertheless, iom32u4.h defines TIMSK5, but without defining individual bits in it:

```
...
#define TIMSK4 _SFR_MEM8(0x72)
#define TOIE4 2
#define OCIE4B 5
#define OCIE4A 6
#define OCIE4D 7

#define TIMSK5 _SFR_MEM8(0x73)

#ifndef __ASSEMBLER__
#define ADC _SFR_MEM16(0x78)
#endif
#define ADCW _SFR_MEM16(0x78)
...
```

So the main question is why the CPU header defines TIMSK5 at all. But at least it explains the partly complicated #if defined(...) sequences in Tone.cpp, where the presence of all macros is carefully checked before actually doing something.

I've fixed this now by extending these preprocessor checks. The toneMelody example was tested with these changes on the following platforms:

* Compiles & runs on Leonardo
* Compiles & runs on "bare" ATmega8 (\w 16 MHz XTAL)
* Compiles on Uno
* Compiles on Yun
* Compiles on LilyPad
* Compiles on Mini (\w ATmega168)

I'll push the new fix in a minute.

jan-r:
I've fixed this now by extending these preprocessor checks. The toneMelody example was tested with these changes on the following platforms:

Compiles & runs on Leonardo
Compiles & runs on "bare" ATmega8 (\w 16 MHz XTAL)
Compiles on Uno
Compiles on Yun
Compiles on LilyPad
Compiles on Mini (\w ATmega168)

facchinm:
Thank you @jan-r !

PR #2972, opened 2015-04-13
Update Tone.cpp to fix #2923

### LINKED ISSUE #2923: Tone.cpp stops compile when TCCR0A/TCCR0B defined and WGM01 not defined
In case using an own pins_arduino.h for the ATtiny861 the compile stops in Tone.cpp.

This can be avoided adding an additional "#if defined" statement which catches the case of undefined WGM01:

Tone.cpp line 154: 
#if defined(TCCR0A) && defined(TCCR0B) 
-> 
#if defined(TCCR0A) && defined(TCCR0B) && defined(WGM01)

jipp:
Tone.cpp stops compile when TCCR0A/TCCR0B defined and WGM01 not defined #2923


facchinm:
Manually merged

Issue #2957, opened 2015-04-10
Improve all warnings


ffissore:
See https://github.com/arduino/Arduino/commit/61592d78fa71c023e4f1577f7e2b6bada2661a13#commitcomment-10668365

matthijskooijman:
In that comment, @cmaglie wrote:

> Sorry I was a bit in a hurry, here a less terse comment ;-)
> 
> IMHO it should be implemented in a way similar to what we do for tools verbosity:
> 
> ```
> tools.avrdude.bootloader.params.verbose=-v
> tools.avrdude.bootloader.params.quiet=-q -q
> tools.avrdude.bootloader.pattern="{cmd.path}" "-C{config.path}" {bootloader.verbose} -p{build.mcu} -c{protocol} {program.extra_params} "-Uflash:w:{runtime.platform.path}/bootloaders/{bootloader.file}:i" -Ulock:w:{bootloader.lock_bits}:m
> ```
> 
> in this case `tools.avrdude.bootloader.params` is loaded with the contents of `tools.avrdude.bootloader.params.quite` if the verbosity is off or with the contents of `tools.avrdude.bootloader.params.verbose` if the verbosity is on.
> 
> For the "warning level" setting we can do something like:
> ```
> compiler.warning_flags.none=-w
> compiler.warning_flags.normal=
> compiler.warning_flags.all=-Wall
> compiler.warning_flags.extra=-Wall -Wextra
> 
> recipe.c.o.pattern="{compiler.path}{compiler.c.cmd}" {compiler.c.flags} {compiler.warning_flags} ............
> ```

It seems this was implemented in
https://github.com/arduino/Arduino/commit/b42c6667e1c6ca748b3a24bdd87c5716332a9c39
already. I haven't looked closely at that commit yet, but I was writing
a response to @cmaglie's proposal already when I found that commit.

I totally agree with @cmaglie's comment - hardcoding these flags is not
the way to go. It's a bit of a pity that there is no PR for this change
to discuss things before merging into master, now the master branch is
seeing multiple different versions of this feature.

The approach proposed seems good to me. I even like it better than the
bootloader params one, since the `{compiler.warning_flags}` replacement
refers to e.g. `compiler.warning_flags.none` more clearly than
`{bootloader.verbose}` to e.g.
`tools.avrdude.bootloader.params.verbose`.

Ideally, the warning levels aren't fixed, but the list of possible
values are taken from platform.txt, and some platforms can define more,
[...]

PR #2973, opened 2015-04-13
Reviewed warnings key names and dropdown in Preferences

ffissore:
/cc @matthijskooijman

Still a single dropdown. I'll stick with one dropdown only until the next next release so we can gather some feedback from the community

See #2957


ffissore:
@ArduinoBot build this please

matthijskooijman:
This change looks good to me. As discussed, we can later expand this by adding a `compiler.sketch.warning_flags` setting, which can default to "default" (or perhaps a newly added "selected"), without affecting the current setting (which can still default to "none" as now).

Issue #2868, opened 2015-04-01
CardInfo example doesn't work with nightly


Pitel:
Works with 1.6.1, can't test with 1.6.2 because I'm on Linux (64bit).

When compiled and uploaded with 1.6.1, I get nice info about card and files. On nightly I just get this:
```
Initializing SD card...initialization failed. Things to check:
* is a card is inserted?
* Is your wiring correct?
* did you change the chipSelect pin to match your shield or module?
```

I cleaned `.arduino15` before installing nightly. Nightly's MD5 is `b20bce289c0dc848eee665f81f60b50a`.

Pitel:
I also fixed the typos in #2869 :)

facchinm:
Hi @Pitel ,
does it work if you add 
```arduino
digitalWrite(10, HIGH)
```
just after 
``` arduino 
pinMode(10, OUTPUT);
``` 
?

It could be related to a change (#2659) that solves a problem for someone but could create a lot of trouble to existing sketches.
I'll wait for your feedback before reverting it :smirk:

Pitel:
Yes, it works after adding `pinMode(10, OUTPUT)`.

vicatcu:
I've made an update regarding issue #2659 that may satisfy both concerns. See comments in the.

Pitel:
Well, I'm quite new to Arduino, so I don't know much about this SPI thing. But my general opinion is against legacy stuff. If there is new better/faster/more correct way of doing somehting, and it's not backward compatible, then use the new way. But! Make the change as visible as possible, so that one can easily find how to adopt this new change in their old project.

facchinm:
Fixed in #2974

PR #2974, opened 2015-04-13
SD: remove any reference to pinMode(10, OUTPUT) in the examples

facchinm:
correct SS pin setup is already handled by SPI subsystem.
this should prevent future issues like #2868

current implementation assures that:
* pin10 is OUTPUT HIGH if SPI.begin() is called and the pin was unconfigured
* pin10 state is not modified if pinMode(10, OUTPUT) is called before SPI.begin()
* pin10 is INPUT HI-Z if nor pinMode(10, OUTPUT) nor SPI.begin() are called

### DIFF
--- a/libraries/SD/examples/CardInfo/CardInfo.ino
+++ b/libraries/SD/examples/CardInfo/CardInfo.ino
@@ -44,12 +44,6 @@ void setup()
 
 
   Serial.print("\nInitializing SD card...");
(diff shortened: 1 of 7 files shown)


q2dg:
A question: in case of using Ethernet shield (where SS pin is 4), a line like pinMode(4,OUTPUT); should be still necessary??

PR #2980, opened 2015-04-15
Delete temporary files on exit

### LINKED ISSUE #2971: Compiler "temp" files - Question / configuration or a  bug?
With about 30 h / cpp files being compiled into windows ...Temp directory I have noticed they DO NOT get erased / deleted and the directory keeps growing after each compile. 
Is this normal ? 

I am still happy  running 1.6.0 Nov issue.

ffissore:
The IDE wanted but failed to delete temporary files on IDE exit. Feature restored.
Temp files will be preserved until the IDE is closed
Added a command line option `--preserve-temp-files` for preserving temp files when using the command line
Fixes #2971
/cc @VaclavSal @mschlenker @matthijskooijman


matthijskooijman:
I added one comment inline, other than that the code looks ok to me.

PR #2990, opened 2015-04-17
Half-steps in stepper library for 4-wire motors

testato:
from Issue 139 on code.google.com
i only create the pull request, the code is from John at http://www.midity.com/


facchinm:
duplicate of #2991

testato:
it is not a duplicate, this is the header file, the other is the .cpp

matthijskooijman:
You should make a clone using the git software, make the necessary changes and then commit and push them. Using the webinterface for pullrequest is only useful for trivial typo-fixes, anything more complicated needs a proper git client. If none of this means anything to you, you should probably find a tutorial for git online and work through that.

testato:
ok, i will commit a new complete pull request by my desktop github
thanks

PR #3020, opened 2015-04-22
Absolute coordinates for HID mouse

awatterott:
The define `MOUSE_ABS_ENABLED` enables absolute instead of relative coordinates for the HID mouse device. This can be used for a touch panel driver.


ffissore:
@ArduinoBot build this please

awatterott:
I have checked the test build and it is working.

hardware/arduino/avr/cores/arduino/USBAPI.h
```diff
@@ -110,11 +110,15 @@ class Mouse_
 	Mouse_(void);
 	void begin(void);
 	void end(void);
+#ifdef MOUSE_ABS_ENABLED
+	void moveAbs(unsigned int x, unsigned int y, signed char wheel, unsigned char buttons); // x and y have the range of 0 to 4095
```
damellis: For consistency with the Arduino API style guide (http://www.arduino.cc/en/Reference/APIStyleGuide), this should probably be "moveAbsolute()" not "moveAbs()".

PaulStoffregen:
When I tried this on Teensy, a few unexpected issues came up.

The main one was absolute positioning is incompatible with nearly all games.  Desktop usage works fine, but apparently most game software requires relative.

The other problem, still unresolved, was incompatibility with most Linux systems if both relative and absolute are in use.  Windows and Mac seem to handle that fine, but there are problems with X11.

awatterott:
We are using the mouse device with absolute positioning on our [HDMI displays](https://github.com/watterott). It is a board with a TFP401A HDMI to RGB converter and ATmega32u4 as touch and backlight controller. The only problem we found is that some systems (e.g. OpenELEC) want the position data in screen coordinates and not in the range of 0...4095. On our display adapters we know the display resolution so it is no problem to send screen coordinates.
Maybe this is also the issue you have? 
On a standard Raspbian system with LXDE desktop the mouse is working with the default coordinates 0...4095.

NicoHood:
I can report the same issues. I also noticed that the absolute mouse doesnt work under windows server 2012.

And for some weird reasons it only worked if another consumer device is also in the same multireport. But I never found a reason. But I found a report combination that seemed to work fine on most systems.

@awatterott try this: https://github.com/NicoHood/HID

awatterott:
@NicoHood Do you refer to my code or your code? Have you tried to test screen coordinates if it did not work?

NicoHood:
I tested with my own core. USB can cause so many problems, it would be very critical to add this PR like this. They should think of different USB options the user can choose and keep the default setting to not kill the windows driver recognition.

awatterott:
I found no real problems with the absolute mouse and coordinates in the range of 0 to 4095 (see my descriptor). This is also what Microchips [AR1100](www.microchip.com/AR1100) touch controllers use. I have tested it with the following Windows versions: XP, 7 and 8.1

NicoHood:
@awatterott Andreas, the problem is, that if you change the USB HID Descriptor some things could be broken on several OS. I've tested a lot of USB stuff and Paul Stoffregen will aggree that very fast with a few changes things can go really wrong.

I could give you a lot of examples. But here is a simple one: I've added Media keys to a report. Everything seems to work pretty good. And someone else reported that it doesnt work at all under MAC. Even the keyboard was dead. Simply because a single line in the descriptor was wrong so the OS didnt accept it. Not sure if it was technically wrong because it worked under the other OS.

Well now you dont use Media keys. But still there is this driver problem. If you now switch back to the old core, Windows might throw errors. And also it might not work under all OS, maybe the 2nd Mouse breaks the whole function.

Its not that I dont support your idea. I am not even one from the Arduino team. I've created a lot of custom stuff with USB and the Arduino IDE if you've clicked the link above. Your idea is great and really useful! But simply adding this stuff to the Core HID files isnt a good idea. The user should be aware what he is doing, if he wants additional absolute mouse positioning. There are to many untested scenarios where bugs can occur with this addition. The Arduino team wont accept any USB PR I bet. They should redevelop the whole core and take some time to get it done better.

What you could do for your shield users are two things:
Provide a new Hardware folder for the sketchbook with your new core
Link to the HID Project that I've made which also has this function. Also feel free to test it.

hardware/arduino/avr/cores/arduino/USBAPI.h
NicoHood: Or MoveTo() as Paul has named it.

awatterott:
I see no problems, why the changes should not be included in the official core. The code is tested and the absolute mouse will only be activated, if the user does it via the define statement.

NicoHood:
Your HID report contains Buttons and a scroll wheel. You can do 3 things:
* Use them with the API
* Remove them and use the normal mouse along with the abs mouse. The normal mouse does the buttons and optional the normal variable positioning. The abs mouse is then only used for X+Y.
* Remove them from the report only so you have no click functions.

Also I see no reason why to use abs mouse without buttons (right/left click)?

awatterott:
We use the mouse for a touchpanel and there is only one button... I have changed it, so that the absolute mouse API is compatible with the normal mouse API.

hardware/arduino/avr/cores/arduino/HID.cpp
```diff
@@ -211,17 +260,49 @@ void Mouse_::end(void)
...
+		y = 4095;
+	_y = y;
+
+	m[0] = _buttons;
+	m[1] = LSB(x);
+	m[2] = MSB(x);
+	m[3] = LSB(y);
+	m[4] = MSB(y);
+	m[5] = wheel;
+	HID_SendReport(1,m,6);
```
NicoHood: use sizeof(m) here

facchinm:
@awatterott , this PR can be an extremely useful test case for PR #3304 .
I'd be glad if you could test the modular HID framework within your environment, trying both a modified version of  `Mouse` library and the `CompleteHID` library

facchinm:
Closed with  #1803

awatterott:
@facchinm I have tested the new USB core with our absolute mouse changes and everything is working. The patching without changing core files is really good.

NicoHood:
Absolute Mouse is also available here:
https://github.com/NicoHood/HID/tree/dev_2_4

PR #3023, opened 2015-04-23
making UART/USART handler weak to allow custom external handlers

ricardocrudo:
I've opened a thread about this here: https://groups.google.com/a/arduino.cc/forum/#!topic/developers/FIjuI4z06tY


NicoHood:
I'd also suggest to do this for AVR if possible

ricardocrudo:
The AVR one uses the ISR macro, probably define it again in the user code will work.

ricardocrudo:
polite bump

facchinm:
Moved to arduino/ArduinoCore-sam#1

PR #3037, opened 2015-04-26
Corrected a bug in the SD card library. The SDClass class makes a refere...

squelched:
...nce to "SD.card" instead of just "card". SD is a global instance of SDClass. This prevents any other instance of SDClass from functioning correctly. The fix also allows SDClass to be used with an SD card which is removed and replaced, whereas previously, using the global instance SD did not allow this due to the limitation of begin() which cannot be called more than once.


facchinm:
Merged in 1c25d2432 with a slightly reworked commit description
Thank you very much!

PR #3039, opened 2015-04-27
Allow RX or TX only; pin = -1 for unused

sbrynen:
By specifying an TX or RX pin of -1, you can use the library for receive only or transmit only if you're trying to save pins.


facchinm:
@ArduinoBot build this please

NicoHood:
I'd suggest to pass the pins via template. This would save memory and time. Since only a single SoftSerial instance can exist (due to the ISRs) this would totally make sense.

An example can be found here (ignore the variadic template addition):
https://github.com/NicoHood/MSGEQ7

sbrynen:
The problem is that this library is widely used and needs to be backwards compatible; changing the input to use templates as opposed to actual pin numbers would break countless existing implementations.  However, defining a SERIAL_NONE to -1 probably makes sense, and I'll do that tonight.

NicoHood:
Its a single line in the code that people would have to change. But you are right, there is a lot of crap in the code due to compatiblity only reasons.

But with the new library manager we could add a version label and make a new version. People with an older sketch can use the old library. And as far as I remember the eeprom library changed too. And a template is a very trivial thing that the user has to change.

damellis:
In general, we've tried to avoid introducing new syntax in Arduino, like the <> usually used for templates. Although templates definitely seem like the right solution for some of these issues (passing compile-time constants into libraries), so it would be great if there was a way to use them with the current (constructor-like) syntax.

NicoHood:
Templates are great. And people who never look behind the source wont care anyways if its <> or (). FastLED also uses templates for example. And I found them VERY useful in my recent libraries and will continue to use them. Its nothing overcomplicated.

sbrynen:
Changed the -1 to a constant  _SS_UNUSED to signal an unused pin, which is a little more obvious; so it becomes
```SoftwareSerial  rcvSerial(RXPIN, _SS_UNUSED);```

damellis:
Having a constant sounds good to me. In general, though, we've tried to
avoid abbreviations and we typically don't use an _prefix. So this should
probably be SOFTWARESERIAL_UNUSED.

sbrynen:
Rebasing will re add pull request

matthijskooijman:
FWIW, you could have (can still) force-pushed to your own branch to update the PR, without having to create a new one.

sbrynen:
Not a clue how to do that.  Problem is my pull request now shows a whole 
bunch of other people's commits.  Don't know how that happens or how to 
fix it.

PR #3053, opened 2015-04-29
Option for SOFTWARESERIAL_UNUSED to allow TX or RX only configs

sbrynen:
This patch allows you to pass a pin# of "SOFTWARESERIAL_UNUSED" for a transmit or receive pin, to avoid having to use two pins in situations where you only want receive or transmit.


matthijskooijman:
This PR contains two commits, the second of which removes a merge conflict introduced in the first. They should be squashed together instead of remaining as two separate commits (you can use `git rebase -i origin/master` for that, assuming that `origin` is the upstream repository).

Looking at the first commit, it contains a lot of whitespace changes and introduces a README, both of which are ok changes, but they should really live in their own commits.

As for the actual code changes, they look ok at first glance (though above things make things a bit harder  to review). I did notice an if like this a few times:

    if ((_receivePin != SOFTWARESERIAL_UNUSED) && (active_object == this))

However, AFAICS  `active_object` can never be set to this for objects without a valid receive pin? Not sure if it's worht checking the receive pin here (though being on the safe side might be ok, not sure).

sbrynen:
You're right about the active_object.  When I 1st hacked this in I wasn't sure if the object existed with portions of it for TX, so I added the extra check.  Anyway, I suspect an integer test is somewhat faster than an object test, so I don't really see too much harm

I'll try the rebase, but they never seem to work properly for me, there's something fundamental about git that just doesn't sink into my mind.

PR #3073, opened 2015-05-02
Remove trailing whitespaces

utix:
Pure cosmetics.
Just a commit to remove all trailing spaces from cpp and c files. (*.c, *.cpp and *.h)


ffissore:
I would love to merge this as I love cosmetics but these changes are likely to break many other open PRs, so we just can't. It also changes way too many files. Thank you all the same

PR #3085, opened 2015-05-05
Add extra build symbols and recipe

PaulStoffregen:
These are extra features needed to build the Teensy core library.  They might be useful for others too?  At the very least, this stuff should not cause any issues with boards not using these features.

### DIFF
--- a/arduino-core/src/processing/app/debug/Compiler.java
+++ b/arduino-core/src/processing/app/debug/Compiler.java
@@ -34,6 +34,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.Date;
+import java.util.GregorianCalendar;
 
 import cc.arduino.packages.BoardPort;
 import cc.arduino.packages.Uploader;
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I added two minor nitpicks, but overall these look like simple and acceptable changes to me.

PaulStoffregen:
Sure, the names could be changed if anyone feels really strongly.  I'll just have to go back and re-edit my patches in 1.6.1 and 1.6.3 with whatever the new names end up being.

ffissore:
I'm going to merge this, but a better long term solution to this kind of toolchain needs is to provide hooks between the *official* phases. We already have `hooks.prebuild` and `hooks.postbuild`, we need others such as `hooks.sketch.prebuild`, `hooks.linker.postlink` and so on

ffissore:
@PaulStoffregen Please take a look at #3236 I've cherry-picked your first commit and added one that spread hooks all over the build process. Your `build.elfpatch` recipe could then become `recipe.hooks.linking.postlink.0.pattern` and in case you needed one more, you'd just add a `recipe.hooks.linking.postlink.1.pattern`

ffissore:
Last comment came a bit out of the blue: me and @cmaglie had a quick chat and agreed in implementing a generic solution rather that adding a specific build step with a specific semantic that may not fit others needs

ffissore:
Closing as superseded by #3236. Please convert your additional recipes to hooks. You can find some docs here https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#pre-and-post-build-hooks-since-ide-165

Issue #2969, opened 2015-04-12
i2c example using reserved i2c address


bperrybap:
The example on this page:
http://arduino.cc/en/Tutorial/MasterWriter

Is using i2c address 0x4 which appears to be in the reserved address space.
0 to 7 are reserved.

Not sure what to pick for an address. It probabaly doesn't matter, but it would be best to update the example to avoid using an address from the reserved address range.

It's been a while since I've tested, but I believe that these reserved addresses don't work on the pic32 processors.

agdl:
@bperrybap you are right, but the example is intended to be used with two Arduino so the address soesn't matter. Thank you!

bperrybap:
Yes it does!
Just because it doesn't matter on an AVR doesn't mean it doesn't matter on other chips.
The low addresses will not work on Pic32 chips. I have no idea what happens on some of the ARM based chips.
Why continue to promote an example that known to be incorrect and violates specifications?
Fix the code. It is a trivial fix.

matthijskooijman:
+1 for fixing this.

Nantonos:
Hello Arturo,

Friday, April 17, 2015, 11:07:04 AM, you wrote:

> @bperrybapyou are right, but the example is intended to be used
> with two Arduino so the address soesn't matter. Thank you!

That is a totally bogus answer.

And when someone builds on that example to talk between an Arduino and
something else?

Fix the example, please.

-- 
Best regards,
 Nantonos                            mailto:nantonos@epona.net

agdl:
I made a PR because the on-line page takes the code from git. I fix the problem in all the examples. Let me know if it is ok.

https://github.com/arduino/Arduino/pull/3100

agdl:
I adde this statement here:
www.arduino.cc/en/Reference/Wire

"However the addresses from 0 to 7 are not used because are reserved so the first address that can be used is 8. "

Is it ok in this way?

PR #3100, opened 2015-05-07
Fix for #2969

agdl:


matthijskooijman:
Could you please give this commit a proper commit message, that describes what it does? Just referencing to an issue isn't very friendly wrt searching / scanning of the commit log, and makes the commit message useless if the issue list ever gets lost. If you want to refer to an issue, use the "Fixes #2969" syntax somewhere in the commit message body (i.e. not the first line), which will even cause github to autoclose the issue.

bperrybap:
The new example code in the repo with address 8 looks ok, it might also be useful to update the actual web page here:
http://www.arduino.cc/en/Tutorial/MasterWriter
with a note indicating that i2c addresses 0-7 should be avoided as those are reserved.

ffissore:
Indeed @agdl . This is not the first time. Hopefully it's the last

bperrybap:
Huh? I'm now totally lost.
Is the Tutorial web page going to be updated to include a text note about not using reserved addresses 0-7?
Fixed examples is great but I think users should also be warned about using the reserved addresses in the descriptive text.

agdl:
@bperrybap the on-line page take the code from git so it is automatically updated when the PR is merged. For the descriptive part i will add in all the on-line example the statement:

It is suggested to avoid addresses from 0 to 7 ok?

bperrybap:
Cool. Very nice. Thank You.

matthijskooijman:
Why is this closed?

ffissore:
So @agdl can make another one, with a proper commit message, and a proper PR title and a proper PR description

PR #3109, opened 2015-05-08
Fix for #2969 (I2C reserved addresses)

agdl:
This fixes the Wire examples that uses I2C reserved address (from 0 to 7) substituting them with 8 that is the first one available and that can be used.

I also modified the wire reference
http://www.arduino.cc/en/reference/wire
according to this fact.

### DIFF
--- a/hardware/arduino/avr/libraries/Wire/examples/master_reader/master_reader.ino
+++ b/hardware/arduino/avr/libraries/Wire/examples/master_reader/master_reader.ino
@@ -20,7 +20,7 @@ void setup()
 
 void loop()
 {
-  Wire.requestFrom(2, 6);    // request 6 bytes from slave device #2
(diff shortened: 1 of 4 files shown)


matthijskooijman:
This seems like a duplicate of #3100 that was closed for no apparent reason. Why a second PR?

Also, my comments from #3100 have not been addressed, or responded to:

> Could you please give this commit a proper commit message, that describes what it does? Just referencing to an issue isn't very friendly wrt searching / scanning of the commit log, and makes the commit message useless if the issue list ever gets lost. If you want to refer to an issue, use the "Fixes #2969" syntax somewhere in the commit message body (i.e. not the first line), which will even cause github to autoclose the issue.

agdl:
@matthijskooijman as @ffissore suggested i made a second PR that includes an explanation if what the fix does and this is what I did. For the commit message I will modify it.

PR #3129, opened 2015-05-13
Bug (solved): D24 bit value is wrong in digital pin masks.

rafacouto:
In Leonardo hardware pin assignment, the correspondence between bit mask and pin PD5 is missing, while PD4 is repeated as D24, so __D24 should be PD5__. 

Atmega32U4 PD5 pin is _XCK1/~CTS_ and it is connected to TX led on Leonardo board. After this patch, D24 is assigned to PD5 and these lines can blink TX led:

```
void setup() {
  pinMode(24, OUTPUT);
}

void loop() {
 digitalWrite(24, HIGH);   // turn the LED on (HIGH is the voltage level)
 delay(500);              // wait for a second
 digitalWrite(24, LOW);    // turn the LED off by making the voltage LOW
 delay(500); 
}
```

This bug was found by @xdesig (member of the Escornabot project :) This is his hardware unit test: https://www.youtube.com/watch?v=nClBTrmG7Sk

### DIFF
--- a/hardware/arduino/avr/variants/leonardo/pins_arduino.h
+++ b/hardware/arduino/avr/variants/leonardo/pins_arduino.h
@@ -173,8 +173,10 @@ extern const uint8_t PROGMEM analog_pin_to_channel_PGM[];
 // MOSI		D16		PB2					MOSI,PCINT2
 // SS		D17		PB0					RXLED,SS/PCINT0
 //
-// TXLED			PD5
-// RXLED		    PB0
+// Connected LEDs on board for TX and RX
+// TXLED	D24		PD5					XCK1
+// RXLED	D17		PB0
+//
 // HWB				PE2					HWB
 
 // these arrays map port names (e.g. port B) to the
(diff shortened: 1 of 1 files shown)


NicoHood:
That is correct, D4 is A6 so pin 24 is useless and can be assigned to the other LED which is not used yet.

I'd also vote for a
`#define LED_BUILTIN_TX 24`
`#define LED_BUILTIN_RX 17`

NicoHood:
Its such a simple fix, why dont you apply this?

matthijskooijman:
I don't think this fix is correct. Pin 24 was previously used for (virtual) pin "A6", which is the analog input mode on pin D4. This needs a duplicate pin number to be assigned to pin PD4 (as well as others), because `analogRead()` is expected to work with both pin numbers and analog channel numbers (e.g. 0 for A0). If you would just say `A6 = 4` and then do `analogRead(A6)`, this would cause ambiguity: Did you mean pin 4 (A6) or channel 4 (A4)?

So I believe that D24 must remain unchanged. Adding a pin for the TX led makes sense, though I'm not sure if it would be ok to add it at the end - some code might be assuming that the last `NUM_ANALOG_INPUTS` are analog pins (though perhaps not). See also #2071, where some similar problems were encountered.

NicoHood:
Oh you are right. But why not use another pin for the led then?

#define LED_BUILTIN_TX 30
#define LED_BUILTIN_RX 17

cmaglie:
As stated by @matthijskooijman the current pin assignment is correct.
Closing as invalid.

PR #3141, opened 2015-05-14
Fix for FileUtils.recursiveDelete() when dealing with symbolic links

cmaglie:
More detailed info in the commit message:

```
The documentation for File.delete() says that the method return true
if the the file is successfully deleted, otherwise false is returned.
An exception is thrown only when the file is not accessible (for
permission problem).

Removing the extra check solves another problem, for example in a
folder with the following situation:

    linkToFileA -> FileA
    FileA

if we remove FileA, we remain with a broken link that can't be removed
using FileUtils.deleteIfExists() because calling File.exists() on a
broken link returns *false*. This commit solve this problem.
```

note that the bug not always happens, it depends on the order the files are deleted.

### DIFF
--- a/arduino-core/src/processing/app/helpers/FileUtils.java
+++ b/arduino-core/src/processing/app/helpers/FileUtils.java
@@ -274,10 +274,6 @@ public class FileUtils {
       return true;
     }
 
-    if (!file.exists()) {
-      return false;
-    }
-
     return file.delete();
   }
 
--- a/arduino-core/src/processing/app/helpers/FileUtils.java
+++ b/arduino-core/src/processing/app/helpers/FileUtils.java
@@ -85,7 +85,7 @@ public class FileUtils {
         recursiveDelete(current);
       }
     }
-    deleteIfExists(file);
+    file.delete();
   }
 
   public static File createTempFolder() throws IOException {


arduino-core/src/processing/app/helpers/FileUtils.java
```diff
@@ -85,7 +85,7 @@ public static void recursiveDelete(File file) {
         recursiveDelete(current);
       }
     }
-    deleteIfExists(file);
+    file.delete();
```
ffissore: Since you deleted the `.exists()` check in `deleteIfExists`, could you still call that function instead of `file.delete()`? It uses a defensive approach checking if File is `null`

cmaglie: It's explained in the second commit: `file is already checked for being not null, no need to check again`
the check is here: https://github.com/arduino/Arduino/pull/3141/files#diff-fb3046d74e0c4a32b94edd584638ecd3L76

BTW if you prefer to keep the defensive approach and check anyway, just drop the second commit and keep only the first one!

ffissore: Missed that. Thanks

PR #3172, opened 2015-05-16
Optimizations: remove multiple calls to the status() function.

mathertel:
At these places the status function is called multiple times without need.
Because there is a lot of Wire function inside I suggest reducing that and speeding up the functions.


facchinm:
Merged as a9cdd44d27a

PR #3173, opened 2015-05-17
update Firmata to v2.4.3

soundanalogous:
See [revisions.txt](https://github.com/soundanalogous/Arduino/blob/update-Firmata/libraries/Firmata/extras/revisions.txt) for a list of changes since Firmata v2.3.6. Now that Arduino 1.6.x is the master branch I'll try to submit Firmata updates more regularly. Please coordinate with me if any changes are to be made. I'd rather flow changes from the Firmata repo to the Arduino repo if possible (aside from automated changes that are made across all Arduino core libraries - I keep a watch on those and update in the Firmata repo as necessary).


cmaglie:
Oh ok, I've just read this one :-) so I repeat my proposal here too:

It seems that the official repository of Firmata is ready for the library manager, so we can just replace https://github.com/arduino-libraries/Firmata with a fork of https://github.com/firmata/arduino so you don't need to bother submitting pullrequest.

soundanalogous:
That sounds great! [https://github.com/firmata/arduino](https://github.com/firmata/arduino) is the official repository, the master branch is stable and all releases are tagged. The new library and board manager are great BTW!

soundanalogous:
One thing to note is because I have unit tests in the Firmata repo and other features not typically found in other core libraries, I typically run a [build script](https://github.com/firmata/arduino/blob/master/release.sh) that generates the structure of Firmata in the Arduino repo. However in order to support the library manager, I could update the file structure in the Firmata repo as follows:

- move Boards.h, Firmata.cpp and Firmata.h to /src/.
- either move the `/test/` directory to `/extras/test/` or keep it at the root. I'm not sure what the policy is on location of unit tests. I'd prefer to keep them at the root level if possible.

Or, I could create a separate repo `firmata-arduino-release` that contains the output of the build script and the Arduino library manager could reference the release repo.

cmaglie:
* Note that the use of `src` is not mandatory, doing so will break compatibility with IDE 1.0.x, so this is up to you. If you decide to use the `src` folder you must move **all** the source code inside it, also the `utility` folder must go there. Moreover the `src` folder is recursively compiled so you are free to call any sub-folder as you prefer (`utility` could be renamed to anything more meaningful and other folders may be added as well). One difference is that only `src` will be part of the include path so if you include something inside the `utility` folder you must write `#include "utility/something.h"` instead of `#include "something.h"`

* `test` may stay in the root folder

* Just keep your readme, we added them for github user's :-)

So basically you don't need to do anything, I'm going to add the library to the library manager.

cmaglie:
Ok, it turns out that this needs some more clerical work to handle the already released versions from the old repository. I'll check this out tomorrow.

cmaglie:
Done, now all the latest versions are available through library manager.

soundanalogous:
Are tags cached locally? There are 2 tags (v2.6.2 and v2.6.1) that actually were for an experimental future version of Firmata. Since these tags are higher numbers, 2.6.2 is now the default in the library manager, but it should not be used. I have deleted these 2 tags from the repo (just now), but they still show up in the library manager. I'll check again a little later to see if the case is it takes time to propagate.

ffissore:
lib job never deletes. It only adds. Having two more versions is no big
deal IMHO: users will likely stick with the latest
Il giorno 21/mag/2015 17.50, "Jeff Hoefs" <notifications@github.com> ha
scritto:

soundanalogous:
The problem is the versions listed as the latest (v2.6.2 and v2.6.1) are essentially broken. Stable Firmata is currently v2.4.3. The v.2.6.x series is experimental and was actually moved to a new repository ConfigurableFirmata. I forgot to remove the old tags. I guess the only way I can fix this then is to jump the Firmata version from 2.4.3 to 2.7 just for the sake preventing users from updating to a broken version of Firmata.

ffissore:
I see. I'll regenerate it then

ffissore:
Ok, done

PR #3180, opened 2015-05-18
Updated keywords.txt

### LINKED ISSUE #3175: wrong "word" color...
I've just installed IDE 1.6.4 and recompiled some old sketches...
I've noticed that "word" variable identifier is showed in GREEN... Should it not be in BLUE ?
I think this is something like "char" issue... :)

Chris--A:
This commit fixes some highlighting issues due to multiple keyword
entries.

The missing C++ keywords have been added.

A set of standard fixed sized integers has been added also: `uint8_t`,
`int32_t`,...

This also resolves #3175


facchinm:
@00alis can you check this?

Chris--A:
Here is a list of C++ keywords, and the intX_t variations you can paste into the IDE.

```C++
alignas
alignof
asm
auto
bool
break
case
catch
char
char16_t
char32_t
class
const
constexpr
const_cast
continue
decltype
default
delete
do
double
dynamic_cast
else
enum
explicit
export
extern
false
float
for
friend
goto
if
inline
int
long
mutable
namespace
new
noexcept
nullptr
operator
private
protected
public
register
reinterpret_cast
return
short
signed
sizeof
static
static_assert
static_cast
struct
switch
template
thread_local
throw
true
try
typedef
typeid
typename
union
using
virtual
void
volatile
while
unsigned
wchar_t
byte
char
long
float
double
word

DEC
BIN
HEX
OCT

int8_t 
int16_t 
int32_t 
int64_t 
uint8_t 
uint16_t 
uint32_t 
uint64_t
```

PR #3185, opened 2015-05-19
Cut off board/port name when greater than 50

### LINKED ISSUE #3104: Board and port names are cut off in the Tools menu.
The name of the selected items that now appear in the Tools menu are often cut off. For example, I see 'Board: "Arduino Duemilan..."'. I assume this is an attempt to keep the menu from getting too big, but it seems like we should at least make the maximum length long enough for the names we use. I think the longest one is "Arduino Duemilanove or Diecimila". 

The problem is perhaps worse for serial ports, at least on Mac, because the serial number of some ports is cut off, making it hard to distinguish between two different boards. For example, I see only "/dev/cu.usbmodem..." and not the 14511 suffix that tells me which Arduino it is.

ffissore:
Fixes #3104

Please try the IDE resulting from this PR and express your (dis)agreement for
1) the current behaviour, which truncates board/port names in order to avoid looooong names
2) the proposed behaviour, which doesn't truncate at all, even if that may lead to very long names
/cc @damellis

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -801,7 +801,7 @@ public class Editor extends JFrame implements RunnerListener {
             if (sel == null) {
               if (!name.equals(basename)) menu.setText(basename);
             } else {
-              if (sel.length() > 17) sel = sel.substring(0, 16) + "...";
+              if (sel.length() > 50) sel = sel.substring(0, 50) + "...";
               String newname = basename + ": \"" + sel + "\"";
               if (!name.equals(newname)) menu.setText(newname);
             }


PaulStoffregen:
I believe allowing unlimited length is unwise.  A very long string really mess up the menu.

My original decision to limit to 16 chars was conservative.  16 seemed like a lot more than zero.  Since this patch was "unofficial" for years, I didn't want to dramatically alter the menu size & appearance.

Let's keep the limit, but raise it to something large, like 25 or 30 or even 40 chars.

ffissore:
Agree. Let's turn it to a round 50

PaulStoffregen:
50 seems like as good a limit as any.  If we later do run into GUI problems, the limit can always be adjusted.

PR #3186, opened 2015-05-19
Added support for .hh and .hpp libraries

### LINKED ISSUE #2372: Allow .hh as header file extension
The Arduino library handling and build requires .h as extension of header files. It would be great to also allow .hh as extension as this is often used for C++ and is an important trigger for C++ mode in GNU Emacs. 

Best regards, Mikael

Chris--A:
This allows libraries to be imported correctly when they use .hh or .hpp extensions.

Resolves #2372


ffissore:
thanks!

mikaelpatel:
Great!

PR #3194, opened 2015-05-20
Got rid on JNA

ffissore:
Removed jna jars and every reference to it.
[Replaced](https://github.com/arduino/Arduino/pull/3194/files#diff-c2beb43483fc657f9dd1907703e83c93) the last two Platform methods with a call to external program `reg query` (windows only)
Testing its output with `RegQueryParserTest`


ffissore:
@pfeerick would you please try this build?

pfeerick:
Just tested this build on Windows x64, worked perfectly! Looks like you fixed the problem (or should I say removed it?!)

ffissore:
:) Best moment is when you remove stuff and things work better

PR #3207, opened 2015-05-21
Fixed bug with flushing

chromhelm:
I tried to make a RS485 bridge and i use Serial.flush() to wait for the serial transmission is complete.
but it turned out that this never happened.
I used this sketch.
You need a oscilloscope, to measure pin 2 to verify it.
 
void setup()
{
  Serial.begin(38400);
  Serial1.begin(38400);
  pinMode(2, OUTPUT);
}

void loop()
{
  if(Serial.available())
  {    
      digitalWrite(2, HIGH);
      Serial1.write(Serial.read());
      Serial1.flush();
      digitalWrite(2, LOW);
  }
  
  if(Serial1.available())Serial.write(Serial1.read());
}

### DIFF
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
@@ -220,6 +220,7 @@ size_t HardwareSerial::write(uint8_t c)
   if (_tx_buffer_head == _tx_buffer_tail && bit_is_set(*_ucsra, UDRE0)) {
     *_udr = c;
     sbi(*_ucsra, TXC0);
+     _written = true;
     return 1;
   }
   tx_buffer_index_t i = (_tx_buffer_head + 1) % SERIAL_TX_BUFFER_SIZE;


facchinm:
Hi @chromhelm , I took a look at your sketch; 
it seems that, shorting PIN 0 and 1 (Serial1 on Leonardo), the `flush()` works as expected and the text written in serial monitor is echoed back. Watching PIN 2 with a scope shows the correct behaviour.

Also, the `_written` variable you are setting to `true` inside `flush()` is only `false` between `begin()` and the first `write()`, so when your sketch executes the `flush()` the variable is already `true`.

Am I missing something?

chromhelm:
Hi facchinm

I forgot to say, that this fails only in core version 1.6, in core version 1.0 it works fine

facchinm:
I can't reproduce the issue but the fix is indeed correct. 
The bug was introduced with commit 9ad14b2f0cb15599.
Thank you very much!

matthijskooijman:
@chromhelm, good catch, thanks!

Might be even better to just change `_written` at the top of `write()` instead of in two places as now? Might save a few bytes of flash space.

chromhelm:
Yes, i made a new pull request with your suggestion. @matthijskooijman 
 Saving same bytes in HardwareSerial::write #3254

NicoHood:
This crashes my Arduino.Using a custom 1.6.6 build of a few days ago.

```cpp

// constants won't change. Used here to set a pin number :
const int ledPin =  13;      // the number of the LED pin

// Variables will change :
int ledState = LOW;             // ledState used to set the LED

// Generally, you should use "unsigned long" for variables that hold time
// The value will quickly become too large for an int to store
unsigned long previousMillis = 0;        // will store last time LED was updated

void setup() {
  // set the digital pin as output:
  pinMode(ledPin, OUTPUT);

  Serial.begin(2000000);
}

void loop() {
  unsigned long currentMillis = millis();

  if (currentMillis - previousMillis >= 1000) {
    // if the LED is off turn it on and vice-versa:
    if (ledState == LOW) {
      ledState = HIGH;
    } else {
      ledState = LOW;
    }

    for (int i = 0; i < 50; i++) {
      Serial.write('a');
      Serial.flush();
      delay(6);
    }

    // save the last time you blinked the LED
    previousMillis = millis();

    // set the LED with the ledState of the variable:
    digitalWrite(ledPin, ledState);
  }
}
```

NicoHood:
A different (quick) test so see that the flush() function causes this problem:
```cpp


const int dledPin =  13;
const int ledPin =  12;
bool ledState = LOW;
unsigned long previousMillis = 0;

void setup() {
  pinMode(ledPin, OUTPUT);
  pinMode(dledPin, OUTPUT);
  digitalWrite(dledPin, 0);
  Serial.begin(2000000);
}

void loop() {
  unsigned long currentMillis = millis();

  if ((currentMillis - previousMillis) >= 1000) {

    ledState = !ledState;

    digitalWrite(dledPin, 1);
    for (int i = 0; i < 250; i++) {

      Serial.write('a');
      digitalWrite(dledPin, 0);
      Serial.flush();
      digitalWrite(dledPin, 1);
      delay(6);
    }
    digitalWrite(dledPin, 1);

    previousMillis = millis();

    digitalWrite(ledPin, ledState);

  }
}
```

monitor with a 2nd arduino:
```cpp
void setup() {
  Serial.begin(0);
  Serial1.begin(2000000);
}

void loop() {
  if (Serial1.available()) {
    Serial1.read();
    static int c = 0;
    c++;
    Serial.println(c);
  }
}
#if SERIAL_RX_BUFFER_SIZE <200
#error
#endif
```

this count up to 73 for me. Could be because the receiver misses some bytes. I gave the receiver 256 bytes RX buffer so it is capable of all bytes. Its seems that the 73 is near the sending 64 bytes TX buffer.

I used an arduino uno as sender and the 16u2 (with hoodloader) as receiver. You can use any other external serial device/arduino for this as well.

NicoHood:
I know the next sketch is not that clear, but it show that the flush function stays in the while loop.

```cpp

const int dledPin =  12;
const int ledPin =  13;
bool ledState = LOW;
unsigned long previousMillis = 0;

void setup() {
  pinMode(ledPin, OUTPUT);
  pinMode(dledPin, OUTPUT);
  pinMode(11, OUTPUT);
  digitalWrite(dledPin, 0);
  digitalWrite(ledPin, 0);
  digitalWrite(11, 0);
  delay(100);
  digitalWrite(dledPin, 1); //ch 5
  delay(100);
  digitalWrite(ledPin, 1); // ch4
  delay(100);
  digitalWrite(11, 1); //ch6
  delay(1000);
  digitalWrite(dledPin, 0);
  digitalWrite(ledPin, 0);
  digitalWrite(11, 0);

  Serial.begin(2000000);
  Serial.write('b');
}

void loop() {
  unsigned long currentMillis = millis();

  if ((currentMillis - previousMillis) >= 1000) {

    ledState = !ledState;

    digitalWrite(dledPin, 1);
    for (int i = 0; i < 250; i++) {

      Serial.write('a');
      PORTB |= (1 << 4);
      Serial.flush();
      PORTB &= ~(1 << 4);
      delay(6);
    }

    previousMillis = millis();

    digitalWrite(ledPin, ledState);

  }
}
```

Edit in hwserial.cpp:
```cpp
void HardwareSerial::flush()
{
  // If we have never written a byte, no need to flush. This special
  // case is needed since there is no way to force the TXC (transmit
  // complete) bit to 1 during initialization
  if (!_written)
    return;

  while (bit_is_set(*_ucsrb, UDRIE0) || bit_is_clear(*_ucsra, TXC0)) {
  PORTB |= (1<<3);
    if (bit_is_clear(SREG, SREG_I) && bit_is_set(*_ucsrb, UDRIE0))
	// Interrupts are globally disabled, but the DR empty
	// interrupt should be enabled, so poll the DR empty flag to
	// prevent deadlock
	if (bit_is_set(*_ucsra, UDRE0))
	  _tx_udr_empty_irq();
	  
	  PORTB &= ~(1<<3);
  }
  // If we get here, nothing is queued anymore (DRIE is disabled) and
  // the hardware finished tranmission (TXC is set).
}
```

screenshot from 2015-09-13 19 49 00

You can see that channel 5 keeps high (loop stuck in flush() function). After that Channel 6 keeps switching the state (while inside flush).

**So there must be something wrong with the while loop inside flush().**

NicoHood:
Also if the Serial TX buffer is >256 this needs to be inside a cli() sei():
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp#L243

Since I tested with 64 bytes this should be fine for the issue above, but can cause errors for bigger buffers.

Issue #884, opened 2012-11-15
Optimization in print.cpp [imported]


cmaglie:
This is [Issue 884](http://code.google.com/p/arduino/issues/detail?id=884) moved from a Google Code project.
Added by 2012-04-11T11:57:21.000Z by [rob.till...@gmail.com](http://code.google.com/u/113198963792305667520/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium

### Original description

print.cpp contains 

```c++
size_t Print::println(void)
{
  size_t n = print('\r');
  n += print('\n');
  return n;
}
```

which can be optimized to

```c++
size_t Print::println(void)
{
   return write ("\r\n");
}
```

It is simpler and it has less call overhead.

For a background discussion, see: 
- http://arduino.cc/forum/index.php/topic,100712.0.html -

facchinm:
Fixed with #3211

PR #3211, opened 2015-05-22
Println optimize

Chris--A:
This PR resolves issue #884

The sketch below dropped 30 bytes. The increase to ram is only 2 bytes (AVR). The changes aren't as big on the SAM core, but still show a slight improvement in code size.

```arduino
/***

  Standard:
    Sketch uses 5,634 bytes (18%) of program storage space. Maximum is 32,256 bytes.
    Global variables use 228 bytes (11%) of dynamic memory, leaving 1,820 bytes for local variables. Maximum is 2,048 bytes.
     
  Proposed:
    Sketch uses 5,604 bytes (17%) of program storage space. Maximum is 32,256 bytes.
    Global variables use 230 bytes (11%) of dynamic memory, leaving 1,818 bytes for local variables. Maximum is 2,048 bytes.
  
***/

void setup(){
  Serial.begin( 9600 );
  Serial.println( F("Printing a flash string!") );
  Serial.println( 12.34567f, 4 );
  Serial.println( "A plain string" );
  Serial.println( 'a' );
  Serial.println( String( "A String" ) );
  Serial.println( 123456, BIN );
  Serial.println( 123456, OCT );
  Serial.println( 123456, DEC );
  Serial.println( 123456, HEX );
}

void loop(){}
```

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -122,9 +122,7 @@ size_t Print::print(const Printable& x)
 
 size_t Print::println(void)
 {
-  size_t n = print('\r');
-  n += print('\n');
-  return n;
+  return write("\r\n");
 }
 
 size_t Print::println(const String &s)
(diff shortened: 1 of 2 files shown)


facchinm:
Merged, thank you!

PR #3225, opened 2015-05-25
MacOSX: attempting disabling OpenGL rendering pipeline

ffissore:
Some Macs have a dual GPU setup and the IDE (as well as other programs btw) is turning on the most powerful one, resulting in a shorter battery life


ffissore:
@ArduinoBot build this please

ffissore:
User from http://forum.arduino.cc/index.php?topic=321263 never replied and apparently these kind of macs are not widespread

Issue #3174, opened 2015-05-17
[Ubuntu 1.6.4] Xorg crashes when scrolling down the boards manager


NicoHood:
I have several boards installed within my sketchbook/hardware folder. Tried with my HoodLoader2, HID-Project and the sparkfun folder.

Thats a lot of boards, so you have to scroll down to reach the last entry. However if you try to scroll down the whole X session crashes and you have to login again. Every program like browser crashes as well.

Tried with lastest standard Ubuntu 64 bit, IDE 1.6.4 and java7. I did not test any older release yet.

ffissore:
Does it happen with latest nightly?

NicoHood:
Yes, problem still there. X_x

ffissore:
Confirmed. Veeery annoying.

ffissore:
Fixed. Fix will be available with next hourly build

ffissore:
Nope, bug is still there

ffissore:
Various attempts to solve the issue failed. However, when running the IDE with latest java 8, the issues did not show up. I then suggest you to install latest java 8 following [this tutorial for ubuntu](http://ubuntuhandbook.org/index.php/2015/01/install-openjdk-8-ubuntu-14-04-12-04-lts/) or downloading it directly from [oracle website](http://www.oracle.com/technetwork/java/javase/downloads/index.html) and changing the `java` command [here](https://github.com/arduino/Arduino/blob/master/build/linux/dist/arduino#L31) specifying the full path (mine is `/opt/jvm/jdk1.8.0/bin/java`)

NicoHood:
The problem is still there.
Telling the people to install java8 and pretend the bug is fixed is not a solution. I suggest two things:

Add a note to the installing instructions/download section that you need java 8 (not tested for me though)
Or open the issue and give away some cash if someone else is able to fix it.

Its a critical bug, all programs crash, all data is gone because of this fucking IDE. If someone accidentally gets this bug, this will drive him crazy. Especially when you havent saved your code/docs.

I noticed some changes in the nightly (of course). It would be nice if you could pin the Arduino IDE to the Ubuntu launcher as well. I've created a shortcut myself but it will open a 2nd Icon on the launcher.

ffissore:
Please avoid using harsh words. If the problem lies in how the jvm handles graphics, we can't do much but ask to use an updated jvm.

NicoHood:
[...]

PR #3226, opened 2015-05-25
Replaced/improved Linux install and uninstall scripts.

fiechr:
This install script adds a menu item, desktop icon and file association (mime type) for the current user on Linux - even if the Arduino installation is placed outside of the user's home dir (such as in /opt or /usr/local) - which didn't worked before.

If the xdg-utils aren't available on the target system, a more basic approach is used (just generate and copy the file).

The uninstall script is a simple wrapper and calls the install script to undo the changes.

Also added three more icon resolutions (64px, 72px, 96px) - by using the 256px file - maybe there is a better (vector based) source available for exporting to these resolutions...

Modified build.xml for copying the existing/added icons and existing mime.xml file, both changed/added scripts and adjusted file permissions (ugo+x) for all *.sh files to be executable by others by default after archive extraction. A more conservative approach would be to set only install/uninstall.sh ugo+x.

I tested this on different machines and it works fine so far.


ffissore:
@ArduinoBot build this please

fiechr:
Tested the 32bit and 64bit Linux builds in- and outside the users home dir. Works as expected.

NicoHood:
Is there any documenation on how to run those scripts?
Do I have to call the 1.6.3 uninstall script if I have 1.6.3 installed or can I use the 1.6.4 uninstall script or even run the install script without uninstalling?

fiechr:
Basically install.sh will add entries and icons for the *current* user, uninstall.sh will remove them. There shouldn't be any way to break anything which can't be restored by the script itself.

There is no need to uninstall first, because these these functions are called anyway when installing.
If there is an existing menu entry created by the current install script, it will be removed, too.

NicoHood:
I've downloaded the PR and only added those two .sh files to my current nightly built (from today).
It deleted my menu bar icon which I created on my own. It added a desktop icon which cannot be executed and has no icon.

Before I added a .desktop entry into the application dir of ubuntu myself. It worked and I could pin it to the menu bar. It opened another instance which wasnt perfect but it worked.

Not sure what I am doing wrong.

fiechr:
To copy the scripts won't be enough. You also need the icon dir and the desktop.template file in the lib/ subdir. Both are in the pull request, too.
I only tested the complete build with all changes applied, so I can't guarantee if it will work this way - but it should.

NicoHood:
The desktop icon is now working. I moved this to the menu bar as well.

The menu item is opening another instance of the IDE once again.
So I have the shortcut and the actual instance in my menu bar, 2 icons.

How do I fix this?

fiechr:
The install script adds two independent links. One in the general apps menu (Unity Dash menu for Ubuntu) and an additional one onto the desktop.

The script does *not* create a shortcut in the launcher. If the user wants it there, he/she has to start the app via the menu and then choose "lock to launcher" (or similar). This part is entirely distribution depended (in this case Ubuntu Unity).

If you drop the desktop icon into the launcher, its like adding a second app with the same name. So the behavior you describe is no surprise.

Personally I already thought of not adding the desktop icon at all, because it's kind of redundant anyway.

NicoHood:
There is no "lock to launcher" option. I can only drag an drop and this opens a second icon.

fiechr:
You have to right click on the icon after you opened the program from the Dash menu:
arduinoideunitlauncher

NicoHood:
Sadly i cannot create a screenshot when the sebmenu is opened.
Where you have "Arduino IDE" I have "sketch_jun05a [...]" like the name in the opened window. And if O lock this icon the IDE wont start on a click. If I move the desktop icon there it works but opens another instance as described in my last post.

fiechr:
Hmm... this is interesting. This seems to be a Java related issue, like it is described here:
http://askubuntu.com/a/367851

> Many Java-based applications have the problem where the windows created by the application are not associated with the shortcut used to start the application (so the dock contains multiple copies of the same icon).

"Unfortunately" my Arduino windows already get automatically grouped by Unity if I open more than one sketch. Maybe because I use the Sun JDK instead of the OpenJDK? I'm using Ubuntu 14.04.

Using xprop on the Arduino window gives me:

```
$ xprop WM_CLASS
WM_CLASS(STRING) = "sun-awt-X11-XFramePeer", "processing-app-Base"

```

You could try adding
```
StartupWMClass=processing-app-Base
```
to the arduino-arduinoide.desktop file in ~/.local/share/applications

If this helps, I could add this to the .desktop file.

NicoHood:
I am using java 8 due to this bug (used the linked installation instructions) https://github.com/arduino/Arduino/issues/3174
I am also using the stable Ubuntu release which you can download from the website. I thinks it has a lower version number than yours.

```
 xprop WM_CLASS
WM_CLASS(STRING) = "sun-awt-X11-XFramePeer", "processing-app-Base"
```

`StartupWMClass=processing-app-Base`

This helps!

NicoHood:
.ino files now open with the IDE, it would be nice to add the icon to the file as well (if possible).
Is there also a recommended location where the IDE should be placed? I have it somewhere in my home folder.

fiechr:
AFAIK file icons can only be changed by writing to a dir outside the user's home dir - which I wanted to avoid, because everything else works fine without root access. Also I would need a scalable image file (SVG) of the used IDE icon for this.

I guess most people put it in the home folder - which is absolutely fine. I like to put additional "external" software in /opt - but that's mostly a matter of taste.

NicoHood:
So I have to put the arduino folder there and run the script from this location again?
/opt makes totally sense. Thanks

So the script now works fine. I havent checked the script itself since I dont know enough about scripts. But it works and should be added to the IDE.

fiechr:
> So I have to put the arduino folder there and run the script from this location again?

Yes, that should do it.

NicoHood:
@ffissore can you please review this so we see it in the new 1.6.5 release?

matthijskooijman:
@ArduinoBot build this please

NicoHood:
Works perfectly out of the box for me (elementary os, based on ubuntu).
Would like to see this in the new release!

NicoHood:
Why got this removed from the milestone? Its such a useful fix, I really like to see this!

fiechr:
Anything I need to fix/change?

NicoHood:
Cant anyone review this?

matthijskooijman:
I just reviewed the code, looks ok to me. I wonder if the "legacy" code really needs to be kept around in this script? I'd say that if you want to use XDG-style shortcuts, it is ok to just require xdg-utils to be installed. If you really don't want xdg-utils, you can still just create the .desktop manually. @cmaglie, any view on this?

Regarding the startup wmclass thing, having that set to "processing" probably isn't ideal, since it might conflict with the actual processing. Any idea if we can change that? Or is that tied to the main class name perhaps?

Finally, I think the script should support the `-h` and `--help` options. When I have an unknown script, I usually run it with `--help` to see what it will be doing exactly and I expect either help output, or an error that `--help` isn't supported. Right now, the script just proceeds with installing, which isn't so great.

I just tested the current build on Debian Jessie, which works as expected :-)

NicoHood:
I aggree with the -h option.

The script works perfectly for me. Ive used it once and the shortcut just keeps working since then, even with newer builds. I used Elementary OS here, based on ubuntu, based on debian.

fiechr:
> Regarding the startup wmclass thing, having that set to "processing" probably isn't ideal, since it might conflict with the actual processing. Any idea if we can change that? Or is that tied to the main class name perhaps?

Yeah, I agree, that bothered me a little bit, too - it's definitely not ideal. But it seems the only reliable ID a (Java) window/app can be identified by. I have no idea how much effort it would be to change this.

> I just reviewed the code, looks ok to me. I wonder if the "legacy" code really needs to be kept around in this script?

The main reason I put this in in the first place, is for removing existing links/symbols created by the existing install script.
But I also don't know how many distributions come with XDG tools (installed), so I wanted to make sure that a least something useful is created - which, I guess, is better than giving an error message or do nothing.

About the ```--help / -h``` option:
I also think that's a good idea - I will add this. :-)

(First I thought about asking the user before doing anything - but then user interaction would be required - and I also don't wanted it to be too complicated.)

matthijskooijman:
> The main reason I put this in in the first place, is for removing existing links/symbols created by the existing install script.
It doesn't seem unreasonable to expect people to uninstall the old version before installing this one, though it might be good to also uninstall it before installing, indeed.

> But I also don't know how many distributions come with XDG tools (installed), so I wanted to make sure that a least something useful is created - which, I guess, is better than giving an error message or do nothing.

Having an error message that says "please install xdg-utils" seems reasonable to me? The end result is actually _better_ than falling back to a short-cut-only install.

matthijskooijman:
> About the --help / -h option:
I also think that's a good idea - I will add this. :-)

It's probably a good idea to also error out when any unknown options are given?

fiechr:
I added a help option:
```
$ ./install -h

This script will add an Arduino IDE menu entry, icons and
file associations for the current user.

Optional arguments are:

	-u, --uninstall          Removes icons and menu item.

	-h, --help               Shows this help again.
```

All other arguments will produce an error:
```
$ ./install.sh --unknown

Invalid option -- '--unknown'
```
(and after this show the help above)

If xdg-utils isn't found on the system the help will also show a note that they are recommended.

matthijskooijman:
Help option looks good to me!

jonnius:
Please fix issue #4168, too! In your PR it still persists.

fiechr:
Thanks! I will look into it!

fiechr:
I changed the script so it will (basically) use the following method to find the correct Desktop directory:
http://www.freedesktop.org/wiki/Software/xdg-user-dirs/
This should also work for different languages/names of this folder.
If nothing is found this way, it will fall back to ~/Desktop (if it exists).

This change only affects the "simple install" mode, because the default install using the xdg-utils should already figure out the correct target directory by itself.

It would be nice if someone could test this on a system where the desktop folder's name isn't "Desktop" (and/or isn't placed in $HOME).

If you already have xdg-utils installed and also want to test the script without it, just edit the install.sh script file and replace
```return 0```
with
```return 1```
on line 182.

**Comment by jonnius** [comment]
The path is correct now. But the created desktop file on desktop is empty. Here is the output: https://gist.github.com/jonnius/1a4c75ba012b8baab090

**Comment by fiechr** [comment]
Just to be sure: Can you shortly describe how you tested this? Did you just take the script files or rebuild the whole package including all changes? Because there are other files included which are needed.

For example:
sed claims "[...]/Arduino/build/linux/dist/lib/desktop.template" can't be found. But it's part of the Arduino package/archive (of this PR) - so it *has* to be there.

The same is true for all the icon files and the mime xml file.

**Comment by jonnius** [comment]
I cloned this branch and just started the script. 
 * The icons seem to be located in build/shared/icons, but it is looking for it in build/linux/dist/lib/icons. 
 * The desktop.template is located in build/linux/dist/desktop.template, but it is looking for build/linux/dist/lib/desktop.template. 
 * I cannot find any arduino-arduinoide.xml in the branch at all...

**Comment by fiechr** [comment]
Ok, that explains it. You need to build the dist package first!

Go to Arduino/build in the cloned repo/branch and start the provided build-all script. This will create two archives in the Arduino dir (32 and 64bit) which are the archives you want to install/extract (like the official ones you can download on the website). You only need one of these - depending on your system.
The script will also try to build the Windows version, which will most likely fail - just ignore that.

**Comment by jonnius** [comment]
I ran the build-all script and as you mentioned, linux build was successful and windows build was not. But after that I still have the same issue as before.

**Comment by fiechr** [comment]
Where exactly did you extract the created archive? If it's really the exact same error as before, then it seems you are still starting the script from within the repository structure.

**Comment by jonnius** [comment]
Sorry, i was in the wrong directory. I can confirm that it works as expected! Thanks!

**Comment by ffissore** [comment]
@fiechr I've rebased your PR and pushed it to master. It will be available in next hourly build http://www.arduino.cc/en/Main/Software#hourly. Thank you very much for your time and patience

**Comment by NicoHood** [comment]
Except that my raspberry pi (arm) IDE doesnt start at all the install script adds a few warnings:
```
pi@raspberrypi ~/Documents/Arduino/Arduino/build $ ./linux/work/install.sh 
Adding desktop shortcut, menu item and file associations for Arduino IDE...touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
touch: cannot touch ‘/home/pi/.local/share/icons/hicolor/.xdg-icon-resource-dummy’: No such file or directory
 done!
```

Might be related to the LXDE desktop, idk. Just want to mention it here. Hope I can get https://github.com/arduino/Arduino/pull/3549 running some day.

NicoHood:
I currently try to update the [AUR arduino package](https://aur.archlinux.org/packages/arduino/) and got a few questions:
* Which license is the correct for Arduino? GPL2?
* Which dependencies does Arduino have? I think no java is required at all and you only need gtk2 at least. Is this correct?

Could anyone please have a look at the file here:
https://gist.github.com/NicoHood/a50936c9c8af23a889c3176c5ddbe131/revisions

Issue #220, opened 2012-11-15
Increase/decrease indent doesn't work on OS X with swedish keyboard layout. [imported]


cmaglie:
This is [Issue 220](http://code.google.com/p/arduino/issues/detail?id=220) moved from a Google Code project.
Added by 2010-03-15T19:40:59.000Z by [krook.jo...@gmail.com](http://code.google.com/u/107155206166980269532/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium

### Original description

**What steps will reproduce the problem?**
1. Press Command
2. Press Alt
3. Press 8/9

**What is the expected output? What do you see instead?**

The row or all selected text should change indentation. Instead, nothing happens.

**What version of the Arduino software are you using? On what operating**
**system?  Which Arduino board are you using?**

Arduino 0018 on OS X with swedish keyboard.

**Please provide any additional information below.**

Seems like the problem is with shortcuts which need Alt to be pressed. The shortcut for 
comment/uncomment works fine. That is: 
1. Command 
2. Shift 
3. 7. 
A possible solution could be to change shortcuts to Command-( and Command-). 

The issue is similar to http://code.google.com/p/arduino/issues/detail?id=76

ffissore:
Fixed with #3229

PR #3229, opened 2015-05-25
Stop using [ and ] as indentation menu accelerators

### LINKED ISSUE #3224: "Decrease Indent" shortcut doesn't work in IDE 1.6.5
I'm not sure this issue already raised or not.

In the edit window, the shortcut key "Ctrl" + "[" (Ctrl + Open Bracket) doesn't work.
But both of the Edit -> Decrease Indent menu and Decrease Indent of the popup menu work fine.

By the way, are there any way to localize shortcut key names: 

- Ctrl + Slash                -> Ctrl + / 
- Ctrl + Open Bracket   -> Ctrl + [
- Ctrl + Close Bracket   -> Ctrl + ]

ffissore:
Use TAB/SHIFT+TAB instead. Fixes #3224 and #220
This problem affects Processing as well. See https://github.com/processing/processing/issues/2199
Brackets may work on US keyboards, although no editor that I know uses them. TAB is a much more popular and widely accepted shortcut for indentation

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1420,7 +1420,8 @@ public class Editor extends JFrame implements RunnerListener {
     });
     menu.add(commentItem);
 
-    JMenuItem increaseIndentItem = newJMenuItem(_("Increase Indent"), ']');
+    JMenuItem increaseIndentItem = new JMenuItem(_("Increase Indent"));
+    increaseIndentItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_TAB, 0));
     increaseIndentItem.addActionListener(new ActionListener() {
         public void actionPerformed(ActionEvent e) {
           handleIndentOutdent(true);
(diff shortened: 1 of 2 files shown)


damellis:
Xcode uses Cmd+[ and Cmd+] for indent and unindent. I'm a fan of them too. Any reason not to support both them and tab / shift-tab?

ffissore:
We can support both but we can only write one on the edit menu

ffissore:
What about this? CMD+brakets are still available so, if you come from XCode, they'll just work. However TAB|SHIFT TAB are advertized on the Edit menu

Jed-SWSDT:
At least with version 1.6.5, the functionality of [ and ] is still not correct. It used to move the entire line regardless of where the cursor was in it. Now that only works for decreasing. For increasing it adds a tab where the cursor is. I understand that this may not have been a standard feature across editors but some of us relied on it. It was an easy way to change the line indent without having to move the cursor to the beginning (requiring another key combo or mouse move). Is there any plan on returning the functionality to its original intent or are the [ and ] now just going to be a different way of hitting tab/shift-tab?

ffissore:
Yes, please open a dedicated issue where you mention that CMD+[ is not increasing indentation but rather inserting a tab

PR #3239, opened 2015-05-27
MacOSX: CMD+SHIFT-LEFT/RIGHT selects the entire line. Fixes #3238

### LINKED ISSUE #3238: CMD+SHIFT+ARROW should select text on Mac OS X
I believe that the keyboard shortcut for selecting text is wrong on Mac OS X after the introduction of RSyntaxTextArea.

Currently the shortcut is FN+SHIFT+ARROW, but in Arduino 1.6.4 and below it was CMD+SHIFT+ARROW.

ffissore:
/cc @Lauszus

### DIFF
--- a/app/src/processing/app/syntax/SketchTextAreaDefaultInputMap.java
+++ b/app/src/processing/app/syntax/SketchTextAreaDefaultInputMap.java
@@ -46,6 +46,9 @@ public class SketchTextAreaDefaultInputMap extends RSyntaxTextAreaDefaultInputMa
 
       put(KeyStroke.getKeyStroke(KeyEvent.VK_OPEN_BRACKET, defaultModifier), DefaultEditorKit.insertTabAction);
       put(KeyStroke.getKeyStroke(KeyEvent.VK_CLOSE_BRACKET, defaultModifier), RSyntaxTextAreaEditorKit.rstaDecreaseIndentAction);
+
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_UP, defaultModifier | shift), DefaultEditorKit.selectionBeginAction);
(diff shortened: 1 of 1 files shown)


Lauszus:
@ffissore yes that solves it, but I also think it would be a good idea to add support for CMD+SHIFT+UP/DOWN while we are at it.

ffissore:
@Lauszus what do CMD+SHIFT+UP/DOWN do?

Lauszus:
It should select all text above and below the cursor respectively.

ffissore:
It should already work. I'll know a colleague and steal her mac

ffissore:
Fixed. @ArduinoBot will prepare a build

Lauszus:
Haha @ffissore it is no nice to steal ;)

Anyway it is still not working 100% correct.

I have taken a couple of screenshot in order to explain the current behaviour and what is wrong with it.

Cursor was moved to the end of the line and CMD+SHIFT+LEFT was pressed:
screen1

Then I pressed UP while still holding CMD+SHIFT, instead of continuing the selecting it starts from the cursor position instead:
screen2

This is how it should look like:
screen3

Note I got the last screenshot by manually selecting the text.

Lauszus:
Btw is there anyway to just tell it to use the same behaviour as FN+SHIFT+ARROW, as it is working correctly?

ffissore:
Keyboard mappings are defined in https://github.com/arduino/Arduino/blob/master/app/src/processing/app/syntax/SketchTextAreaDefaultInputMap.java and superclasses. I can't spot FN: I can only see shift, alt, cmd, ctrl.
I've merged this PR because, at least, the implemented behaviour is now more similar to the desired one

Issue #956, opened 2012-11-15
Editor does not handle backslash at line end of comments [imported]


cmaglie:
This is [Issue 956](http://code.google.com/p/arduino/issues/detail?id=956) moved from a Google Code project.
Added by 2012-06-11T10:01:57.000Z by [meiklehm...@gmail.com](http://code.google.com/u/109795282464955675286/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium, Component-Core

### Original description

I'm using Arduino 1.0.1 and added some comments to a .h file to describe some constants. Some lines contained a backslash at the end. I was surprised that my software did not work correctly. It took me a long time to find out that the trailing backslash was the reason because in the case the next line also will be treated as comment. It would have been much easier for me to find this out if the syntax coloring would have shown the next line in "comment color", too.

Example:
unsigned char values[] = {
  45, 12, 56, // \
  45, 13, 57,
  45, 13, 58
};

will be treated as

unsigned char values[] = {
  45, 12, 56,
  45, 13, 58
};

by the compiler.

Chris--A:
@ffissore 

In the latest build this does not appear to be the case; the sketch below compiles fine, however its an error. It should cause a compile error and fail.

```arduino

unsigned char values[] = {
45, 12, 56, // \
45, 13, 57,
45, 13, 58
};

unsigned char values1[] = {
45, 12, 56,
45, 13, 57,
45, 13, 58
};

bool test[ sizeof( values ) == sizeof( values1 ) ? 1 : 0x7FFFFFFFFFFFFFFFULL ];

void setup() {}
void loop() {}
```


On looking at the pre-processed file, it appears Arduino has removed the comment, and as a result it has un-commented the next line.

So I believe this is now a bug in the IDE pre-processor.

```arduino
#line 1 "sketch_may27a.ino"

#include "Arduino.h"
void setup();
void loop();
#line 2
unsigned char values[] = {
45, 12, 56,       
45, 13, 57,
45, 13, 58
};

unsigned char values1[] = {
45, 12, 56,
45, 13, 57,
45, 13, 58
};

bool test[ sizeof( values ) == sizeof( values1 ) ? 1 : 0x7FFFFFFFFFFFFFFFULL ];

void setup() {}
void loop() {}
```

ffissore:
Yet another preprocessor issue

ffissore:
Thanks for the update @Chris--A

[...]

PR #3240, opened 2015-05-27
Fix for single line comments ending in \﻿

Chris--A:
This is a fix for #956 

Comments are still stripped, however, single line comments with a multi line directive are replaced with spaces while preserving the comment and multi line slash.

```arduino

//A standard comment
unsigned char values[] = {
45, 12, 56, // A comment with a multi line pre-processor \
45, 13, 57,
45, 13, 58
};

/***
 * 
 * A fancy comment
 * 
 */

void setup() {/* An inline comment */}
void loop() {}
```

After pre-processing:

```arduino
#line 1 "comment_error.ino"

                    
#include "Arduino.h"
void setup();
void loop();
#line 3
unsigned char values[] = {
45, 12, 56,                                            //\
45, 13, 57,
45, 13, 58
};

    
   
                  
   
   

void setup() {                       }
void loop() {}

```

### DIFF
--- a/arduino-core/src/processing/app/preproc/PdePreprocessor.java
+++ b/arduino-core/src/processing/app/preproc/PdePreprocessor.java
@@ -394,10 +394,24 @@ public class PdePreprocessor {
           (p[index+1] == '/')) {
         p[index++] = ' ';
         p[index++] = ' ';
-        while ((index < p.length) &&
-               (p[index] != '\n')) {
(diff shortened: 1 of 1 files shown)


ffissore:
PdePreprocessor is a pit I don't want to dig further. #2729 fixes many of its issues but it's also outdated. I'll work on it next week. Thank you all the same

PR #3221, opened 2015-05-25
Add missing case to detachinterrupt

vicatcu:
Addresses issue #1191 and issue #879. Adds missing case to detachInterrupt for INT2.

PR #3244, opened 2015-05-28
Check if INT2 is actually defined before issuing warning in detachInterrupt

Lauszus:
This was caused after: https://github.com/arduino/Arduino/pull/3221


cmaglie:
Thanks!

PR #3254, opened 2015-05-29
Saving same bytes in HardwareSerial::write

chromhelm:


cmaglie:
Manually rebased and merged, thanks @chromhelm!

PR #3269, opened 2015-06-01
Add a Stream::find(char) method

### LINKED ISSUE #847: Add a Stream::find(char) method [imported]
This is [Issue 847](http://code.google.com/p/arduino/issues/detail?id=847) moved from a Google Code project.
Added by 2012-03-06T16:59:23.000Z by [pierre.b...@gmail.com](http://code.google.com/u/100150763547863412928/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Enhancement, Priority-Medium

### Original description

**What change would like to see?**
find(char) doesn't exist, which means you need to wrap a single character as a string.

**Why?**
The current situation is completely fine with literals - e.g find("$") instead of find('$') - but it can get unnecessarily complex with variables.

For instance you can't write:
char target;
...
find(target);

Instead you need to do something along the lines of:
char target;
...
char buf[2];
buf[0] = target;
buf[1] = 0x00;
find(buf);

facchinm:
Thanks to #2696, adding the method is trivial
Fixes #847 and replaces #65

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Stream.h
+++ b/hardware/arduino/avr/cores/arduino/Stream.h
@@ -64,6 +64,8 @@ class Stream : public Print
   bool find(uint8_t *target, size_t length) { return find ((char *)target, length); }
   // returns true if target string is found, false if timed out
 
+  bool find(char target) { return find (&target, 1); }
+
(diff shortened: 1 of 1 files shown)


Chris--A:
This is missing the SAM version, available here: #3337

Issue #3274, opened 2015-06-02
YunClient's connected() returns false with unread bytes


RobAtticus:
According to its documentation, YunClient should return true with unread bytes:
http://www.arduino.cc/en/Reference/YunClientConnected

However, if the connection is closed at a TCP level, it will return false even with bytes left unread.

RobAtticus:
I'm not sure this is a documentation problem as other clients, e.g., WifiClient, have the same documentation. I believe the YunClient does not conform to other Clients.

RobAtticus:
I have submitted a PR to address this and another issue I found:
https://github.com/arduino/Arduino/pull/3288

cmaglie:
Fixed by #3288

PR #3288, opened 2015-06-04
Fixes to YunClient in connected() and stop()

RobAtticus:
This pull request fixes YunClient to better conform to other Clients like EthernetClient and WiFiClient:

1) `connected()` now returns true if there are unread bytes, like other Clients and the docs say.
2) `stop()` now resets `buffered` and `readPos` so reusing a client with unread bytes does not cause those old bytes to be read.

### DIFF
--- a/libraries/Bridge/src/YunClient.cpp
+++ b/libraries/Bridge/src/YunClient.cpp
@@ -114,6 +114,10 @@ void YunClient::flush() {
 uint8_t YunClient::connected() {
   if (!opened)
     return false;
+  // Client is "connected" if it has unread bytes
+  if (available())
+    return true;
+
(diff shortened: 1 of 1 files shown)


RobAtticus:
This address issue https://github.com/arduino/Arduino/issues/3274

RobAtticus:
Hi, is there a reason this got bumped from 1.6.5? I would very much like to start being able to safely use these methods in work I'm currently doing.

cmaglie:
Manually merged into master:
2774f5a4c9db5d981b74dc8ce70e8346857285dc
6f566776f9b74dfc25a62a0121bbb6bb3accdcbc

Thank you!

PR #2385, opened 2014-10-23
support for Dangerous Prototypes Bus Pirate as ISP

ziotibia81:

PR #3292, opened 2015-06-05
Add support for Bus Pirate as a programmer.

liyanage:
This adds support for Dangerous Prototypes' Bus Pirate as a programmer. Avrdude already supports this, so this is just a small configuration file change.


matthijskooijman:
This is a duplicate of #2385. The change is pretty much identical, except the other PR has some verbose comments.

liyanage:
Glad to hear it's already tracked. I see this bug was assigned to the 1.6.5 milestone, but the original isn't. Is that going to make it into 1.6.5?

cmaglie:
Done!

Issue #3080, opened 2015-05-05
Add sketch subdir in the Include Path


Testato:
Actually is possible include in the sketch an header file only if it is in:

pippo\pippo.ino
pippo\x.h
c:\x.h

Please add the possibilities of use subdir in the sketch:

pippo\somedir\x.h

ffissore:
Can you also code and [put on gist](http://gist.github.com/) a short example sketch that we can use to test the feature?

MauroMombelli:
see #1152, the sdlib is a perfet example that actually works because a special case in the code

matthijskooijman:
Seems you might be mixing up things. For "new-style" libraries, source files in subdirectories of the src/ directory are compiled, but AFAIK these subdirectories are _not_ added to the include path. Doing so would actually be counter-productive - it would merge all those directories in a single namespace, greatly increasing the chance of naming conflicts.

If you want to include pippo\somedir\x.h from pippo.ino, you can just use `#include "somedir/x.h"`, which is nice and explicit, without relying on any magic include path. Are you not aware of this option, or is there something wrong with this option? Perhaps you mean something entirely different - you should really provide a more verbose explanation and/or  examples...

Testato:
#include "somedir/x.h"

this is exactly the think that do not work, i do not skeak about a library problem, but simply including header like your example. This is very usefull on big project.

I uploaded a simple sketch+subdir here:
https://drive.google.com/file/d/0Bw27MYXTb2FXQ0kxZXdMQkpmajg/view?usp=sharing

matthijskooijman:
Could it be a \ vs / issue? I think includes normally use /, even on Windows. Try using a / in your include too and tell use if that helps.

Testato:
same problem whit \ or /

Testato:
no, the problem is the same.

On Arduino IDE, at least on Windows but i think on linux also, there is no differenze:

you can do this:
#include "C:/test.h"

or this
#include "C:\test.h"

Try yourself whit my uploaded sketch

Testato:
this is the example in the uploaded file:

```C++
#include "test1.h"	    // This work
#include "C:\test2.h"       // This work
#include "SomeDir\test3.h"  // This is the ISSUE request

void setup() {
}

void loop() {
}
```

matthijskooijman:
[...]

PR #3303, opened 2015-06-08
RFC > Compiler: core files in "core" subfolder, libraries in "libraries"

ffissore:
With this PR, temporary build folder no longer gather all object files in one folder, flat. Core object files go to "core" folder, libraries to "libraries" (note: libraries already went to a folder name after the library name; I've added an additional folder level because it seemed more tidy **to me**

/cc @NicoHood @matthijskooijman @PaulStoffregen @cmaglie @facchinm


ffissore:
Please note this ugly line
https://github.com/arduino/Arduino/pull/3303/files#diff-9cd90eab2dcf151a63c144aff0e043b9R1149
It's needed becase SAM platform.txt hardcodes a file path: `{build.path}/syscalls_sam3.c.o`

ffissore:
Please give this PR a try using your cores

NicoHood:
What issue are you referring to? That the .ino file should not be names as a .cpp/.h file?

matthijskooijman:
Would it make sense to also have "sketch" folder and compiler the sketch in there? That would prevent problems when the sketch contains a subdirectory called "core" for example when fixing #3080 later.

ffissore:
:+1:

ffissore:
This latest build is the result of a force push: I've removed the java hardcoded replacement of that `{build.path}/syscalls_sam3.c.o` using `platform.keys.rewrite.txt` and I've put sketch sources and object files into `sketch` folder

Issue #2474, opened 2014-11-29
Critical Leonardo + Micro Bootkey conflicts


NicoHood:
The way we write the Bootloader key in the ram is not reliable. You can destroy important data since you can revert the bootloader reset. With a 32u4 it might be very unlikely (because of that much ram) but still possible and then people are wondering about weird bugs. And with a 16u2 (500b ram) it is even more likely that this will happen if the reset is reverted.

Referring to this lines:
https://github.com/arduino/Arduino/blob/master/hardware/arduino/cores/arduino/CDC.cpp#L94-L108

So what we need to do is to ensure this address is really NOT USED.

3 Solutions i have in mind:
Reserve the Bootkey address so its not used in any case. See this for a possible solution:
http://electronics.stackexchange.com/questions/140516/where-to-put-bootloader-key-in-ram-what-address-is-the-last-ram-address

The other solution would be to force a reset in any case or add some other debounce routines.

The last might be the best but 'biggest' change. Also save the value to a fixed address, add this to the bootloader and Arduino core. Shrink the size to 8bit since its reserved anyways and move it to the beginning of the ram to not fragment data. I know you wont be happy about this since you need to update all Bootloaders and the USB Core.

I think the simplest option to fix this is the first. Someone should have thought about this first but well, it would be a mess to tell the customers now to update the bootloader and most just cant do it.
But what we should do for further boards is to use the first ram address 0x100 and a 8 bit value (option3). I will also use this for my 16u2/32u2 core that i will publish soon (also as pullup request maybe).

NicoHood:
Also the Bootloader itself should be fixed. Because it starts the sketch when the USB clock is still on. A better way would be a complete watchdog reset with different fuses then (like [HoodLoader2](https://github.com/NicoHood/HoodLoader2)).

Is anyone willing to accept/improve that? I would work on this, but not if nobody cares and the work is for nothing.

yyyc514:
Is there some way for a sketch to override the 0x800 location in CDC.cpp (part of core).  I can imagine various ways to fix this but all of them require changing that address in CDC to something else... is there any each way for a sketch (or library) to do this without having to create a whole custom Arduino core lib?

yyyc514:
[...]

PR #3304, opened 2015-06-08
[RFC] Pluggable USB core for AVR

facchinm:
This PR introduces the concept of Pluggable USB modules to help getting the best from our tiny ATmega32u4 without overhauling the core library with functions that could be useful for no more than 10% of the users.

Using this method we can _plug_ a library using the USB core by simply including its header in the sketch, and if not used it will not occupy additional resources.

It is of course a work in progress (thus tagged as RFC), still not ported to DUE core, and VERY probably bug-prone. Also, the memory footprint is not minimized.

_Useful infos_
* the CDC module is now non-removable (commit 43701fd)
* the Pluggable USB module inherits most of the code from old HID and replaces its functions with hooks
* max additional endpoints is arbitrarily set to 3
* HID, Mouse, Keyboard and MouseAndKeyboard libraries are "generated" from `master` code (so all code from #1803 can be merged in a second time)
* a MIDIUSB library based on PluggableUSB has been added (it is a very raw porting of #2943) only to demonstrate how to plug multiple modules
* ~~`setupUSB() ` weak function has been added to `main.c` , so it will execute before `setup()` if `USBCON` is defined.~~

_Known issues_
* I'll add them here as soon as they get discovered (no doubt this list will grow soon :smile: )

_How to try it_
I decided not to change any provided example . If you wish to test the core, download the @ArduinoBot generated packages and modify the examples in this way:

On the top of the file:

* Mouse only examples: add 
```C
#include "Mouse.h"
#include "HID.h"
``` 
* Keyboard only examples: add 
```C
#include "Keyboard.h"
#include "HID.h"
``` 
* Mouse and Keyboard examples: add 
```C
#include "MouseAndKeyboard.h"
#include "HID.h"
``` 
* HID and MIDIUSB example: 
```arduino
#include "MIDIUSB.h"
#include "HID.h"
#include "Mouse.h"
```


NicoHood:
This really sounds good to me. I havent looked at the code yet, but I will test this for sure.
I also made the CDC etc pluggable. Not via include, just because I moved the code in a seperate .cpp file which is not linked if not used. This saves a lot of space.

Also I would provide an include "USB.h" to enable USB stuff. Then users can remove the whole USB core and save even more space if not needed and also avoid problems with the PC.

You may want to have a look at my core and at the dev branches to get new Ideas possibly:
https://github.com/NicoHood/HID

facchinm:
Hi Nico,
thanks for taking the time to watch this PR!
I obviously took a look at your core before starting this effort and I think that most of your patches (and other peoples' patches as well) could be integrated if we are able to modularize the USB :smirk: 
The biggest problem in using your approach is that you need to:
* provide a lot of board variants (too noisy) or
* let the user modify the `#defines` in the core itself (too advanced)

Also, your core handles a lot of HID-based devices but it's not ready for other kind of devices (like MIDI #2943 or something else that I can't imagine at the moment :smile: )

However, if you could test the patchset it would be great, I'm only waiting for the feedback!

NicoHood:
Well I have a midi option that I could integrate. Its just under a dev state so I did not create a proper menu yet. Also it has a 2nd Audio device with errors under windows. I just used the arcore code for this.

I know that the boards file is not perfect, but it was the only way for me to simply add a menu to the IDE.

I really like your IDE to make the USB stuff more flexible. You know that this will change a lot of code. Is the Arduino Team willing to change it? Because so many people provided new features but mostly none of them were integrated yet. You are collaborator, does this mean that we can count on a PR merge if its good?

You should also consider to fix this bug when you update the USB core:
https://github.com/arduino/Arduino/issues/2474#issuecomment-111397112

And you should also consider to update the bootloader fuses and watchdog behaviour in order to fully deactivate the USB-Core. Right now the bootloader starts the sketch directly. You could better to a watchdog reset and start the code independent. The difference is that all settings like the USB clock is reset. Without this fix you have to add the USB clock ISR into the core as workaround. I could explain it better if someone is willing to update the bootloader in general. This would make life much easier.

I am not likely able to test your code this/next week, I am too busy, sorry. But I will have a look for sure.

matthijskooijman:
Overall, this looks like a nice implementation, using a linked list of modules and separate libraries is an elegant way to solve this.

> setupUSB() weak function has been added to main.c , so it will execute before setup() if USBCON is defined.

Isn't a weak setup function completely counter to the idea of pluggable modules? Only one module can implement this setup function, otherwise you'll get linker errors (or if the modules make their implementation weak too, they'll get silently ignored). Wouldn't an init function (since the setup name seems to be used already for processing control packets) in the callbacks make sense instead?


Would it make sense to write some documentation about how to use this now? Having documentation would help to understand and review the PR, and writing documentation *after* merging usually means it'll take months before that happens...

NicoHood:
What do you mean by cb variable?

Can we also see u2 Series support with this update? HL2 is getting more and more popular this could be a great time to merge everything together.

NicoHood:
I will add comments/thoughts to the PR here (this post will be edited):

I'd remove the u8 since it just makes things more complicated with the compatibility. Just use uint8_t if you ask me.

hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -445,15 +431,12 @@ int USB_RecvControl(void* d, int len)
...
-#endif
+	CDC_GetInterface(&interfaces);
 
-#ifdef HID_ENABLED
-	total += HID_GetInterface(&interfaces);
+#ifdef PLUGGABLE_USB_ENABLED
+	PUSB_GetInterface(&interfaces);
 #endif
 
 	return interfaces;
```
NicoHood: This returns always 0. Whats the sense of that function?

hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -317,19 +306,19 @@ int USB_Send(u8 ep, const void* d, int len)
...
 
-#ifdef HID_ENABLED
-	EP_TYPE_INTERRUPT_IN		// HID_ENDPOINT_INT
+#ifdef PLUGGABLE_USB_ENABLED
+	//allocate 3 endpoints and remove const so they can be changed by the user
+	0,
+	0,
+	0,
 #endif
 };
```
NicoHood: I dont know of there is a definition of the number of endpoints, but please consider that other MCUs dont have 6 endpoint and that not every endpoint can take double 64 bytes. If you want to make it flexible then let the use decide about the size and double bank or not.

hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -72,18 +65,14 @@ const u8 STRING_PRODUCT[] PROGMEM = USB_PRODUCT;
 const u8 STRING_MANUFACTURER[] PROGMEM = USB_MANUFACTURER;
 
 
-#ifdef CDC_ENABLED
```
NicoHood: I'd also not force the CDC to be automatically added. This is not pluggable if you ask me. There should be a way to remove it and use other functions for usb. And you should add weak events for the CDC Serial as in my HID Project.

hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -17,17 +17,10 @@
...
-#define EP_TYPE_BULK_IN				0x81
-#define EP_TYPE_BULK_OUT			0x80
-#define EP_TYPE_INTERRUPT_IN		0xC1
-#define EP_TYPE_INTERRUPT_OUT		0xC0
-#define EP_TYPE_ISOCHRONOUS_IN		0x41
-#define EP_TYPE_ISOCHRONOUS_OUT		0x40
-
 /** Pulse generation counters to keep track of the number of milliseconds remaining for each pulse type */
 #define TX_RX_LED_PULSE_MS 100
 volatile u8 TxLEDPulse; /**< Milliseconds remaining for data Tx LED pulse */
```
NicoHood: I'd remove the u8 since it just makes things more complicated with the compatibility. Just use uint8_t if you ask me.

hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -72,18 +65,14 @@ const u8 STRING_PRODUCT[] PROGMEM = USB_PRODUCT;
...
-#endif
 
 //	DEVICE DESCRIPTOR
 const DeviceDescriptor USB_DeviceDescriptor =
 	D_DEVICE(0x00,0x00,0x00,64,USB_VID,USB_PID,0x100,IMANUFACTURER,IPRODUCT,0,1);
 
-const DeviceDescriptor USB_DeviceDescriptorA =
-	D_DEVICE(DEVICE_CLASS,0x00,0x00,64,USB_VID,USB_PID,0x100,IMANUFACTURER,IPRODUCT,0,1);
+const DeviceDescriptor USB_DeviceDescriptorB =
+	D_DEVICE(0xEF,0x02,0x01,64,USB_VID,USB_PID,0x100,IMANUFACTURER,IPRODUCT,0,1);
```
NicoHood: I might get this wrong, but you have to add 0xEF, 0x02 and 0x01 to the upper Descriptor. Thats how I did it in my source at least.

Also I'd go for definitions of those values so they are better understandable.

hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
```diff
@@ -0,0 +1,109 @@
...
+static PUSBListNode* rootNode = NULL;
+static PUSBListNode* lastNode = NULL;
+
+int8_t PUSB_GetInterface(u8* interfaceNum)
+{
+	int8_t ret = 0;
+	PUSBListNode* node = rootNode;
+	for (u8 i=0; i<modules_count; i++) {
+		ret = node->cb.getInterface(interfaceNum);
+		node = node->next;
```
NicoHood: Posted the comment in a commit, but still applies here. I am not sure if all those functions make so much sense with the retvalue. Also lastNode is not really used.

hardware/arduino/avr/cores/arduino/USBCore.cpp
matthijskooijman: Isn't there something defined in io.h that indicates how many endpoints are supported? Not sure what you mean with the double-thing (I don't really know the USB details).

hardware/arduino/avr/cores/arduino/USBCore.cpp
matthijskooijman: +1, but converting types is completely unrelated to this change and should probably be done separately to not complicate this PR.

matthijskooijman:
> What do you mean by cb variable?

I presume this was in reply to [this comment](https://github.com/facchinm/Arduino/commit/790cf7d7a0f6a6756375c3a7287d209676162a0f#commitcomment-11684765)

I mean that instead of:

	cb.setup = &HID_Setup;
	cb.getInterface = &HID_GetInterface;
	cb.getDescriptor = &HID_GetDescriptor;
	cb.numEndpoints = 1;
	cb.endpointType[0] = EP_TYPE_INTERRUPT_IN;

you could do:

	static const PUSBCallbacks cb = {
		.setup = &HID_Setup,
		.getInterface = &HID_GetInterface,
		etc.
	}

hardware/arduino/avr/cores/arduino/USBCore.cpp
NicoHood: USB Banks can have 8,16,32,64 bytes of data. Double or single sized. Double means it has 2x64 in this case. But the u2 Series doesnt support that if you do this on any endpoint. It has less ram for the USB stuff (has a special name in the datasheet). Also to avoid problems with custom endpoint descriptors I'd let the programmer choose if its double banked or not and if it 8,16,32,64 bytes.

hardware/arduino/avr/cores/arduino/USBCore.cpp
facchinm: it does't return 0 because `interfaces` is incremented by the callee (instead, the original `total` variable was completely useles)

hardware/arduino/avr/cores/arduino/USBCore.cpp
facchinm: Yes, all the stuff is very 32u4-centric, so your help in the u2 side and your deep knowledge of the AVR USB stack are very precious!

hardware/arduino/avr/cores/arduino/USBCore.cpp
facchinm: CDC removal was a choice (indeed the commit is atomic and can be safely reverted)

facchinm:
@NicoHood @matthijskooijman Thank you so much for taking time to analyse the (huge) PR. I'll try to update it following your advices as soon as I can get some spare time.

About OS-specific issues, I think that Nico is our man, so any help is really appreciated!

facchinm:
@matthijskooijman  about the weak function, I could not imagine anything better. 

From my tests the (also weak) `setupUSB` functions contained in the library do overwrite the one in `main.c` (I think it's only a linking order side effect but it works). 

However, anyone can freely overwrite it with its own non-weak implementation (like in the MIDI+HID example). Any alternative however is very welcome :smile:

matthijskooijman:
> However, anyone can freely overwrite it with its own non-weak implementation (like in the MIDI+HID example). Any alternative however is very welcome
The problem is that when you override the `setupUSB` function, all previous overrides will be discarded. This means that a sketch that includes multiple USB libraries, must provide its own `setupUSB` function that calls all the appropriate `begin` methods.

> @matthijskooijman about the weak function, I could not imagine anything better. 

Looking more closely, I see you are using `setupUSB` to call e.g. `HID.begin()`, which eventually calls `PUSB_AddFunction()` to register the usb module and its callbacks. My previous suggestion, to use a callback instead of a weak `setupUSB` function of course doesn't work, since it is needed to register the callbacks in the first place.

There is a  good alternative, though: Use a global object, whose constructor calls `PUSB_AddFunction()`. Constructors of global objects are called on startup automatically.

As an example of how this works, see ModuleHandler and Module in https://github.com/Pinoccio/library-pinoccio/tree/master/src/modules Those register a Module* in a linked list inside ModuleHandler, which is the equivalent of calling `PUSB_AddFunction`

One caveat is that this all runs before `main()` is even started and the initialization order is undefined, so you should check that `PUSB_AddFunction` does not need any constructors or other initalization code to run (global variables initialized to 0 or some other constant are ok, zeroing of the .bss section and loading of the .data section happens before calling global constructor). A quick look around the code shows that this is already ok.

I even think that this approach doesn't require separate libraries for every module, since the compiler can  completely remove a module, including its global constructor, if it's not referenced anywhere at all. I'm not 100% sure about this, though. And in most cases, using separate libraries is sane approach regardless of this.



I had a closer look at the HID libraries, and I'm not so happy with how those look. AFAICS, the Mouse and Keyboard libraries both define a few specially-named symbols, which are then used by the HID library. This of course doesn't allow more than one HID "submodule", which is why there is a KeyboardAndMouse library, which is of course fairly ugly.

Ideally, an approach with callbacks similar to PUSB_AddFunction is also used for HID submodules, allowing them to be completely separate from each other. However, I'm not sure if the HID descriptors are actually easy to compose (e.g. can you just concatenate or otherwise combine the Mouse and Keyboard descriptors into something valid)?

facchinm:
* You are right about the ugliness of HID submodules (their need of special symbols especially) but the descriptor must be unique so only one `#include` is allowed "by design" (also enforced by 18a2f2a)
Of course a fully modular solution is the way to go but from a library creator's point of view it would probably be more complicated than defining a specially named function (I'm thinking about descriptors overriding)
* Thanks a lot for the global constructor tip, I missed that completely (I'm a C guy :smile: )

Now the produced binary should be comparable (in size) with the non-modular one so we can start thinking about integrating the various PR as libraries and test the results

facchinm:
@ArduinoBot build this please

facchinm:
Size comparison with the current code:

Sketch | Modular | Old
------------ | -------------|-------------
KeyboardLogout | 5574-218 |5030-153
KeyboardSerial | 4868-216	| 4324-151
KeyboardAndMouseControl |	5704-217	 | 5102-151
ButtonMouseControl | 4962-195 | 5110-155
empty sketch |	3476-133	 | 4254-151
ASCIITable | 4364-211 | 5144-231

facchinm:
https://github.com/facchinm/Arduino/tree/testPlugUSB_PR+1803 contains a complete porting of PR #1803 adapted to the Pluggable framework

facchinm:
Almost end of week update:
* HID is now pluggable itself! You can add
```
#include "Mouse.h"
#include "Keyboard.h"
#include "HID.h"
```
and the descriptor will be the combination of the submodules' descriptors. The method is not very memory efficient; however you can still write a library which plugs on PluggableUSB directly, embedding in it an HID implementation and only the functions you need.

* the PR now contains @matthijskooijman C++11 patches to get rid of a nasty warning

* next step: enable `lto` (reduces flash occupation by 400 bytes) :smile: https://gist.github.com/facchinm/432fbb8ca4d9705644bd

@NicoHood @matthijskooijman @obra @follower @cmaglie @madrang @damellis @PaulStoffregen @weizenspreu and all people interested in USB on AVR, please test the latest @ArduinoBot image to make sure that nothing breaks spectacularly before merging :grin:

Thank you very much!

obra:
Oh wow. That's amazing. Thank you so, so, so much. I'm on the road this week but will test when I can.

NicoHood:
@yyyc514 of course it is interrupt driven. Not sure if your idea then still makes sense.

@ facchinm Also see this:
https://github.com/NicoHood/HID/issues/42

Going to test this sooner or later... Sounds good at least!

NicoHood:
huh? that was quick. I didnt even had a chance to test this yet.
Did you got feedback from somewhere else? I am a little surprised that nothing happened related to USB for so long, and now this fix is merged so fast.

Not that I feel bad about it, I should really go ahead and try this now!

cmaglie:
I think that this is one of the most-desired patch ever (just look the number of PR/Issues), so we decided to merge it upstream to accelerate tests and adoption. 

On our side we tested it on everything, BTW this patch is not yet released, so we are still in time to fix bugs or (in the worst case but I hope not :)) completely revert it.

NicoHood:
Good idea, I think its a good idea to improve the usb stack.

weizenspreu:
What do you mean by "quick"? This topic took now 1.5 years to come any further.

NicoHood:
Laugh out loud :D 

Yes I know. And this PR started about a month ago and is now merged. Nevermind, its GOOD that there IS progress.

facchinm:
Initial howto on writing a PluggableUSB or PluggableHID based library

https://github.com/arduino/Arduino/wiki/PluggableUSB-and-PluggableHID-howto

NicoHood:
May you eliminate the u8 which only causes problems?

NicoHood:
Any chance to see a pluggable keyboard layout?
https://github.com/NicoHood/HID/pull/33

facchinm:
Well Nico, a library providing multiple layouts is now possible, you can write it by yourself and we'll be glad to include it in the library manager :wink:

NicoHood:
Looking at the code now, adding notes here via edit:

* Overall it looks very nice to me and a simple example just works fine.
* Found a typo in the instruction: `Then you can then perform USB writes calling, for example, `
* begin() should return a bool and if possible it should wait till the HID is initialized. Maybe this is impossible, but I noticed problems like this: https://github.com/NicoHood/HID/issues/29 What about while(!HID); like the Serial?
u8 should be replaced with uint8_t if you ask me.

facchinm:
Adding the `bool()` operator is a good idea! 
About `u8`, they will be replaced in the libs when PR #3542 will be merged.
Typo fixed, thanks :wink: 

Endpoint flexible size is a bit of a mess, but if you come up with a solution I'll be very happy to review it. Also, feel free to open PRs for Keyboard and Mouse libraries modifications.

NicoHood:
I am currently working on a PR to backport all my HID Project fixes. Already done with the major core fixes, HID libraries will follow.

The flexible EP_SIZE needs a bit more core changes, however I applied my simple fix for now in the PR (soon).

What is more important is this:
https://github.com/NicoHood/HID/compare/daa80ec7bfe395aa52d1eb9e16169b95341f8873...dev#diff-5478f073c4245fb29e2a0975e99c647eR101

At the moment you set those values wrong. USB cannot be pluggable in every situation. So I dont know how to solve this the best way, but we should at least try to add a correct CDC descriptor.

Let me give my fixes a last check and I'll open a new PR for this.

NicoHood:
* An HID report backwards would be nice. This way you can read the keyboard led states. My solution was fixed for the keyboard leds, but one could adapt it to the pluggable HID maybe. https://github.com/NicoHood/HID/compare/daa80ec7bfe395aa52d1eb9e16169b95341f8873...dev#diff-fac0f634437717ff0b4218df8034dba3R157
* Would you be willing to merge a Gamepad, Absolute Mouse and Consumer Keys API? Or should I provide this in my Github repo via library manager? I'd change a few things in the standard APIs though.
* You should add a not how to renew the drivers, as HID will conflict otherwise under windows if you change the descriptor. https://github.com/NicoHood/HID/wiki/Troubleshoot-FAQ#switching-the-hid-core

NicoHood:
There we go:
https://github.com/arduino/Arduino/pull/3562

NicoHood:
@facchinm Hey, you did a lot of great work. Would you be so kind to have a look at the PR to apply my small fixes?

If you disagree with the CDC Events and control line stuff I can remove it. There is another PR my mathijs I think who did something different, maybe better than that. But the rest of my PR is not much to look at but would make the core more flexible and fixes a few problems.

Issue #3310, opened 2015-06-09
pinMode OUTPUT sets HIGH after digitalWrite set LOW on Arduino Due


3esmit:
Architecture: SAM
Board: Arduino Due
When the pin is defined as OUTPUT it automatically sets HIGH in that pin, even when digitalWrite previously set it LOW. This is different behavior from AVR. 

I need all DO set low at startup, so in Arduino/hardware/arduino/sam/cores/arduino/wiring_digital.c. To make pinMode output LOW by default, change PIO_OUTPUT_1 to PIO_OUTPUT_0.

See http://forum.arduino.cc/index.php?topic=185230.0 and http://forum.arduino.cc/index.php?topic=154869.0

DavidEGrayson:
Setting the pin to low by default as pull request #3317 does might be safer than setting it high, but it still doesn't make the behavior of the Arduino Due compatible with other Arduino boards so it's not that great.

On the Arduino Uno and Leonardo, if you want to set an output to be high without having it temporarily drive low (which could break things), then you have to do it with this sequence of commands:

~~~~
digitalWrite(x, HIGH);
pinMode(x, OUTPUT);
~~~~

And if you want to drive it low without temporarily driving it high, you have to use this sequence:

~~~~
digitalWrite(x, LOW);
pinMode(x, OUTPUT);
~~~~

Currently, the first sequence would work on the Arduino Due and the second sequence would not work.  Pull request #3317 simply changes it so that the first sequence doesn't work and the second sequence does work.

In my opinion, for any board to be called "Arduino compatible" *both* of those sequences should work.  If one of them doesn't work, then you will have trouble running code and libraries that depend on the traditional behavior.

There is at least one library where the code had to be changed to work around this non-standard behavior of the Arduino Due: https://github.com/pololu/a4990-motor-shield/blob/master/A4990MotorShield/A4990MotorShield.cpp#L12-24

I haven't looked into the details, but it might be that for the Arduino DUE, the Arduino IDE simply needs to emulate the PORTx and DDRx bits of the AVR.

Even if you don't care about compatibility with existing Arduino code and other Arduino boards, it seems like it is an important feature to be able to set an I/O pin to be low without having it temporarily drive high, and vice versa.  Right now it looks like the Due just doesn't support that.

3esmit:
I see that my issue is wrongly described, sorry about that.
[...]

PR #3317, opened 2015-06-11
Defaults output pins to LOW on Due

tamul:
When a pin is designated as an output on the Arduino Due, the pin is set
to a HIGH logic level. Changing the default pin state to LOW makes the
behaviour correspond with AVR. See #3310

### DIFF
--- a/hardware/arduino/sam/cores/arduino/wiring_digital.c
+++ b/hardware/arduino/sam/cores/arduino/wiring_digital.c
@@ -54,7 +54,7 @@ extern void pinMode( uint32_t ulPin, uint32_t ulMode )
         case OUTPUT:
             PIO_Configure(
             	g_APinDescription[ulPin].pPort,
-            	PIO_OUTPUT_1,
+            	PIO_OUTPUT_0,
             	g_APinDescription[ulPin].ulPin,
             	g_APinDescription[ulPin].ulPinConfiguration ) ;
 


DavidEGrayson:
Setting the pin to low by default as this pull request does might be safer than setting it high, but it still doesn't make the behavior of the Arduino Due compatible with other Arduino boards so it's not that great.

On the Arduino Uno and Leonardo, if you want to set an output to be high without having it temporarily drive low (which could break things), then you have to do it with this sequence of commands:

~~~~
digitalWrite(x, HIGH);
pinMode(x, OUTPUT);
~~~~

And if you want to drive it low without temporarily driving it high, you have to use this sequence:

~~~~
digitalWrite(x, LOW);
pinMode(x, OUTPUT);
~~~~

Currently, the first sequence would work on the Arduino Due and the second sequence would not work.  This pull request simply changes it so that the first sequence doesn't work and the second sequence does work.

In my opinion, for any board to be called "Arduino compatible" *both* of those sequences should work.  If one of them doesn't work, then you will have trouble running code and libraries that depend on the traditional behavior.

There is at least one library where the code had to be changed to work around this non-standard behavior of the Arduino Due: https://github.com/pololu/a4990-motor-shield/blob/master/A4990MotorShield/A4990MotorShield.cpp#L12-24

I haven't looked into the details, but it might be that for the Arduino DUE, the Arduino IDE simply needs to emulate the PORTx and DDRx bits of the AVR.

Even if you don't care about compatibility with existing Arduino code and other Arduino boards, it seems like it is an important feature to be able to set an I/O pin to be low without having it temporarily drive high, and vice versa.  Right now it looks like the Due just doesn't support that.

facchinm:
Replaced by #3524

PR #3324, opened 2015-06-12
Fix build on windows (rename/move).

### LINKED ISSUE #3322: build error - 1.6.5
During build time, the following error message is displayed and build failed. 'Apache Ant(TM) version 1.9.4 compiled on April 29 2014' is used on Cygwin.

error message
============
```
untar-unzip-checksum:
     [echo] Testing checksum of "Firmata-2.4.3.zip"

unzip:
   [rename] DEPRECATED - The rename task is deprecated.  Use move instead.

BUILD FAILED
```

Chris--A:
It appears ant is no longer supporting `rename` in windows.
This patch replaces these commands with `move`.

This fixes #3322

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -187,14 +187,14 @@
       <param name="final_folder" value="${target.path}/libraries/Firmata-2.4.3" />
       <param name="dest_folder" value="${target.path}/libraries" />
     </antcall>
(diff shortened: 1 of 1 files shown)


ffissore:
:+1:

Issue #2585, opened 2015-01-28
Starter Kit Project 3 Love-o-meter GLITCH


rcpseattle:
In the documentation and the sample code for the Love-o-meter project (Project 3 in starter kit), there appears to be a glitch (copy of relevant code below):

Issue is best explained with example: 
Assume, baseline Temperature is set to 20 degrees C, then:

If you GET WARMER gradually then 
T<20 - no lights
T=21 - NO LIGHTS (keeps doing what it was doing, i.e. no lights)
T=22 - 1 LED
T=23 - 1 LED and so on...

If you GET COLDER gradually then 
T=23 - 1 LED
T=22 - 1 LED
T=21 - 1 LED (keeps doing what it was doing, i.e. 1 LED)
T=20 - 1 LED (keeps doing what it was doing, i.e. 1 LED)
T < 20 - no light

This means that 
during heating the temp interval 20<=T<22 has no lights, 
during cooling the temp interval 20<=T<22 has 1 LED (or better, it keeps whatever the previous state was).

Not too big a deal but I suspect this was unintentional.

Code:
 // if the current temperature is lower than the baseline
  // turn off all LEDs
  if(temperature < baselineTemp){
    digitalWrite(2, LOW);
    digitalWrite(3, LOW);
    digitalWrite(4, LOW);
  } // if the temperature rises 2-4 degrees, turn an LED on 
  else if(temperature >= baselineTemp+2 && temperature < baselineTemp+4){
    digitalWrite(2, HIGH);
    digitalWrite(3, LOW);
    digitalWrite(4, LOW);
  } // if the temperature rises 4-6 degrees, turn a second LED on  
  else if(temperature >= baselineTemp+4 && temperature < baselineTemp+6){
    digitalWrite(2, HIGH);
    digitalWrite(3, HIGH);
    digitalWrite(4, LOW);
  } // if the temperature rises more than 6 degrees, turn all LEDs on
  else if(temperature >= baselineTemp+6){
    digitalWrite(2, HIGH);
    digitalWrite(3, HIGH);
    digitalWrite(4, HIGH);

agdl:
@rcpseattle yes you are right, in order to have the same behaviour the first if must be if if(temperature < baselineTemp + 2) i corrected it and pushed on Git. Thank you!

PR #3326, opened 2015-06-12
Fixed behaviour in the falling phase PJ03 StarterKit

agdl:
According to #2585 there was a "glitch" in the falling phase because of a missing +2 in the first condition


matthijskooijman:
In the future, can we please try to have a bit more meaningful commit messages? They should be self-contained and make sense without having to look at the diff and file list, as far as I'm concerned. In this case, an indication that it applies to an example for this starter kit would be great, as well as some guidance as to what this "falling phase" is...

PR #3344, opened 2015-06-16
Cleaning up Java 8 illegal identifier warnings

jverne:
Bulk rename of "_" identifier used across the IDE to quiet Java 8 warnings (and eventual Java 9 errors.)

The classpath change was to get head to build in Eclipse.


ffissore:
These one needs some more love because the `_` identifier is used by some python scripts to extract strings that need to be translated. Those scripts too need to use the new `getString` method

ffissore:
Those scripts are stored at https://github.com/arduino/Arduino/tree/master/arduino-core/src/processing/app/i18n/python

cmaglie:
I'd also like to use a more terse name for the method to not clutter too much the code, something like `tr(...)` or so.

jverne:
I'll look into the python scripts tonight, as well as renaming the identifier. I wasn't happy with the generic name of the static myself. Look for another pull request later today, Eastern Time.

jverne:
I'm going to redo this one.

ffissore:
@jverne any news?

jverne:
I got the simpler rename of the identifier changes done locally, but stalled trying to understand the Python parts, which are rather opaque (to me anyway.)

As an aside, I think one or more of the source files are encoded oddly because Eclipse cannot refactor cleanly without erroring out. So the rename of _ to something else had to be a creative direct edit everywhere.

I'll take a stab at figuring out (and smoke-testing, if that is possible) the Python dependencies this weekend. If anyone has hints about that, this might help. Then I can pull form the master, merge, and redo the pull request.

ffissore:
@jverne then please rebase your PR with the current master so we can merge it. We'll think about the python scripts

ffissore:
Fixed with 50cacc1756748e2a9ba24692ad397476380af5b1. Thank you

Issue #1005, opened 2012-11-15
If you create 27 sketches in one day you are advised to "go for a walk" [imported]


cmaglie:
This is [Issue 1005](http://code.google.com/p/arduino/issues/detail?id=1005) moved from a Google Code project.
Added by 2012-08-12T07:02:20.000Z by [n...@gammon.com.au](http://code.google.com/u/111479721553138312180/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium

### Original description

**What steps will reproduce the problem?**
1. Open the IDE
2. Hit Ctrl+N (or Command+N) 27 times to create 27 sketches

**What is the expected output? What do you see instead?**

I expect to be able to create 27 sketches using the IDE.

Instead I am advised by a dialog box:

"You've reached the limit of auto naming of new sketches for the day. How about going for a walk instead?"

If I attempt again I get a different message:

"No really, time for some fresh air for you."

Seriously?

**What version of the Arduino software are you using? On what operating**
**system?  Which Arduino board are you using?**

Version 1.0.1
Mac OS/X
Board not relevant

**Please provide any additional information below.**

See this thread:

http://arduino.cc/forum/index.php?topic=118187

Screenshots of the dialog boxes there.

Humorous though the messages may be, you should be able to create 27 sketches in one day. Think of a school lab.

Chris--A:
This issue has been resolved in #3366 
The maximum is now well over 600 so hopefully this problem should not be an issue.

PR #3366, opened 2015-06-19
Increase maximum number of spawned instances.

Chris--A:
This commit increases the number of sketches able to be spawned from within the IDE. It now allows 676 sketches rather than 26 ('a' to 'zz' rather than 'a' to  'z').﻿

Resolves: #1005


ffissore:
Thanks!

Chris--A:
Nice :+1:

PR #3384, opened 2015-06-21
Avoid rounding errors in baudrate calculation.

isering:
The current baudrate calculation formula results in non optimum baud rate register values, because the formula always rounds UBRR values down:

```
UBRR0L = (uint8_t)(F_CPU/(BAUD_RATE*16L)-1);
UBRR0H = (F_CPU/(BAUD_RATE*16L)-1) >> 8;
```

Correct rounding is achieved using the following formula:

```
UBRR0L = (uint8_t)((F_CPU+BAUD_RATE*8L)/(BAUD_RATE*16L)-1);
UBRR0H = ((F_CPU+BAUD_RATE*8L)/(BAUD_RATE*16L)-1) >> 8;
```

For example, using `F_CPU=20000000UL` and `BAUD_RATE=115200UL`, the first formula results in `UBRR=9`, while the second formula results in `UBRR=10`, where the correct value would be `UBRR=9.85`.

See also [mikrocontroller.net](https://www.mikrocontroller.net/articles/AVR-Tutorial:_UART#UART_konfigurieren) (German).

### DIFF
--- a/hardware/arduino/avr/bootloaders/atmega/ATmegaBOOT_168.c
+++ b/hardware/arduino/avr/bootloaders/atmega/ATmegaBOOT_168.c
@@ -351,33 +351,33 @@ int main(void)
 	/* initialize UART(s) depending on CPU defined */
 #if defined(__AVR_ATmega128__) || defined(__AVR_ATmega1280__)
 	if(bootuart == 1) {
-		UBRR0L = (uint8_t)(F_CPU/(BAUD_RATE*16L)-1);
-		UBRR0H = (F_CPU/(BAUD_RATE*16L)-1) >> 8;
+		UBRR0L = (uint8_t)((F_CPU+BAUD_RATE*8L)/(BAUD_RATE*16L)-1);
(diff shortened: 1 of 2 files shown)


matthijskooijman:
I wonder if it would make sense to use the same [formula used by HardwareSerial.cpp](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp#L106-L118):

    baud_setting = (F_CPU / 8 / baud - 1) / 2;

or, when running with U2X0:

    (F_CPU / 4 / baud - 1) / 2

I believe this also results in optimal rounding (and I think it essentially boils down to the same as the formula proposed, though I didn't look close enough to be sure).

Using identical formulas should maximize compatibility in any case.

jiser:
You are right, for valid input arguments both formulas yield the same results.

facchinm:
I'm closing this PR due to the bootloader/firmware policy (they are being updated only if they go get burned in production boards)

PR #3386, opened 2015-06-21
Set DTR and RTS in serial monitor

scottwday:
Sets DTS and RTS pins in serial monitor.

Useful in projects where these lines control reset and chip-select pins.

Set serial.monitorDtr and serial.monitorRts in preferences.txt to configure the pins

### DIFF
--- a/app/src/processing/app/SerialMonitor.java
+++ b/app/src/processing/app/SerialMonitor.java
@@ -96,6 +96,19 @@ public class SerialMonitor extends AbstractMonitor {
         addToUpdateBuffer(buff, n);
       }
     };
+	
+	int dtrState = PreferencesData.getInteger("serial.monitorDtr", -1);
+	if (dtrState == 0)
+		serial.setDTR(false);
+	if (dtrState == 1)
+		serial.setDTR(true);
+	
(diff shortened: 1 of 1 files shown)


NicoHood:
Good idea, why not?

What happens if this setting is not represented at all?

I'd also vote for a manual option in the monitor itself (like the baud rate), not just via a preferences.txt file.

scottwday:
If the setting isn't present in preferences.txt (i.e. for most people) then DTS behaves as per normal (stays low) and RTS isn't affected at all.

Putting in controls in the GUI would be nice, but not sure if it's worth the effort. I put this in for ESP8266 to hold the reset and boot mode in the right state, but it's not something I need to change very often.

If there was a GUI option then I'd like it to be hidden by default because arduino is supposed to be minimalistic.

Chris--A:
@scottwday 
I agree, an interface may be better, but at least having the option is great.
I'll have to give this a test, I currently use two buttons to reset/flash my ESP modules, an automated feature would be very welcomed.

cranphin:
Any progress on this at all, where's it stuck at?

cmaglie:
Fixed by #4102 that is similar but allows the configuration per-board.

Issue #3351, opened 2015-06-16
[Ubuntu?] Serial baud 250000 not opened properly


NicoHood:
As documentated and discussed here:
https://github.com/urjaman/fast-usbserial/issues/1

If you open the Serial at baud 250000 it will open the Serial at the last selected baud instead. And thatswhy it doesnt work. This was tested on Ubuntu 14.04 x64 with IDE 1.6.5 release. It might be an OS issue due to a non standard baud. Or it is an IDE problem. Otherwise 250k could work. The program minicom can open it properly under Ubuntu. This was not tested on Windows. See discussion there for more information.

BrentWilkins:
I didn't throw an oscope on the serial port, but if I select 250K baud in Serial.begin() & in the serial monitor, data passes just fine on the current version of OSX.

NicoHood:
No, we are talking about the USB CDC Serial, not the Hardware Serial. Please read the Problem fully.
The OS passes wrong parameters to the USB device so the USB device sets the USART wrong. And the wrong baud is the last selected baud. This bug happens with 250k, 500k, 1m and 2m. Possibly with any other non standard baudrate.

matthijskooijman:
WIth USB CDC serial on the Leonardo, the selected baud rate shouldn't matter at all, it is not used. So I presume you mean CDC serial on the 16u2 on e.g. a Uno, which connects to the HardwareSerial on the 328 side. Assuming that @BrentWilkins was using a Uno or similar, his test confirmed that the entire stack supports 250k, from Serial monitor to HardwareSerial.

IIRC reliably selecting non-standard baudrates is a bit tricky on Linux, so it's likely that his works on OSX, but breaks on Linux.

I quickly tested this sketch:

    void setup() {
      Serial.begin(250000);
      Serial.println("Hello, world!");
    }

    void loop() {
    }

Opening the serial monitor at 250000 shows a bit of garbage, suggesting that indeed the serial port wasn't opened at 250kbaud at the PC side. This is probaly something that needs to be fixed in the Java serial library used (Can't remember if that's rxtx or jssc).

matthijskooijman:
As for selecting non-standard baudrates, see this post:

http://marc.info/?l=linux-serial&m=143049135615695&w=2

(conclusion is that it sucks to set non-standard baudrates, but I think 250000 might be "standard enough" to work with the right prodding)

NicoHood:
[...]

PR #3389, opened 2015-06-22
Added 74880 baud rate to the serial monitor.

Chris--A:
Now that the latest IDE incorporates a board/core manager, the
ESP8266 community is now using this software. The baud rate of 74880
is important for debugging as it is what the bootloader uses.﻿

### DIFF
--- a/app/src/processing/app/AbstractMonitor.java
+++ b/app/src/processing/app/AbstractMonitor.java
@@ -140,7 +140,7 @@ public abstract class AbstractMonitor extends JFrame implements ActionListener {
     }
     lineEndings.setMaximumSize(lineEndings.getMinimumSize());
 
-    String[] serialRateStrings = {"300", "1200", "2400", "4800", "9600", "19200", "38400", "57600", "115200", "230400", "250000"};
+    String[] serialRateStrings = {"300", "1200", "2400", "4800", "9600", "19200", "38400", "57600", "74880", "115200", "230400", "250000"};
 
     serialRates = new JComboBox();
     for (String rate : serialRateStrings) {


matthijskooijman:
That's not a standard baudrate, I think? What protocol does the FTDI chip use? CDC ACM like the Uno r3, or is there an FTDI chip on there?

 I ask because on Linux, non-standard baudrates are tricky (to the point of not working at all using CDC ACM, see https://github.com/arduino/Arduino/issues/3351#issuecomment-113873045https://github.com/arduino/Arduino/issues/3351#issuecomment-113873045).

Chris--A:
The board I use to communicate with the ESP8266 (ESP-12) is a basic FT232RL module. I'm guessing CDC however its whatever the windows built in driver uses. I'll have a look later on and see if I can find any information.

However this baud rate is the default used on all ESP8266 chips.

matthijskooijman:
No, the FTDI modules use a different protocol than CDC devices. Both are supported on Windows and Linux, but the Linux FTDI driver supports the API that the IDE uses (through the JSSC library) to set custom baudrates.

Testato:
the ESP8266 is a chip with hardware serial at 3,3V level. It is a your choose which serial-usb device use on the project.
NodeMCU board ver0.9 use the CHxxx series, NodeMCU board ver 1.0 usethe CPxxx series. But this is not important. 
The 74880 baudrate is directly related to the quartz used on ESP8266, if you use a 40MHz clock the standard baud is 115200, but on all ESP8266 that we have on ebay and other manufacter is 26MHz and automatically the baudrate become 74880
So i think it is good add it to SerialMonitor speed selection

cmaglie:
Looks like this baud rate is really wanted. So I'm going to merge this even if this is going to work only with FTDI chip.

PaulStoffregen:
This may have unintended consequences, especially on Linux.

Long ago, the serial monitor had 14400 and 28800 in the baud rate menu.  With /dev/ttyACM0, setting either of those caused a bug in the driver to lock up the device.  Thereafter, the serial monitor couldn't properly open the port, which made it impossible to change to a working baud rate.

Hopefully we won't see that on modern Linux and JSSC.

cmaglie:
@PaulStoffregen
I've tried to set the 74880 back and forth, my ttyACM0 is not able to set it (it keep the previously set speed) but there are no other side effect, I can change the speed back to another value.

BTW your comment makes me thing that we can actually get an error value from JSSC that tell us if the baudrate change has been successful or not. I'll check this out.

cmaglie:
Here we go: #3420 

It seems to be working, at least for me, when I try to set 74880 baud the error is printed on the console.

It would be nice if someone can confirm that Linux+ESP (or any other FTDI based converted) instead works.

PR #3404, opened 2015-06-23
Windows: new Arduino15 location and automatic migration

### LINKED ISSUE #2902: 1.6.2+ Use of Windows Appdata folder appears to breach Microsoft usage guidelines
See https://msdn.microsoft.com/en-us/library/windows/apps/hh465094.aspx

The Windows AppData folder is not intended to be used to store binaries or other large files

(From the page linked to)

"Use roaming for user preferences and customizations, links, and small data files. For example, use roaming to preserve a user's background color preference across all devices."

"Don't roam large sets of app data. There's a limit to the amount of app data an app may roam; use RoamingStorageQuota property to get this maximum. If an app hits this limit, no data can roam until the size of the app data store no longer exceeds the limit. When you design your app, consider how to put a bound on larger data so as to not exceed the limit. For example, if saving a game state requires 10KB each, the app might only allow the user store up to 10 games."

Also. The folder will sync via Microsoft's cloud syncing service for some users, which may be a bandwidth issue to some users.

ffissore:
This PR aims at solving the long debate that took place on the developers mailing list about the location of cores/tools installed through Boards Manager. On windows it **was** `Roaming\Arduino15`, which breaks Microsoft rules and causes troubles to who actually like own profile to roam between different computers

New location is `%LOCALAPPDATA%\Arduino15`

When no custom preferences file location is specified (via CLI) and if the IDE finds something in `%APPDATA%\Arduino15` and no `%LOCALAPPDATA%\Arduino15` is present, IDE prefs files will be moved. Nothing will happen otherwise.

Fixes #2902

/cc @rogerclarkmelbourne

### DIFF
--- a/arduino-core/src/processing/app/BaseNoGui.java
+++ b/arduino-core/src/processing/app/BaseNoGui.java
@@ -962,7 +962,7 @@ public class BaseNoGui {
     }
   }
 
-  static public void initParameters(String args[]) {
+  static public void initParameters(String args[]) throws IOException {
     String preferencesFile = null;
 
     // Do a first pass over the commandline arguments, the rest of them
--- a/arduino-core/src/processing/app/Platform.java
+++ b/arduino-core/src/processing/app/Platform.java
@@ -245,4 +245,8 @@ public class Platform {
     Process process = Runtime.getRuntime().exec(new String[]{"chmod", Integer.toOctalString(mode), file.getAbsolutePath()}, null, null);
     process.waitFor();
   }
+
+  public void fixSettingsLocation() throws IOException {
+    //noop
+  }
 }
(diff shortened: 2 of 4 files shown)


rogerclarkmelbourne:
Federico

 

Thanks.

 

I will download the Nightly build

 

Cheers


Roger

rogerclarkmelbourne:
I tried the nightly build

arduino-nightly

http://downloads.arduino.cc/arduino-nightly-windows.zip

But it still seems to create the folder in %appdata% i.e Appdata roaming.

Should the fix be in the nightly build ?

Chris--A:
It will not be available in the nightly build until its merged. You can clone ffissore:windows-appdata and build it or download one of the archives listed above in @ArduinoBot's message.

rogerclarkmelbourne:
Thanks Chris

The location is not my only  worry, so I will wait until its in a nightly build

Issue #3413, opened 2015-06-24
Windows build fails, because it looks for JDK release file in the wrong place


martinjos:
Trying to build the project on Windows results in a build failure because of this line:

    <loadproperties srcfile="${WINDOWS_BUNDLED_JVM}/release" prefix="windows"/>

WINDOWS_BUNDLED_JVM is set to java.home, which in my case is "C:\Program Files\Java\jdk1.8.0_45\jre". Note "\jre" at the end.

The trouble is, jre\release does not exist - the release file is only present in the top-level jdk directory. Changing JAVA_HOME and/or my path does not help as it still seems to pick the contained jre regardless of how I set this up. Also, if I set JAVA_HOME to point to a standalone JRE, the build complains that it should point to the JDK.

I have solved this by simply trying to get the properties from both possible locations:

    
    <property file="${WINDOWS_BUNDLED_JVM}/release" prefix="windows"/>
    <property file="${WINDOWS_BUNDLED_JVM}/../release" prefix="windows"/>

(See pull request [#3414](https://github.com/arduino/Arduino/pull/3414).)

Note that it is necessary to use &lt;property file=...&gt; instead of &lt;loadproperties srcfile=...&gt;, as the latter produces an unavoidable fatal error if the source file is missing. It does not seem to matter what order these lines are in - I have checked both ways.

I am reluctant to go any further to make sure that this only includes at-most-one properties file, because of the complexity of getting Ant to do anything so outlandish as write an "if" statement.

PR #3414, opened 2015-06-24
Windows build: look for JDK release file in parent dir.

martinjos:
Fixes [#3413](https://github.com/arduino/Arduino/issues/3413) (Windows build fails, because it looks for JDK release file in the wrong place)


Chris--A:
This is confirmed. I had previously copied the release file into the jre folder to fix this.

Issue #3412, opened 2015-06-24
Legacy libraries linked multiple times, causing duplicate symbol errors


martinjos:
In the latest trunk, there appears to be a problem with including multiple headers from the same legacy library. The problem does not occur in 1.6.5.

As an example, if you take the MPU6050 motion processor library (Arduino variant) from the current i2cdevlib, there shouldn't be a problem with including both MPU6050_6Axis_MotionApps20.h and MPU6050.h (as the latter has a multiple-inclusion guard).

However, what actually happens is that first, you get two copies of the "Using library MPU6050..." message (with verbose output turned on). And then, when it gets to the final link stage, the object files for this library are present twice in the linker command line. Which causes the link to fail because of all the duplicate symbols.

This indicates that importedLibraries (which is a LibraryList, a subclass of LinkedList) in processing.app.debug.Compiler contains a duplicate UserLibrary. This list is populated within Compiler.preprocess(), after stage 2. To guard against duplicates, this code calls LibraryList.contains(UserLibrary), which is implemented by LinkedList.contains(), which uses UserLibrary.equals() to test for equality, which is implemented by ContributedLibrary.equals().

Only one line has changed in ContributedLibrary.java since 1.6.5:

$ git diff 1.6.5 arduino-core/src/cc/arduino/contributions/libraries/ContributedLibrary.java
&lt; boolean versionEquals = thisVersion == otherVersion || (thisVersion != null && otherVersion != null && thisVersion.equals(otherVersion));
&gt; boolean versionEquals = thisVersion != null && otherVersion != null && thisVersion.equals(otherVersion);

I have just confirmed that reverting this line to the 1.6.5 version removes the problem described above. This must be because the version for a legacy library is null, causing the former condition to be true, and the latter condition to be false.

This change happened as part of ffissore's ["Code cleanup"](https://github.com/martinjos/Arduino/commit/fbb61ffd3cb1162b53b3d2aa5028f808a923c3e4) of 6 days ago.

I would suggest adding a comment about the fact that version strings for legacy libraries are null (and two legacy libs should always pass this test), to prevent future regressions.

I've created a pull request ([#3415](https://github.com/arduino/Arduino/pull/3415)).

ffissore:
Fixed by #3415

PR #3415, opened 2015-06-24
Handle duplicates of legacy libs correctly. (#3412)

martinjos:
Fixes issue [#3412](https://github.com/arduino/Arduino/issues/3412) (Legacy libraries linked multiple times, causing duplicate symbol errors).


ffissore:
Thanks! Good catch!

PR #3420, opened 2015-06-25
Serial monitor: added check when setting serial port parameters

cmaglie:
This allows to detect for invalid baud rate settings in particular on Linux where the kernel do not allows non-standard baud rates on some devices.

See #3389 #3351

### DIFF
--- a/arduino-core/src/processing/app/Serial.java
+++ b/arduino-core/src/processing/app/Serial.java
@@ -112,7 +112,11 @@ public class Serial implements SerialPortEventListener {
     try {
       port = new SerialPort(iname);
       port.openPort();
-      port.setParams(irate, idatabits, stopbits, parity, true, true);
+      boolean res = port.setParams(irate, idatabits, stopbits, parity, true, true);
+      if (!res) {
+        System.err.println(format(_("Error while setting serial port parameters: {0} {1} {2} {3}"),
+                                  irate, iparity, idatabits, istopbits));
+      }
       port.addEventListener(this);
     } catch (SerialPortException e) {
       if (e.getPortName().startsWith("/dev") && SerialPortException.TYPE_PERMISSION_DENIED.equals(e.getExceptionType())) {


matthijskooijman:
Would  you still want to open up the serial monitor when settting the parameters fails? AFAIU the error is shown in the main window, so opening the monitor could mask the error?

For completeness, the other setParams call (for doing the 1200baud touch IIRC) should also be checked?

cmaglie:
Well those are GUI considerations:

> Would you still want to open up the serial monitor when settting the parameters fails?

The widget to change speed is displayed within the serial monitor window, if we keep the serial monitor closed the user has no way to change the speed again to a supported value and it's basically stuck.

> AFAIU the error is shown in the main window, so opening the monitor could mask the error?

Yes and it may be covered by the serial monitor window, not exactly the best, but in my test it worked nicely. How can we do otherwise? a possibility is to pop-up a warning in a dialog box, but this is very annoying because forces the user to click the "dismiss" button every time.

> For completeness, the other setParams call (for doing the 1200baud touch IIRC) should also be checked?

Yes I can easily add the check there too, in this case printing the error on the console is fine I guess.

matthijskooijman:
> The widget to change speed is displayed within the serial monitor window, if we keep the serial monitor closed the user has no way to change the speed again to a supported value and it's basically stuck.

Oh, good point.

> Yes and it may be covered by the serial monitor window, not exactly the best, but in my test it worked nicely. How can we do otherwise? a possibility is to pop-up a warning in a dialog box, but this is very annoying because forces the user to click the "dismiss" button every time.

OTOH, how often will you be running into this error? A dialog box sounds acceptable to me.

A better alternative would be to show the error in some part of the serial monitor window. The data area probably isn't the right place, that would be confusing. Putting a label or something next to the "autoscroll" checkbox would work, but could get messy quickly. Ideally you'd just have an "error" bar that gets added above the bottom bar (and auto-hides perhaps), but that's probably a lot of work for this small usecase?

NicoHood:
Could someone retrigger the bot?

matthijskooijman:
@NicoHood, I don't think this PR was changed since the last build? If you want to build this PR on top of git master, @cmaglie will have to rebase it first.

NicoHood:
I just wanted to try it with the recent master. If it works why not include this simple fix.

ffissore:
Rebased and merged with ec7cc8c5cf6cc766b28f554729bbd98e00188d98

PR #3421, opened 2015-06-25
Introducing APP_DIR property

### LINKED ISSUE #1493: arduino commandline only full path accepted
the .ino full path must be given.

either of following two will not work
$arduino some_proj.ino 
$arduino ./some_proj.ino
it works only in this way.
$arduino  `pwd`/some_proj.ino

my environment.
arduino version: downloaded from here http://downloads.arduino.cc/arduino-1.5.2-linux64.tgz
java version "1.7.0_21"
OpenJDK Runtime Environment (IcedTea 2.3.9) (7u21-2.3.9-0ubuntu0.12.04.1)
OpenJDK 64-Bit Server VM (build 23.7-b01, mixed mode)

ffissore:
It allows the IDE to know its installation folder.

Also upgrading appbundler with a patched version: https://bitbucket.org/ffissore/appbundler
It allows to know the current working directory

These two put together, we can now rely on APP_DIR when loading IDE resources while being sure current working directory is properly set, thus being much more friendly when run from CLI

Fixes #1493

/cc @wayoda @haobug @cmaglie @matthijskooijman 

We need to make a windows installer as well, to be sure this doesn't break double clicking a `.ino` file


matthijskooijman:
AFAIU, the OSX bundler app already recognized the $APPROOT variable, but you just modified to not change the current directory, right? Any plans on submitting these changes upstream?

And wouldn't this allow the --curdir handling to be removed?

matthijskooijman:
Oh, and kudos for diving into this! Thanks!

ffissore:
The bundler knows about APP_ROOT but doesn't propagate it to the app. I've submitted the changes upstream: https://bitbucket.org/infinitekind/appbundler/pull-request/32/workingdir-set-to-nsfilemanager/diff

Indeed this makes the --curdir hack useless. However, I'd leave it there for backwards compatibility

ffissore:
I've removed the indentation change for a clearer diff

ffissore:
@ArduinoBot build this please

ffissore:
Whoops, didn't push `.sha` file because of a too wide `.gitignore`. Force pushed

matthijskooijman:
> The bundler knows about APP_ROOT but doesn't propagate it to the app.
I'm confused - it seems APP_ROOT is already interpolated into (among others) all option values, see [this bit of code](https://bitbucket.org/infinitekind/appbundler/src/f715a34bff953ab03b81aa278fe5c61340583200/appbundler/native/main.m?at=default#cl-360)

The only thing your patch seems to change is, when no "WorkingDirectory" is specified, do noop chdir (to the current directory) instead of to the home directory? (which is fine, just seeing if I understand this correctly)?

> Indeed this makes the --curdir hack useless. However, I'd leave it there for backwards compatibility
Compatibility for what exactly? People running java directly passing --curdir (seems unlikely). Or do you want to keep the code around in case we'll need it later? (seems tricky - if it's not needed it will likely bitrot and no longer work when you need it). In any case, I'd remove the --curdir option from the arduino shell script, that only complicates matters and potentially causes inconsistencies between platforms.


Additionally, there seems to be a stray "w$" (and a missing full stop) in your commit message?

ffissore:
No, that bit of code replaces occurrences of the `$APP_ROOT` placeholder with the actual app root (which is /bla/bla/Arduino.app). I use that feature [here](https://github.com/arduino/Arduino/pull/3421/files#diff-ee1fbe662440a29a5f7dd4fc65a33062R346)

You're right about my patch, expect I fear it's not a noop chdir. I'm not sure, but otherwise I wouldn't be able explain this [code](https://bitbucket.org/infinitekind/appbundler/pull-request/3/add-support-for-setting-a-working/diff): if it was a noop, `chdir` should have been called only when `WorkingDirectory` was set, instead it's called anyway.

Compatibility with those users used to the bug and calling the IDE from the CLI setting CWD to an arbitrary value, different from the actual CWD
I just don't want this kind of scripts to fail tomorrow
```
cd Arduino.app/Contents/
CURDIR=`pwd`
cd MacOS
./Arduino --curdir $CURDIR --verify Java/examples/01.Basics/Blink/Blink.ino --board arduino:avr:uno
```
I'll include removal of `--curdir` from the manpage

ffissore:
Turns out `--curdir` was undocumented. Oh well.
Fixed commit message. Thanks!

matthijskooijman:
> No, that bit of code replaces occurrences of the $APP_ROOT placeholder with the actual app root (which is /bla/bla/Arduino.app). I use that feature here

That's exactly what I meant when I said "AFAIU, the OSX bundler app already recognized the $APPROOT variable,". I think we agree, in any case.

> You're right about my patch, expect I fear it's not a noop chdir. I'm not sure, but otherwise I wouldn't be able explain this code: if it was a noop, chdir should have been called only when WorkingDirectory was set, instead it's called anyway.

It's called in the second case to chdir to $HOME, which is likely not the cwd yet, so it was not a noop previously. Or am I totally misunderstanding you here?

> Turns out --curdir was undocumented. Oh well.

Yes, it was only meant as an option for internal use, never meant to be used externally. IMHO, anyone that was using it anyway (if any at all) are on their own, no need to keep messy code around for them...

ffissore:
`--curdir` dropped, left a blocking error message, in order to force those that have used it to upgrade their scripts

PR #3435, opened 2015-06-29
Recusively copying additional files

### LINKED ISSUE #3080: Add sketch subdir in the Include Path

ffissore:
Compiler: additional files with allowed extentions are recursively copied to build folder. Fixes #3080

### DIFF
--- a/arduino-core/src/cc/arduino/utils/Pair.java
+++ b/arduino-core/src/cc/arduino/utils/Pair.java
@@ -1,0 +1,42 @@
+/*
+ * This file is part of Arduino.
+ *
+ * Copyright 2015 Arduino LLC (http://www.arduino.cc/)
+ *
+ * Arduino is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Hm, something I realized about this change: Before, you could modify and compile a sketch without having to save your changes, I believe. With this change, this only applies to .ino files, but any unsaved changes to other files will be ignored I think?

Wouldn't it be more consistent anyway to add any files in subdirs to the sketch's file list (and have them opened in the IDE) instead of modifying just the compile process?

ffissore:
Right, I took the wrong way when fixing this and it popped up with #3440. I've reverted to the previous code in 9a01e306751a2766958a0a7b381ee1938867f8a3 and isolated the copy of files only to those in subfolders, so now it works again as it used to
Adding subfolders files as tabs would be nice but it's quite a change. Can you open a new dedicated issue?

Issue #2428, opened 2014-11-08
No support for I2C restart condition at DUE; parameter sendStop in endTransmission not considered


AmLimit:
With a DUE I try to connect a MMA8452 but get wrong answers. On scope I can see that the parameter "sendStop" in "uint8_t TwoWire::endTransmission(uint8_t sendStop)" has no effect - instead of a restart condition a stop condition followed by a new start condition is perfomed.
A look into source code offers that the parameter "sendStop" is not used. The header comments are ready but the code doesn't reflect the description.
As I understand the MMA845x manual a restart condition is necessary to handle queries, so the MMA8452 is currently not controllable by the wire library.
May be there are additional issues in twi.c to handle that issue.

AmLimit:
If the item will not be fixed in nearer future: is there any work around to support a MMA845x accelerometer on a DUE?

AmLimit:
Some good news: I found a way to use my MMA8452 on a DUE :-)

To perform the needed repeated start condition to handle the MMA8452 you have to use the internal address functionality of the SAM controller (Chapter TWI / Master Mode / Internal Address). Due to the MMA8452 uses 7-bit addressing this feature offers up to 3 automatically sent request bytes - you don't have to handle it manually (quite a good feature). If I didn't miss any information within the documentation the MMA8452 is always controlled by 1 byte register requests.
To use this feature with my DUE I added another overloaded method of TwoWire::requestFrom to support the needed parameter iaddress (and optionally isize) to TWI_StartRead:

\arduino-1.5.8\hardware\arduino\sam\libraries\Wire\Wire.cpp
===========================================================

uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, uint32_t iaddress, uint8_t isize) {
  if (quantity > BUFFER_LENGTH)
  quantity = BUFFER_LENGTH;

  // perform blocking read into buffer
  int readed = 0;
  TWI_StartRead(twi, address, iaddress, isize);
  do {
    // Stop condition must be set during the reception of last byte
    if (readed + 1 == quantity)
    TWI_SendSTOPCondition( twi);

    TWI_WaitByteReceived(twi, RECV_TIMEOUT);
    rxBuffer[readed++] = TWI_ReadByte(twi);
  } while (readed < quantity);
  TWI_WaitTransferComplete(twi, RECV_TIMEOUT);

  // set rx buffer iterator vars
  rxBufferIndex = 0;
  rxBufferLength = readed;

  return readed;
}

[...]

PR #3442, opened 2015-07-01
Wire: Support for repeated starts added to sam Wire library

mattwlowe:
[Issue 2428.](https://github.com/arduino/Arduino/issues/2428)

Description:

Adds support for repeated starts to the Wire library on sam. Currently a stop is forced after every write transaction. Some devices are not tolerant of this behavior, such as the MPR121 capacitive touch sensor and MMA8452Q 3-axis accelerometer. 

Interface changes made to the sam Wire library were also implemented in the avr Wire library to ensure they remain congruent. Behavior of both libraries remain unchanged when called using any previous methods. New behavior is only reachable via the new overloaded member functions.

Tests:

Tested on Due and Mega with MPR121. Verified with logic analyzer.


facchinm:
Hi @mattwlowe , I took a look at your patch and it doesn't seem ready for upstreaming (magic numbers, unclear operations)
What is wrong with @AmLimit solution on #2428 ? Does't it work with MPR121?

mattwlowe:
Hi @facchinm, this patch **is** @AmLimit's solution. That solution adds a new overloaded method which accepts additional parameters. What I've done is also add the equivalent functionality to the avr Wire library so it behaves consistently across architectures.

I did explore other options to improve the clarity of the code. You can look over those with the two recent commits. While cleaner looking those options bring endianess into the equation. My vote is still for the first solution which seems the most elegant to me. Honestly, I'm not sure a 1-byte shift really constitutes a magic number anyway, but you have three options to choose from. 

It would be nice to get **something** merged upstream to address this issue. There are dozens of forum posts and multiple open issues that would likely be addressed with this patch. If you read the comments on those you'll see it's giving people quite a bit of grief. The bug itself is hard to track down because it requires a logic analyzer and knowledge of the I2C protocol (see below). Different devices also handle the improper requests differently and thus there are a wide range of possible failure modes.

broken_due_rs

facchinm:
Hi @mattwlowe , my apologize, I didn't pay much attention at the SAM part because I knew that AVR already had a way to send a repeated start (through `endTransmission(false)` ).

Now, exposing the same API is always a good thing, I was only concerned by the number `3` and the little explanation about what was going on. You are surely an I2C expert but someone else could read the code and being unable to understand what is happening without some proper comment. 

Something like the following could be OK: 
```
// send a repeated start for device which need to access an internal address
// this function is executed by the hardware TWI module on other processors (for example Due's TWI_IADR and TWI_MMR registers)
```
By the way, the first version of the byteshift was better than the following ones, so I'm going to merge your first commit with some comments added. Of course you can modify them later with another PR.

Thank you very much

PS: I'm leaving the screenshot from SAM3X datasheet here for reference
2015-07-03-163408_1280x768_scrot

facchinm:
Manually merged with 39c3f8b

PR #3444, opened 2015-07-01
LiquidCrystal Library changes for Galileo and Edison compatibility

calvinatintel:
Implemented by @intel-larry
Reviewed by @manoelramonintel


facchinm:
The patch is really clean and moving the initialization from `init()` to `begin()` has no side effect on AVR and ARM boards.
Merging, thank you very much!

Issue #3290, opened 2015-06-04
Wire: reset while transmitting hangs the slave forever


jantje:
When Arduino is reset when it is communicating with a slave over I2C the slave will try to send it's data to arduino for ever and always making the wire code to behave .... well not nice.
This situation can be fixed easily by putting the clock pin high and low as if arduino is reading; during startup.
It would be cool if this code could be part of wire.begin
    pinMode(SCL, OUTPUT);
    for (int i = 0; i < 8; i++)
    {
        digitalWrite(SCL, HIGH);
        delayMicroseconds(3);
        digitalWrite(SCL, LOW);
        delayMicroseconds(3);
    }
    pinMode(SCL, INPUT);
See more background info here http://forum.arduino.cc/index.php?topic=288573.5

PR #3454, opened 2015-07-02
Wire: fix slave hanging forever if master is reset during transmission

jantje:
As requested to add in #3290
I only did the change for the master begins to avoid flipflopping if more than 1 Arduino is used on a I2C.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ avr-toolchain-*.zip
 manifest.mf
 nbbuild.xml
 nbproject
+*~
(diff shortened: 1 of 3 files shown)


stickbreaker:
I think there is a possible problem with this code change.  The bit twiddling to produce 8 clock cycles on SCL actively drive the bus HIGH.  The I2C bus is only spec'd to be driven LOW.  The pullup resistors do the HIGH drive.  if you have another master driving the bus low while this code is driving the same SCL line HIGH which ever 'master' has better drivers wins.  This could damage either 'master' .  Instead of `digitalWrite(SCL,HIGH);` use `pinMode(SCL,INPUT_PULLUP);`  This will produce the same effect safely.

Chuck.

Issue #2911, opened 2015-04-04
Errors if IDE cannot find a Leonardo board


NicoHood:
Open an empty sketch, select Leonardo/Micro board and hit upload with no board connected.
You should remove that debug output:

```
processing.app.debug.RunnerException: Couldn't find a Board on the selected port. Check that you have the correct port selected.  If it is correct, try pressing the board's reset button after initiating the upload.
	at cc.arduino.packages.uploaders.SerialUploader.waitForUploadPort(SerialUploader.java:234)
	at cc.arduino.packages.uploaders.SerialUploader.uploadUsingPreferences(SerialUploader.java:130)
	at processing.app.debug.Compiler.upload(Compiler.java:165)
	at processing.app.Sketch.upload(Sketch.java:1260)
	at processing.app.Sketch.exportApplet(Sketch.java:1202)
	at processing.app.Sketch.exportApplet(Sketch.java:1174)
	at processing.app.Editor$DefaultExportHandler.run(Editor.java:2471)
	at java.lang.Thread.run(Thread.java:745)
Couldn't find a Board on the selected port. Check that you have the correct port selected.  If it is correct, try pressing the board's reset button after initiating the upload.
```

Edit: 1.6.3 on Win8

ffissore:
Fixed by #3483

PR #3483, opened 2015-07-08
Hide stack trace when the SerialUploader cannot find the selected board

sandeepmistry:
This resolves issue #2911. Fix tested on OS X.


ffissore:
Thanks!

PR #3484, opened 2015-07-08
Autocomplete 

ricardojlrufino:
Working log 
# Autocomplete

(https://gist.github.com/ricardojlrufino/19f89884162be39d851a)
## Indexing of Libraries Classes

```
 [*] Arduino Core (AVR)
    NOTE: I believe you need to make manual adjustments, remove some useless things and add other (eg contants HIGH, LOW)
          I'm not reading the #DEFINE for 'variables', because it will turn a mess.  
 [x] Multi Editor (many instances) support.
 [x] User Lib's(3rd-party libs)
 [x] Arduino Default Lib's
 [x] Inheritance Support  (Ex: Serial -> Stream -> Print)
 [ ] Index only the important files of libs (for now is taking all .h)  
```
## Autocomplete methods Classes

```
- [X] Instances
- [X] Static Classes
- [X] Pointer 
- [X] Get the parameters of methods
- [X] Auto-Complete using partial name method (contains!)
- [X] Order by type
- [x] Handle Struct
     [x] Handle static definition (Ex: EEPROM)
- [*] Handle extern (Ex: extern SPIClass SPI || extern EthernetClass Ethernet).
      [] Multiple alias Ex: Serial1, Serial2, etc...
- [*] '#define' statements 
      I'm not reading the #DEFINE for 'variables', because it will turn a mess. 
      NOTE: not parsing blocks like: "#if defined(UBRR1H) extern HardwareSerial Serial1;" in HardwareSerial.h    
  [*] Disable on External MODE.
```
## Sketck - Autocomplete code being edited

```
  [x] Show only included Classes.
  [x] Show the variables of the current scope (use offset techniques)
     [x] Detect current function (based at the cursor location)
     [x] Update the position (offset) of the functions when the document is modified (new lines), without making the parser.
     |- NOTES:
     |- In Autocomplete they must appear first (deone).
     |- Not perfect, variables declared in blocks (if / while) are considered as function scope, to improve the performace
  [] Renaming / save as you must update the name in the Autocomplete Cache
     - have to remove the libraries that BEEN REMOVED (imports removed ..) ...
  [x] Parse source in-memory (do not need to save)
  [x] No show 'functions' outsite functions, only show classes
  [x] Detect custom variable names (EthernetClient client)
  [x] When you open the existing Sketch loads the autocomplete for the added LIBs
  [*] Parser in real time: 
      [x] Make the parser of changed line (ending with ';') and extract variables (and possibly errors).
      [x] Perform parse if you are changing a line with errors (syntax)
      [x] Perform parse if you are changing a line with errors (after compile)
      [x] Perform parse if you are changing a line with variable declararion.

      [] Enhancement: Parser current line using REXEX (for performance)
      - IMPL NOTE: For while it is being made a parser all code, is fast but could be better
                   The whole parser was necessary because i had to update the position (offset) of the methods in the document, 
                   but now this is already being done, I think a REGEX can be better    
      - NOTE: Only a parser(for same file) can run at the same time
```

[x] Async(Multi-tread) library parser... (based on Eclipse CDT AST Parser) 
    This is a external project: 'cplus-libparser', which can be used in other projects. ;)          

[ ] Auto-Detect libraries dependency.
   Related BUG: https://code.google.com/p/arduino/issues/detail?id=236
                https://github.com/arduino/Arduino/pull/2174

[x] Auto generate variable names (ALT + ENTER) inside a function call (Ex: digitalRead(var))
    [x] After inserting into code, it must also appear in autocomplete

[*] Auto-Complete of the main functions parameters
  Ex: Serial.begin({lista develocidades}), digitalWrite( {variables} , [HIGH / LOW])
    [ ] This information should be captured from the source code (documentation)

[*] Show syntax(parser) error's
    NOTE: need to be well tested, I did a few tests

[x] Autocomplete for #include pick up all the libraries;
    NOTE: The window of the options just close with ESC or when you click with the MOUSE out, not close with TAB

[*] - Update LibraryIndex/Autocomplete on change libs of sketch
    [x] Fire on Import from Menu
    [ ] Fire on Import from ZIP.

[*] Store/Load parsed result in lib.index on each folder..
    [ ] establish policies to update the index (LibraryChangeMonitor (only for h. files))
        - How to deal with the upgrade of core libraries? must validate the version?
    [x] You may want to implement a Serializer / Deserializer (NOTE:for now the index is binary )

[] Use 'javax.swing.text.Document' to parser, to avoid strings allocations
   TODO: It has not been confirmed whether there is this problem. but I think so ...
   Need implement: org.eclipse.cdt.internal.core.parser.scanner.AbstractCharArray

[x] Automatic adjustment of the position of the tooltip(completions window) to left or right

[] BUG(linux): The autocomplete window not working when you have another window "always on TOP"

[x] Autocomplete Shortcut's
- dw,dr,aw,ar (digitalWrite, digitalRead, analogWrite, analogRead)
- for,if,while,println

[x] Autocomplete of Sketch with multiple files
   [x] BUG: When the SAVE event is launched cache and clean and only the modified classes are analyzed
   [x] includes with global methods.
   [x] includes with structs and static var.
   [x] SyntaxErrorMarker - when a error occur, the highlight appear in two tabs
   [x] #include - add other files of current Sketch in autocomplete for #include

[x] Compiler/SyntaxError - show multiple compiler erros at same time

[x] BUG - autocomplete cache/listeners problem on 'saveAs' sketch (It is keeping two versions)

[ ] Parse AVR Includes (/hardware/tools/avr/avr/include)
    NOTE: At the moment I'm not sure which files should be done parser
    PROBLEM: Including this autocomplete list can become large and confusing


Chris--A:
I saw the video of this a while ago and it looks really cool.
I would really like to try this out once the merge conflicts are resolved.

facchinm:
Superseded by #6235

PR #3485, opened 2015-07-08
This adds 'hh' and 'hpp' as valid sketch files.

Chris--A:
For completeness a sketch should allow the same extensions as a library. These missing sketch extensions have already been enabled for libraries in #3186.


ffissore:
Thanks!

PR #3496, opened 2015-07-10
WiFi: move init from constructor to method

facchinm:
Zero SPI library is not responding correctly when invoked before main()

AVR compatibility ok
fixes arduino/ArduinoCore-samd#18


facchinm:
@agdl @cmaglie

Issue #3321, opened 2015-06-11
ArduinoISP: reliability and portability improvements.


PeterVH:
(This PR should supersede: #2314, #2315.)

I recently picked up the effort to make ArduinoISP more reliable and better working on more (hopefully all) arduino models. I started again from rsbohn's last work and try to fix only things that help achieving these two goals.

This effort is in [branch issue-3321 of my Arduino fork] (https://github.com/PeterVH/Arduino/tree/issue-3321) and  in the [master branch of my ArduinoISP fork](https://github.com/PeterVH/ArduinoISP/tree/master)

**Status**

This is still a work in progress, but here is a summary of the status:
* Rsbohn's last work introduced some regressions, but these are fixed.

* On the Uno it works much more reliable than the version in the IDE, mainly by simply removing the delay() call from the heartbeat() method.

* On the Due. There is no bit bang spi yet: I think this would better be handled by a SPI.setSpeed() method in the SPI lib (if the selected speed is too slow, the implementation could automatically switch to bit banged SPI).

Therefore on the Due it is already working (untill now I only tested reading signatures) but not for slow targets like the attiny85...

**Testing:** I am composing a test suite bundling a set of about 10 typical/important use cases for ArduinoISP. The suite needs to be fleshed out, but here is a preview:
1. No target connected.  (note this does not work as it should with the current version in the IDE)
2. Burn UNO bootloader
3. Burn mega bootloader (requires programming above 128KB)
4. Program an attiny85
5. Program an attiny841
6. Program an at89s52
7. Chaucer115 to 1284p: Program a big (115KB) sketch into an atmega 1284p
8. Chaucer115 to mega: Program a big (115KB) sketch into an arduino mega
9. LongStoryShort: Program above 128KB.

(the name Chaucer comes from the optiboot test suite that burns big sketches that output part of the "knight's tale" by Chaucer on the console)

**Avrdude:** I think it would be better to upgrade to avrdude 6.1 because this version implements some clever fixes that make it possible to program high addresses (over 128KB) with stk500v1 (which is what ArduinoISP uses). This makes it possible to reliably burn the bootloader of an atmega2560 using ArduinoISP.

facchinm:
Hi @PeterVH , I'm keeping an eye on your repo and it seems that the Bitbang SPI on Due is solved :metal:
If you could open a proper PR the bot will automatically create a build so people could start testing it easily!

PeterVH:
[...]

Issue #1182, opened 2012-12-21
ArduinoISP on the Leonardo does not work on windows


PeterVH:
Versions:
Tried this on windows 7 SP1 and on XP SP2 (though on XP SP2, composite device stuff must be stripped to bring up leanardo all together).
The problem does not occur on linux.
Arduino 1.0.2

What steps will reproduce the problem?
From the IDE, perform following steps:
1. Upload the ArduinoISP sample to the leonardo.
2. Select ArduinoISP as programmer
3. In the preferences select verbose output during uploads.
4. Select an arbitrary board type. (no need to actually connect that target to the leonardo, we just want to see whether serial communictation between pc an leonardo works).
5. Perform "burn bootloader".

Expected output: avrdude should send some bytes to the leonardo and the latter should send some bytes back (this should of coarse result in a failed upload because no target is attached)

Actual output: Leonardo sends nothing back, TX led never lights up. (RX led does)

Analysis: It turns out that on windows when avrdude opens the serial port, neither DTR nor RTS are asserted. Under these conditions, the leonardo does not send out data, see CDC.cpp:

``` C++
size_t Serial_::write(uint8_t c)
{
        /* only try to send bytes if the high-level CDC connection itself 
         is open (not just the pipe) - the OS should set lineState when the port
         is opened and clear lineState when the port is closed.
         bytes sent before the user opens the connection or after
         the connection is closed are lost - just like with a UART. */

        // TODO - ZE - check behavior on different OSes and test what happens if an
        // open connection isn't broken cleanly (cable is yanked out, host dies
        // or locks up, or host virtual serial port hangs) */
        if (_usbLineInfo.lineState > 0) {
                int r = USB_Send(CDC_TX,&c,1);
                if (r > 0) {
                        return r;
                } else {
                        setWriteError();
                        return 0;
                }
        }
        setWriteError();
        return 0;
}
```

[...]

PR #3500, opened 2015-07-10
Improve ArduinoISP sketch

PeterVH:
This is the pull request for [issue 3321](https://github.com/arduino/Arduino/issues/3321)

More info on this change request is [here](https://petervanhoyweghen.wordpress.com/2015/07/21/arduinoisp-reliability-and-portability-improvements).

I will update the progress of testing effort in the table below:

*Test report for linux (Kubuntu 14.04 LTS):*

Test   | Duemilanove | UNO | Leonardo | Due | Mega
-------------------------------------------|------------------|---------|------------|-------|-------------
No target                                       |  OK  | OK  |OK | OK | OK 
Burn UNO bootloader                    | todo | todo |todo | todo | todo 
Burn Arduino Mega bootloader      | OK   | OK |OK | todo | OK 
Program an attiny85 (Chaucer4K) | todo | OK |OK | OK | OK 
Program an attiny2313 (TinyBlink) | OK    | OK |OK | OK | OK 
Program an attiny841 (TinyBlink)   | todo | OK |OK | OK | OK 
Program an at89s52                      | OK | OK |OK | FAIL(1) | OK
Chaucer112K to 1284p                  | OK | OK |todo | OK | OK
Chaucer112K to Mega                   | todo | OK |todo | todo | todo
LongStoryShort                              | todo | todo |todo | todo | todo

(1) an at89s52 works out of spec at 3V3. Downloading hex files, signature etc works but flashing a new hex file requires at least 4V. With a shield containing a 5V buffer this could work...

*Test report for windows:*

Test   | Duemilanove | UNO | Leonardo(3) | Due(3) | Mega
-------------------------------------------|------------------|---------|------------|-------|-------------
No target                                       |  OK| OK(4)|OK| OK | OK(4)
Burn UNO bootloader                    | OK| OK |OK | OK | OK 
Burn Arduino Mega bootloader      | OK| OK |OK | OK| todo (5) 
Program an attiny85 (Chaucer4K) | OK| OK |OK | OK | OK 
Program an attiny2313                  | OK| OK |OK | OK | OK 
Program an attiny841                    | OK| OK |OK| OK | OK 
Program an at89s52                      | OK| OK |OK(1) |FAIL(2) | OK
Chaucer112K to 1284p                  | todo | todo |OK | OK| todo
Chaucer112K to Mega                   | todo | OK |OK | OK| todo (5)

(1) used pre compiled hex files, available in the ArduinoISP-test repo.
(2) an at89s52 works out of spec at 3V3. Downloading hex files, signature etc works but flashing a new hex file requires at least 4V. With a shield containing a 5V buffer this could work...
(3) requires the changes to CDC.cpp proposed in #1182.
(4) the error led flashes briefly because the 16u2 sends a bogus byte when changing the baud rate, this is harmless, stk500v1 recovers from it
(5) only big spenders have two mega's

### DIFF
--- a/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
+++ b/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
@@ -67,7 +67,7 @@
 void pulse(int pin, int times);
 
 void setup() {
-  Serial.begin(9600);
+  Serial.begin(19200);
   SPI.setDataMode(0);
   SPI.setBitOrder(MSBFIRST);
   // Clock Div can be 2,4,8,16,32,64, or 128
@@ -111,17 +111,16 @@ parameter param;
 uint8_t hbval = 128;
 int8_t hbdelta = 8;
 void heartbeat() {
-  if (hbval > 192) {
-    hbdelta = -hbdelta;
-  }
-  if (hbval < 32) {
-    hbdelta = -hbdelta;
-  }
+  static unsigned long last_time = 0;
+  unsigned long now = millis();
+  if ((now - last_time) < 40)
+    return;
+  last_time = now;
+  if (hbval > 192) hbdelta = -hbdelta;
+  if (hbval < 32) hbdelta = -hbdelta;
   hbval += hbdelta;
   analogWrite(LED_HB, hbval);
-  delay(20);
 }
-
 
 void loop(void) {
   // is pmode active?
--- a/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
+++ b/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
@@ -123,6 +123,13 @@ void heartbeat() {
   analogWrite(LED_HB, hbval);
 }
 
+static bool rst_active_high;
+
+void reset_target(bool reset)
+{
+  digitalWrite(RESET, ((reset && rst_active_high) || (!reset && !rst_active_high)) ? HIGH : LOW);
+}
+
 void loop(void) {
   // is pmode active?
   if (pmode) {
@@ -242,15 +249,24 @@ void set_parameters() {
                     + buff[18] * 0x00000100
                     + buff[19];
 
+  // avr devices have active low reset, at89sx are active high
+  rst_active_high = (param.devicecode >= 0xe0);
 }
 
 void start_pmode() {
   SPI.begin();
-  digitalWrite(RESET, HIGH);
+  // SPI.begin() has configured SS as output,
+  // so SPI master mode is selected.
+  // We have defined RESET as pin 10,
+  // which for many arduino's is not the SS pin.
+  // So we have to configure RESET as output here,
+  // (reset_target() first sets the level correct)
+  reset_target(false);
   pinMode(RESET, OUTPUT);
+
   digitalWrite(SCK, LOW);
   delay(20);
-  digitalWrite(RESET, LOW);
+  reset_target(true);
   delay(50);
   spi_transaction(0xAC, 0x53, 0x00, 0x00);
   pmode = 1;
@@ -258,7 +274,7 @@ void start_pmode() {
 
 void end_pmode() {
   SPI.end();
-  digitalWrite(RESET, HIGH);
+  reset_target(false);
   pinMode(RESET, INPUT);
   pmode = 0;
 }
--- a/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
+++ b/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
@@ -274,6 +274,10 @@ void start_pmode() {
 
 void end_pmode() {
   SPI.end();
+  // We're about to take the target out of reset
+  // so configure SPI pins as input
+  pinMode(MOSI, INPUT);
+  pinMode(SCK, INPUT);
   reset_target(false);
   pinMode(RESET, INPUT);
   pmode = 0;
--- a/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
+++ b/build/shared/examples/11.ArduinoISP/ArduinoISP/ArduinoISP.ino
@@ -1,4 +1,4 @@
-// ArduinoISP version 04m3
+// ArduinoISP
 // Copyright (c) 2008-2011 Randall Bohn
 // If you require a license, see
 //     http://www.opensource.org/licenses/bsd-license.php
@@ -6,42 +6,31 @@
 // This sketch turns the Arduino into a AVRISP
 // using the following arduino pins:
 //
-// pin name:    not-mega:         mega(1280 and 2560)
-// slave reset: 10:               10
-// MOSI:        11:               51
-// MISO:        12:               50
-// SCK:         13:               52
+// Pin 10 is used to reset the target microcontroller.
+//
+// The MISO, MOSI and SCK pins are used to communicate with the target,
+// on all Arduinos, these pins can be found on the ICSP header:
+//
+//               MISO °. . 5V (!) Avoid this pin on Due, Zero...
+//               SCK   . . MOSI
+//                     . . GND
+//
+// On some Arduinos (Uno,...), pins MOSI, MISO and SCK are the same pins
+// as digital pin 11, 12 and 13, respectively. That is why many tutorials
+// instruct you to hook up the target to these pins. If you find this wiring
+// more practical, have a define USE_OLD_STYLE_WIRING. This will work even
+// even when not using an Uno. (On an Uno this is not needed).
+// 
+// IMPORTANT: When using an Arduino that is not 5V tolerant (Due, Zero, ...)
+// as the programmer, make sure to not expose any of the programmer's pins to 5V.
+// A simple way to accomplish this is to power the complete system (programmer
+// and target) at 3V3.
 //
 // Put an LED (with resistor) on the following pins:
 // 9: Heartbeat   - shows the programmer is running
 // 8: Error       - Lights up if something goes wrong (use red if that makes sense)
 // 7: Programming - In communication with the slave
 //
-// 23 July 2011 Randall Bohn
-// -Address Arduino issue 509 :: Portability of ArduinoISP
-// http://code.google.com/p/arduino/issues/detail?id=509
-//
-// October 2010 by Randall Bohn
-// - Write to EEPROM > 256 bytes
-// - Better use of LEDs:
-// -- Flash LED_PMODE on each flash commit
-// -- Flash LED_PMODE while writing EEPROM (both give visual feedback of writing progress)
-// - Light LED_ERR whenever we hit a STK_NOSYNC. Turn it off when back in sync.
-// - Use pins_arduino.h (should also work on Arduino Mega)
-//
-// October 2009 by David A. Mellis
-// - Added support for the read signature command
-//
-// February 2009 by Randall Bohn
-// - Added support for writing to EEPROM (what took so long?)
-// Windows users should consider WinAVR's avrdude instead of the
-// avrdude included with Arduino software.
-//
-// January 2008 by Randall Bohn
-// - Thanks to Amplificar for helping me with the STK500 protocol
-// - The AVRISP/STK500 (mk I) protocol is used in the arduino bootloader
-// - The SPI functions herein were developed for the AVR910_ARD programmer
-// - More information at http://code.google.com/p/mega-isp
 
 #include "Arduino.h"
 #undef SERIAL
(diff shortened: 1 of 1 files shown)


ffissore:
It doesn't merge. Can you please rebase the code on top of latest master?

facchinm:
Thank you @PeterVH ! Unfortunately a recent change moved the sketch from `ArduinoISP` folder to `11.ArduinoISP/ArduinoISP` to be coherent with the rest of the examples.

Could you rebase upon the current `master` branch?
Thank you very much

PeterVH:
@facchinm: Done. But this forced me to push -f again.

facchinm:
@PeterVH no problem at all, thank you very much! I'm launching the PR builder to get a complete IDE image with your sketch so anyone interested can test it!

facchinm:
@ArduinoBot build this please

q2dg:
@PeterVH Maybe it's a sillyness, but do you know https://github.com/adafruit/ArduinoISP ?? For gathering some ideas...

facchinm:
@ArduinoBot build this please

PeterVH:
I am afraid I introduced some regression. Some tests,  e.g. the at89s52 fail (burning works but verifying fails...) Will have to check this out.
B.t.w. there is partial information on the tests [here](https://petervanhoyweghen.wordpress.com/2015/07/21/arduinoisp-reliability-and-portability-improvements/)

PeterVH:
It is fixed, several things were wrong: see commit messages.

Also, omitting the error led pulses on duplicate enter pmodes (commit 5a04ab2a) had a side effect for the at89s52: these pulses took 180 msec and the at89 needed this as an extra delay after the chip erase. The correct way to deal with this is through the chip_erase_delay attribute in avrdude.conf. The at89s52 datasheet says this takes 'about 500 msec' so I use 600000. Now my knight rider hex files bur happily again.

Finally I fixed the compiler warnings.

PeterVH:
I think this is about it for now. I will start more systematic testing now. I will update the progress of this in the first entry of this issue.

If someone has access to a Zero or even a Gallileo (@facchinm ?) and could give it a try (even a subset of the tests)  that would be great. Also I have no mac. Of coarse every other form of help in testing is welcome (a Windows adept certainly would save me pain).

@q2dg : it is good to take a look at other forks indeed, and I do that regularly. But note that my approach is to focus on fixing up the sample delivered with the IDE, rather than adding new
features (the at89s52 support being an exception).

facchinm:
@ArduinoBot build this please

facchinm:
Thank you very much @PeterVH for your effort! I'll test it soon on more board and on every OS! :metal:

facchinm:
Hi Peter, I and @cmaglie extensively tested the sketch on a great variety of boards (not with a lot of targets unfortunately).
Zero is working after applying this small patch
```C 
digitalWrite(MOSI, (b & 0x80) ? HIGH : LOW);
```
instead than
```C
digitalWrite(MOSI, b & 0x80);
```
Other platforms using bitbanging works well (Galileo, Edison, Tre and obviously Due).
We didn't test at89s52 or attiny targets but I'm quite confident they should work if working on Due (which you can use as reference platform for bitbanging).

I'd merge it after patching for Zero; in the meantime if you can share the test suite we can setup some automatic testing environment with all the other boards.

Thank you very much!

matthijskooijman:
@facchinm, isn't this a limitation of the Zero core? Even though digitalWrite is documented to accept `HIGH` and `LOW`, in practice it accepts any zero or non-zero value. Is there a compelling reason for the zero not to?

facchinm:
@matthijskooijman you are right, the Zero's `digitalWrite()` is a bit too strict.
If we change 
```C
  switch ( ulVal )
  {
    case LOW:
      PORT->Group[g_APinDescription[ulPin].ulPort].OUTCLR.reg = (1ul << g_APinDescription[ulPin].ulPin) ;
    break ;

    case HIGH:
      PORT->Group[g_APinDescription[ulPin].ulPort].OUTSET.reg = (1ul << g_APinDescription[ulPin].ulPin) ;
    break ;

    default:
    break ;
  }
```
into
```C
  switch ( ulVal )
  {
    case LOW:
      PORT->Group[g_APinDescription[ulPin].ulPort].OUTCLR.reg = (1ul << g_APinDescription[ulPin].ulPin) ;
    break ;

    default:
      PORT->Group[g_APinDescription[ulPin].ulPort].OUTSET.reg = (1ul << g_APinDescription[ulPin].ulPin) ;
    break ;
  }
```
it works out of the box. Can you see any drawback in this change? I'm opening a PR on Zero core repo anyway :smile:

facchinm:
Proposed here https://github.com/arduino/ArduinoCore-samd/pull/26

NicoHood:
Also see this:
https://github.com/arduino/Arduino/pull/3687

And this (not completely related):
https://github.com/arduino/Arduino/issues/3568#issuecomment-131507000

I did a quick test with [HoodLoader2](https://github.com/NicoHood/HoodLoader2) and it seems to work (burned a 32u4 HL2.0.5 Bootloader). It would be nice if you could add pin4 for HoodLoader2 devices, since they normally dont have pin 10 broken out and [u2 devices are now supported by the IDE](https://github.com/arduino/Arduino/pull/3640).

Here are my changes (except style changes) that I used before to use the ISP sketch with all kind of Arduinos. Maybe this helps:
https://gist.github.com/NicoHood/8f074c34fcc9a2a0a627/revisions

PeterVH:
Hi all,

Thanks for the feedback and especially for the testing effort. Great to hear it works on all these boards. I have been on holiday without network access which is why it was silent from my side.

@facchinm: 
about the test suite: it just consists of the description of the various use cases and the source code of the test sketches I use, there are (here)[]

PeterVH:
I have found the problem. There was nothing wrong with the new reset sequence code. I was testing with a Mega as programmer and the serial port was set to a baud rate different from 19200, so avrdude has to set the correct baud rate on the fly. This baud rate change causes the Mega to receive a garbage 0xff byte. The stk500 protocol however recovers from this. Interestingly a duemilanove does not have this problem (ftdi usb to serial converter).

I ran most of the tests the tests again, it looks stable.

So 2 micro things left to do on the sketch: hoodloader config and zero thing.

NicoHood:
@PeterVH the mega and the uno have a 16u2 as USB-Serial bridge.

I have two questions regarding this:
* Does the Uno behave the same with the garbage input?
* Could you test the same settup with HoodLoader2 (not as ISP, as bridge only). Because I might have fixed this garbage byte in my bootloader. If not maybe we can correct this error (at least for hoodloader2 since many people start using it)

Also I do not really understand what you mean with
> so avrdude has to set the correct baud rate on the fly.

Does the ISP now work with the Micro/Leonardo?

PeterVH:
> Does the Uno behave the same with the garbage input?

Yes. I just tried it out. 
> Also I do not really understand what you mean with:
>        _so avrdude has to set the correct baud rate on the fly._

Well, avrdude sets the baudrate when it starts programming, this causes a baud rate change in the u2 and this causes the Uno or Mega receive a garbage byte.

> Could you test the same setup with HoodLoader2

Sure. Will first read about hood loader. I intend to use it as a more convenient way to disable autoreset (instead of the capa) on Mega and Uno. (Currently on the Mega and Due I use a modified u2 firmware that disables autoreset if a jumper is placed over a pair of the 4 broken out gpios)

>  Does the ISP now work with the Micro/Leonardo? 

Yes. See the test report I maintain at the top of this thread. (Be it that on windows you still have to do -carduino).

PeterVH:
@NicoHood: I can confirm that the problem of the garbage byte does _not_ occur when using HoodLoader2 as bridge. 
I added pin configuration for hoodloader2 and tested it.

B.t.w, it is a nice concept you worked out. I plan to keep it installed on my Uno. Though for my use case I'd like the inverse the logic: reset twice => sketch mode, press once or powerup => bootloader mode. That way I could leave ArduinoISP in the u2 and when I want to program something, I would activate the sketch by resetting twice. I'll have a look in the code whether that is easy to do...

NicoHood:
So its a problem of the firmware, good to know.

This line fixes this (i think)
https://github.com/NicoHood/HoodLoader2/blob/dev/avr/bootloaders/HoodLoader2/HoodLoader2.c#L852
https://github.com/NicoHood/HoodLoader2/blob/dev/avr/bootloaders/HoodLoader2/HoodLoader2.c#L921

I could invert the logic, but by default it doesnt make sense to me.
What you can do is:
* Modify this line and recompile the bootloader. It will always start with USB-Serial at startup, reset behaviour is still the same. https://github.com/NicoHood/HoodLoader2/blob/dev/avr/bootloaders/HoodLoader2/HoodLoader2.c#L151-L153
* Look at the runbootloader example. You could add a jumper to the 4 pin header and reset if a jumper wasnt set. This way you dont have to recompile the bootloader.

So maybe its time to update the firmware? The Leonardo bootloader also has some things that can be improved. The Arduino team should think about updating the unos firmware/bootloader and the leonardos bootloader as well.

PeterVH:
@facchinm: Hi Martino.I ran the bulk of the tests again, this time on Windows. Everything worked fine, (see table at the top). Also in the forum, the sketch was able to solve various problems. (this [one](http://forum.arduino.cc/index.php?topic=336837.msg2340052#msg2340052) being the most enthousiastic one :-) ).  So I am now confident this is ready to be merged.

On windows I still need the fixes from #1182 for leonardo and due. I should make a separate pull resuest for that.

@cmaglie: you reported success on the Zero. I bought an Arduino M0 (not Pro) to be able to run that kind of tests myself. (You know I don't have access to Zero from Europe). I run with the arduino.org bootloader and use of coarse the arduino.cc ide and core. I adapted the .cc linker script so that the FLASH addresses match what the bootloader expects. The sketch runs: heartbeat led fades in out. But the communication over SerialUSB does not work correctly (see [here](http://forum.arduino.cc/index.php?topic=347972.msg2403614#msg2403614)). I see no a priory reason why the behavior of an M0 would be different than on a Zero (but I may overlook something essential, any ideas?). The symptoms look like problems in the USB driver. Are you sure the sketch (still) works correctly on the Zero?

facchinm:
Hi Peter, I'm really happy to hear such good news! 

The Zero issue is probably related with SerialUSB code being quite problematic, but it was solved in https://github.com/arduino/ArduinoCore-samd/pull/27 ; that PR will be merged soon but you can test it via Board Manager before release :wink: 

Anyway, I'm going to test the actual code (you to make sure, I'm totally convinced it works great) and merge it soon.

NicoHood:
You may want to try if this works for you:
https://github.com/arduino/Arduino/pull/3839

PeterVH:
@fachinm: you were right about the usb problems, with the pr you mentioned I can now happily test ArduinoISP on my M0. Thanks! (This also proves ArduinoISP is a good test case for new cores. In the past it already brought problems in the Leonardo and Due cores to the surface)

There is another detail I wanted to bring to your attention. Someone in the forum brought up that it is confusing having to upload a sketch called 'ArduinoISP' whilst having to specify 'Arduino as ISP' as a programmer. He suggested to rename the sketch to ArduinoAsISP. I don't know. It is true more and more people in the forum start to refer to this programming method as 'Arduino as ISP'.  I'll leave this matter to the Arduino devs...

BTW. the info about ArduinoISP on the new examples web site is inaccurate and adds more confusion (with a link to the ArduinoISP board).

@NicoHood: that would work for me. Some comments:
- Could you also modify the Due core? Will test it if needed.
- Could you also adapt the comment, it does not hold anymore
- The configured flag is not cleared upon disconnection, but that might not be that important right now.

NicoHood:
It is not cleared on disconnect? Where do we have to handle this?
I only found this:

PeterVH:
I have found my way to your pr #3839, comment over there

facchinm:
Time for merge :smile: :smile:  Thanks a lot Peter!

PeterVH:
Thanks Martino!
Fyi, here is a recent [testimony](http://forum.arduino.cc/index.php?topic=349774.msg2414246#msg2414246) I liked from someone using the sketch to burn a bootloader to an atmega 2560, which failes with an usbasp. I would really enjoy people coming back to using ArduinoISP for burning bootloaders and tiny firmware...

NicoHood:
The leonardo windows bug is still not solved, correct?
Just gave it a quick test with HoodLoader2, burned a uno bootloader. Ultra fast. Pin selection is also correct, good job Peter!

q2dg:
Why there is a "Programmer->ArduinoISP" menu option and a "Programmer->Arduino as ISP" whereas there is only one "Examples->ArduinoISP" example sketch?? Which programmer is the good one?? It's very confusing.

PeterVH:
"Arduino as ISP" 
see above : (https://github.com/arduino/Arduino/pull/3500#issuecomment-142421213)

q2dg:
Thanks. I've seen "ArduinoISP" programmer refers to https://www.arduino.cc/en/Guide/ArduinoISP . In this case, I think this should be marked in menu as "retired" or something similar, but I'm not sure.
Sorry for bother you.

Issue #3064, opened 2015-04-30
Bug: analogRead() for Arduino Due should disable previous ADC channel before enabling new one


shahokun:
On the Arduino Due (Arduino library 1.6.1), if you read several different analog pins in succession using analogRead(), some of the values may be off. For example, I had temperature sensors A and B, which were reading approximately the same value. When I added an analogRead() of another channel prior to reading tempA, the reading of tempA would drop a significant amount (~50 ADC counts). By alternating between the two sequences every second, I was able to confirm this behavior:

```
analogRead(tempA);
analogRead(tempB);
// Both readings are approximately the same.
```

```
analogRead(chX);
analogRead(tempA);
analogRead(tempB);
// The reading for tempA is significantly lower
```

I tracked this down to lines 150-156 in wiring_analog.c:
```
// Enable the corresponding channel
if (ulChannel != latestSelectedChannel) {
	adc_enable_channel( ADC, ulChannel );
	if ( latestSelectedChannel != (uint32_t)-1 )
		adc_disable_channel( ADC, latestSelectedChannel );
	latestSelectedChannel = ulChannel;
}
```

If analogRead specifies a new channel, the new channel is first enabled, then the previous channel is disabled. Presumably this could have some undesirable effects with the ADC multiplexer, such as extra current draw or some switching noise (this may also be board layout or timing dependent). However, if you switch the order in code, i.e. disable the previous channel, then enable the new channel, the ADC readings are consistent every time.

```
// Enable the corresponding channel
if (ulChannel != latestSelectedChannel) {
	if ( latestSelectedChannel != (uint32_t)-1 )
		adc_disable_channel( ADC, latestSelectedChannel );
	adc_enable_channel( ADC, ulChannel );

	latestSelectedChannel = ulChannel;
}
```

I think it would be logical to disable the previous ADC channel prior to enabling the next one, and could seemingly prevent multiple channel read errors such as this one.

maxlazarus:
I just had an issue with this as well using an Atmega2560 Arduino, would appreciate this fix.

facchinm:
Could you check if the same misbehaviour happens after applying #2823?
It's a Due-only patch that was never merged for its slight performance drop on single channel reads (see https://groups.google.com/a/arduino.cc/forum/#!topic/developers/IdhTIT0V5GY for extended considerations)
If it solves the issue consider leaving a comment on the PR to speed-up its inclusion :wink:

shahokun:
[...]

PR #3524, opened 2015-07-14
Due: solve issues with nonstandard pin operations

facchinm:
This PR solves the following Due broken behaviours:

* Real issues:
1- pinMode(OUTPUT) -> analogWrite() -> digitalWrite()	[ digital write is not executed ]
2- pinMode(OUTPUT) -> analogWrite() [if executed more than once]	#2455
3- pinMode(OUTPUT) -> digitalWrite() -> pinMode(INPUT) -> digitalRead()	[if executed more than once] #2823 #3064 #2198 
4- pinMode(INPUT) -> analogRead() -> digitalRead()

* AVR core compatibility issues:
5- pinMode(OUTPUT) [ defaults to HIGH ] #3317 #3310
6- pinMode(OUTPUT) -> digitalRead() [ always return 0 ] #1597 #1533
7- digitalWrite() -> pinMode(OUTPUT) [ does not retain the intended value ]

8- acd trigger #1819 #2286 (#3064)

The Due has lots of RAM so the solution uses 80 bytes to keep track of the pinmux state. A lot of AVR behaviours are "emulated" using this array.

The solution has been benchmarked and the speed tradeoff is:
* `-10%` on pure `digitalWrite` and `digitalRead`
* `-5% to -25%`  on pure `analogRead()`
* `-400%` on `analogRead()`followed by `digitalWrite()` on different pins :scream:

All other cases were broken with the actual code so it's useless to test them

### DIFF
--- a/hardware/arduino/sam/cores/arduino/Arduino.h
+++ b/hardware/arduino/sam/cores/arduino/Arduino.h
@@ -155,6 +155,16 @@ typedef enum _ETCChannel
 #define PIN_ATTR_PWM           (1UL<<3)
 #define PIN_ATTR_TIMER         (1UL<<4)
 
+#define PIN_STATUS_DIGITAL_INPUT_PULLUP  (0x01)
+#define PIN_STATUS_DIGITAL_INPUT         (0x02)
+#define PIN_STATUS_DIGITAL_OUTPUT        (0x03)
+#define PIN_STATUS_ANALOG                (0x04)
+#define PIN_STATUS_PWM                   (0x05)
+#define PIN_STATUS_TIMER                 (0x06)
+#define PIN_STATUS_SERIAL                (0x07)
+#define PIN_STATUS_DW_LOW                (0x10)
+#define PIN_STATUS_DW_HIGH               (0x11)
+
 /* Types used for the tables below */
 typedef struct _PinDescription
 {
@@ -169,6 +179,8 @@ typedef struct _PinDescription
   EPWMChannel ulPWMChannel ;
   ETCChannel ulTCChannel ;
 } PinDescription ;
+
+extern uint8_t g_pinStatus[];
 
 /* Pins table to be instanciated into variant.cpp */
 extern const PinDescription g_APinDescription[] ;
--- a/hardware/arduino/sam/cores/arduino/wiring_digital.c
+++ b/hardware/arduino/sam/cores/arduino/wiring_digital.c
@@ -29,6 +29,13 @@ extern void pinMode( uint32_t ulPin, uint32_t ulMode )
         return ;
     }
 
+#if defined __SAM3X8E__ || defined __SAM3X8H__
+  if(g_APinDescription[ulPin].ulPinType != NO_ADC && adc_get_channel_status(ADC, g_APinDescription[ulPin].ulADCChannelNumber))
+    {
+      adc_disable_channel( ADC, g_APinDescription[ulPin].ulADCChannelNumber);
+    }
+#endif
+
 	switch ( ulMode )
     {
         case INPUT:
--- a/hardware/arduino/sam/cores/arduino/wiring_digital.c
+++ b/hardware/arduino/sam/cores/arduino/wiring_digital.c
@@ -29,12 +29,19 @@ extern void pinMode( uint32_t ulPin, uint32_t ulMode )
         return ;
     }
 
-#if defined __SAM3X8E__ || defined __SAM3X8H__
-  if(g_APinDescription[ulPin].ulPinType != NO_ADC && adc_get_channel_status(ADC, g_APinDescription[ulPin].ulADCChannelNumber))
+  if ((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_ANALOG)
     {
       adc_disable_channel( ADC, g_APinDescription[ulPin].ulADCChannelNumber);
     }
-#endif
+
+  if ((g_pinStatus[ulPin] & 0xF) < PIN_STATUS_DIGITAL_OUTPUT && g_pinStatus[ulPin] != 0)
+    {
+      // return if already configured in the right way
+      if (((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_DIGITAL_INPUT && ulMode == INPUT) ||
+          ((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_DIGITAL_INPUT_PULLUP && ulMode == INPUT_PULLUP) ||
+          ((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_DIGITAL_OUTPUT && ulMode == OUTPUT))
+      return;
+    }
 
 	switch ( ulMode )
     {
@@ -46,6 +53,7 @@ extern void pinMode( uint32_t ulPin, uint32_t ulMode )
             	PIO_INPUT,
             	g_APinDescription[ulPin].ulPin,
             	0 ) ;
+            g_pinStatus[ulPin] = (g_pinStatus[ulPin] & 0xF0) | PIN_STATUS_DIGITAL_INPUT;
         break ;
 
         case INPUT_PULLUP:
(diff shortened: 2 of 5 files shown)


facchinm:
It's important to notice that every core based on Due should add the `g_pinStatus` array to its `variant.cpp`

facchinm:
BTW, the `400%` performance drop can be solved completely if `pinMode(INPUT)` and `pinMode(OUTPUT)` are performed outside `loop()` (which is a best practice in case of different static pins)

facchinm:
@HanaJin @trickedj @tamul @DavidEGrayson @JiapengLi  @billroy @3esmit , could you please test the PR? Could it be a good solution for your issues?

3esmit:
Test is not possible, Arduino SAM Boards module not found in the pull request file, there is no 'sam' folder inside 'hardware/arduino' in pull_requests/arduino-PR-3524-BUILD-361-windows.zip 

Download Arduino SAM Boards (1.6.4) files from Board Manager and test it anyway?

3esmit:
But looking at the code changed, one line is exactly what I did to fix it. But in #3310 @DavidEGrayson realised that is not a "default output state" problem. The problem is that if you send digitalWrite(LOW) before pinMode(OUTPUT) arduino will not output LOW, will output the initial output state, that by default is HIGH.

DavidEGrayson:
I'm not sure what @3esmit is talking about in the last comment.  This pull request runs the following code in pinMode(x, OUTPUT) (in wiring_digital.c) to determine if the output should be high or low:

~~~~
g_pinStatus[ulPin] = ((g_pinStatus[ulPin] & 0xF0) >> 4 ? PIO_OUTPUT_1 : PIO_OUTPUT_0)
~~~~

That's a step in the right direction.  I haven't looked carefully at the rest of the code or tested any of it.

hardware/arduino/sam/cores/arduino/wiring_digital.c
```diff
@@ -49,15 +64,18 @@ extern void pinMode( uint32_t ulPin, uint32_t ulMode )
...
             	g_APinDescription[ulPin].ulPin,
             	PIO_PULLUP ) ;
+            g_pinStatus[ulPin] = (g_pinStatus[ulPin] & 0xF0) | PIN_STATUS_DIGITAL_INPUT_PULLUP;
         break ;
 
         case OUTPUT:
             PIO_Configure(
             	g_APinDescription[ulPin].pPort,
-            	PIO_OUTPUT_1,
+              g_pinStatus[ulPin] = ((g_pinStatus[ulPin] & 0xF0) >> 4 ? PIO_OUTPUT_1 : PIO_OUTPUT_0),
```
DavidEGrayson: There is no need to set g_pinStatus if you are going to do it again on the next line.  There is no need to do `& 0xF0` because those bits you are masking off would get shifted anyway.

3esmit: I didn't checked all diffs, I'm not that deep into arduino code, but your comments makes sense. Thanks for the note.

facchinm: @DavidEGrayson , I'm aware there's no need to mask the variable but all the patch is explicitly masked so there's no need to remove it only here (I bet it will be removed by the compiler anyway, but I'll check later). If you wish something fancier it could be converted to use bit fields, but there won't be any speed improvement

facchinm:
Sorry for the inconvenience guys, the board manager is a great beast but, being the Due core still inside the IDE, it takes extra care to generate a custom core based on this PR (like we already do with the Zero). More news later today!

hardware/arduino/sam/cores/arduino/wiring_digital.c
facchinm: You also spotted a leftover that made its way into the PR , thank you :smile: . Fixed with 5a37e9478

hardware/arduino/sam/cores/arduino/Arduino.h
```diff
@@ -155,6 +155,16 @@ typedef enum _ETCChannel
...
 
+#define PIN_STATUS_DIGITAL_INPUT_PULLUP  (0x01)
+#define PIN_STATUS_DIGITAL_INPUT         (0x02)
+#define PIN_STATUS_DIGITAL_OUTPUT        (0x03)
+#define PIN_STATUS_ANALOG                (0x04)
+#define PIN_STATUS_PWM                   (0x05)
+#define PIN_STATUS_TIMER                 (0x06)
+#define PIN_STATUS_SERIAL                (0x07)
+#define PIN_STATUS_DW_LOW                (0x10)
+#define PIN_STATUS_DW_HIGH               (0x11)
```
aethaniel: shouldn't these definitions be put in a typedef'ed enum?

hardware/arduino/sam/cores/arduino/wiring_digital.c
```diff
@@ -90,6 +114,14 @@ extern void digitalWrite( uint32_t ulPin, uint32_t ulVal )
 
 extern int digitalRead( uint32_t ulPin )
 {
+  if ((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_DIGITAL_OUTPUT) {
+    return (g_pinStatus[ulPin] & 0xF0) >> 4;
+  }
+
+  if ((g_pinStatus[ulPin] & 0xF) == PIN_STATUS_ANALOG) {
+    pinMode(ulPin, INPUT);
```
aethaniel: what is the meaning of calling pinMode here?
I may have missed something but I understood the PIN_STATUS_ANALOG means the pin is under control of ADC or DAC. So, calling pinMode() which active the clock on PIOx and putting the PIOx pin as INPUT will run useless code.

hardware/arduino/sam/cores/arduino/Arduino.h
facchinm: Being a mashup between an enum and a mask (to save some bytes) I'd prefer to keep the hardcoded values, but it's a matter of choice :smile:

hardware/arduino/sam/cores/arduino/wiring_digital.c
facchinm: Indeed, the main goal of this patchset was solving the incompatibilities between pins behaviour on AVR and on SAM core, losing as few performances as we can.
For sure this piece of code can be optimized but, as you said, it's really uncommon to use `pinMode` during the loop, so we can be quite sure that it won't affect most people.

DenDanny:
@facchinm et al.

I was wondering if you could give this beginner a push in the right direction..

I want to switch an Analogue pin between INPUT and OUTPUT in order to switch a LED between a "receiving" and emitting state (LedComm) - which since 1.5.4 don't work no more. Also, i want to use multiple analogue pins (each with an LED).

I can sort of follow (in theory) what's been going on, but practically I have no idea where/how I can these fixes and patches you (and others) seem to have created. Where/what can I download to obtain these fixes?

Regards,
Danny
PS. see my example (that doesn't work) on: https://forum.arduino.cc/index.php?topic=344560.0

facchinm:
Hi @DenDanny, the PR builder for Due core is still far from being complete, so in the meantime, if you wish to test this PR, I'd suggest to extract [this] (https://mega.nz/#!R5JzSaYb!oRss5QcoL1Jv6pvULNzdcNulSQrmnDhtPKPo2F4qve4) file into `sketchbook/hardware` folder. A new menu entry (called `Arduino ARM (32-bits) Boards PR3524`) should appear the next time you restart the IDE, and it contains all the patches.

@HanaJin @trickedj @tamul @DavidEGrayson @JiapengLi @billroy @3esmit sorry for the long delay, if you could test using this quick and dirty method it would be a great help for inclusion. Sorry again.

HanaJin:
Hi @facchinm , I tempted to try the patches, but I got this error when I tried to upload the sketch. Please help.

capture

facchinm:
Hi @HanaJin , the packet contains only the core, not the tools, so you need to download the SAM core first using Board manager and then the IDE will know where to find the compiler.
Sorry for not explaining this in the previous post and thanks for testing this patch!

HanaJin:
Hi @facchinm  sorry to bug you again. I've downloaded the core and tested the original Due which compils fine, but with the patches, it still does not compile.
capture

Note that I checked the .c.o file was right in the build folder.

facchinm:
Ok, found, the source of your problems is the IDE and more specifically `platform.keys.rewrite.txt` file. The fastest (and safest) method to fix compilation is to use nightly build (https://www.arduino.cc/download.php?f=/arduino-nightly-windows.zip) :wink:

facchinm:
To test the PR completely, [here] (https://gist.github.com/facchinm/8ab3b1e43d17ee47edc7]) you can find a board-to-board test with wiring instructions on top.

HanaJin:
Yes, the patches fix the issue I described in #2455. Now I get similar results. Hope this helps.

Issue #3546, opened 2015-07-18
Add variadic print(String...) function


NicoHood:
Just found this:
http://stackoverflow.com/questions/10044449/a-function-with-variable-number-of-arguments-with-known-types-the-c11-way

Add this to Print.h:
```cpp
    template <typename ...TailType>
    size_t println(const String &head, TailType&&... tail)
    {
        size_t r = 0;
        r+=println(head);
        r+=println((tail)...);
    }
```

Now you can do stuff like this:
Serial.println("Hey", "this", "is", "a", "Test");

Every word is printed in a new line. You can also add this for Serial.print (without newline).
Maybe this can be optimized to only print the last input with a new line. An overload with two arguments is needed here I guess.

Not sure if its perfect c++, but this would give us the option to print infinite strings at once.
Maybe one can generalize this even more to also print numbers (without formating of course, just DEC).

Any comments on that?

NicoHood:
Seems I got it working pretty good:
```cpp
    // Addition by NicoHood
    template <typename... Rest> 
    inline size_t println(const Rest&... rest) {
        // A Wrapper is needed here in order to not call the
        // Formatted functions if the input was the integers
        // If you input only two integers the formatted
        // Functions have higher priority
        return printlnWrapper(rest...);
    }

    template <typename First> 
    inline size_t printlnWrapper(const First& first) {
        return println(first);
    }

    template <typename First, typename... Rest> 
    inline size_t printlnWrapper(const First& first, const Rest&... rest) {
        size_t r = 0;
        r+=print(first);
        r+=printlnWrapper(rest...); // recursive call using pack expansion syntax
        return r;
    }
```

Code to test:
```cpp
  Serial.println("1",  "2", 6, "3"," Word " , "4", '!', "5");
  Serial.println();
  Serial.println(44, 44, 44);
  Serial.println();
  Serial.println(44, HEX);
  Serial.println();
```

Chris--A:
At first glance, this seems like a great addition and I'm sure there are some further improvements we can look at. However this code will cause the current API to become C++11 only. Even if the code is wrapped in define guards, the reference will become separated unless a C++98 compatible version is provided using variadic functions.

Chris--A:
I do have an idea though! I give it a test when I'm home in a few hours.

q2dg:
Ooh, yeahh!! +1!!

NicoHood:
[...]

PR #3551, opened 2015-07-19
Add variadic functions

NicoHood:
See https://github.com/arduino/Arduino/issues/3546

Example code:
```cpp
  Serial.println("1",  "2", 6, "3"," Word " , "4", '!', "5");
  Serial.println();
  Serial.println(44, 44, 44);
  Serial.println();
  Serial.println(44, HEX);
  Serial.println();

// this produces the same flash size:
  // Serial.println("1", "2", "3", "4", "5");
  Serial.println("1");
  Serial.println("2");
  Serial.println("3");
  Serial.println("4");
  Serial.println("5");
```

This may also be usable with pinMode(2,6,4,9, INPUT);

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.h
+++ b/hardware/arduino/avr/cores/arduino/Print.h
@@ -79,6 +79,52 @@ class Print
     size_t println(double, int = 2);
     size_t println(const Printable&);
     size_t println(void);
+
+    // Addition by NicoHood
+    template <typename First, typename Second, typename Third, typename... Rest> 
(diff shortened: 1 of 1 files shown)


Chris--A:
```arduino
//If you input only two integers the formatted
// Functions have higher priority
```

There is a lot of ambiguity as the second parameter means two different things depending on the total number of items, rather than the type of inputs. This should be modified to allow standard overload selection/deduction semantics.

One method of fixing this requires the formatting elements to be enums which will remove all ambiguity.

PR #3589, opened 2015-07-26
Add isFlushed() to HardwareSerial (avr) and UARTClass (sam)

thatcadguy:
I wanted a way to see if all data had been transmitted without calling flush(), which is blocking. I took conditionals from write() to see if the port is ready to send more data (i.e. finished sending everything else). Please change/correct as necessary, re-name the function, etc.

In relation to http://forum.arduino.cc/index.php?topic=337900.0


arduinium:
I need this too. Please merge :)

PaulStoffregen:
Does this truly indicate with all bits are transmitted, as flush() does, or will it indicate the flushed condition while the UART is still shifting out the last byte?

thatcadguy:
Hi @PaulStoffregen, I basically took the code from flush() for both the SAM and AVR codebases to make this function. It's my understanding that the TXRDY (for the SAM) is true when the TX line is inactive; between that condition and the checking the buffer position should indicate that everything in the buffer has been sent and that the TX line is finished working. Here are the flush() codes for comparison:

For SAM (from UARTClass.cpp):
```
void UARTClass::flush( void )
{
  while (_tx_buffer->_iHead != _tx_buffer->_iTail); //wait for transmit data to be sent
  // Wait for transmission to complete
  while ((_pUart->UART_SR & UART_SR_TXRDY) != UART_SR_TXRDY)
   ;
}
```

For AVR, I stole the conditional from the beginning of write() in HardwareSerial.cpp actually:
```
  // If the buffer and the data register is empty, just write the byte
  // to the data register and be done. This shortcut helps
  // significantly improve the effective datarate at high (>
  // 500kbit/s) bitrates, where interrupt overhead becomes a slowdown.
  if (_tx_buffer_head == _tx_buffer_tail && bit_is_set(*_ucsra, UDRE0)) {
    *_udr = c;
    sbi(*_ucsra, TXC0);
    return 1;
  }
```

I actually had a chance to test this out on the Due today and it seemed to work correctly, i.e. while inactive TXRDY was true and availableForWrite() was buffer-1 (127 by default).

matthijskooijman:
The AVR code is incomplete, it should also look at the (IIRC) TXC. UDR only indicates that the FIFO byte is empty, there is also a shift register that can still contain a byte currently transmitted. Furthermore, the commit structure isn't ideal yet - one commit per file without a meaningful commit message. This should be one commit per logical change (so perhaps one commit in total, or one for AVR and one for SAM) with a descriptive commit message.

thatcadguy:
Standby, engaging in hand-to-hand combat with Git...going to fix the AVR part and squash the commits.

thatcadguy:
Had to re-do everything, new pull request is #3737

matthijskooijman:
For future reference, you can always force-push to a branch, completely replacing the contents, which will then just update the pullrequest.

PR #3606, opened 2015-07-31
Fix links to 3rd party hardware specification.

reillyeon:
The number of hyphens in this URL has apparently changed.


ffissore:
Thanks!

PR #3620, opened 2015-08-04
File change detection feature

pvrs12:
Hi, I implemented a feature in the Processing project to detect if a file in the sketch was changed externally (since the "Use External Editor" option was removed) and decided to see if I could implement it for the Arduino editor as well. 

This PR should implement the Change Detector as it is in Processing with a few tweaks to support the differences in the Arduino editor. 

The changes made to Sketch, Sketchcode, SketchData, and Editor are required to support the Change Detector.

### DIFF
--- a/app/src/processing/app/ChangeDetector.java
+++ b/app/src/processing/app/ChangeDetector.java
@@ -26,7 +26,7 @@ public class ChangeDetector implements WindowFocusListener {
   private boolean skip = false;
 
 
-  public ChangeDetector(Editor editor) {
+  public ChangeDetector(Editor editor) {    
     this.sketch = editor.sketch;
     this.editor = editor;
   }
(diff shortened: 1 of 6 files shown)


Chris--A:
Shouldn't the watcher be enabled automatically when external editing is enabled, rather than a toggle for each function.

Although I guess this is similar to Notepad++ behavior when you return focus to the window after changing files externally.

pvrs12:
@Chris--A Yes, the way it works right now is if external editing is enabled the editor blocks input and reloads the sketch whenever compiled. 

This watcher was originally written for Processing which removed the external editor option a while ago. Since I'd already written the code for that I figured I'd port it over for Arduino's editor too. 

I think ideally this should be able to replace the existing external editor option (but that isn't my call)

Chris--A:
Ah, I see: A replacement, rather than addition. It would be better like that I think. 

Some way of letting the user know it has been reloaded (or optionally ask) would be good. I like the fact that when I change git branches, notepad++ will ask if I want to reload the file as it was changed externally.

JASON33:
Guys Windows 10. Know.I have.upload have not checked wat com.should it
be.or2.0@3.0usb. Port.asusx551mav

ffissore:
As @Chris--A notices, a "file has changed" notification should be implemented as well. Such notification is a change of an existing working behaviour. We are not going to remove "use external editor". @pvrs12 thank you for your effort but I'm not going to merge this: we are happy enough with the current implementation

pvrs12:
Thanks for considering it, I totally understand your decision. It should also be noted there is a notification in this implementation at [ChangeDetector:147](https://github.com/arduino/Arduino/pull/3620/files#diff-85eedbc341e4cea9f84e4ebeb5be0476R147)

If it is decided that you want this in the future (and it can't be merged) let me know and I'll tweak it if necessary :D

PR #3624, opened 2015-08-04
Fixed slowdown with repeated CDC bool operator test

majenkotech:
The `delay(10)` in the `bool` operator of the USB CDC/ACM Serial operator was being imposed on every single call to the operator when it is not needed.  Instead this `delay(10)` should *only* be called when the line state transitions from ==0 to >0 (i.e., as the line enters the active state from the inactive state), otherwise repeated calls to (eg) `if (Serial)` cause the entire sketch to slow down to a crawl.

By only imposing the delay when the line changes state it avoids that slowdown whilst at the same time giving the system a chance to complete the port opening operation before data is allowed to be sent.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/CDC.cpp
+++ b/hardware/arduino/avr/cores/arduino/CDC.cpp
@@ -28,9 +28,10 @@ typedef struct
 	u8 	bParityType;
 	u8 	bDataBits;
 	u8	lineState;
+    u8  prevLineState;
 } LineInfo;
 
-static volatile LineInfo _usbLineInfo = { 57600, 0x00, 0x00, 0x00, 0x00 };
+static volatile LineInfo _usbLineInfo = { 57600, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
 #define WEAK __attribute__ ((weak))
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Note that #3437 has a fix for the same issue. It works by saving the timestamp of a linestate change, which can prevent unneeded additional delays, but also needs 3 additional bytes of RAM.

hardware/arduino/avr/cores/arduino/CDC.cpp
```diff
@@ -196,13 +197,19 @@ size_t Serial_::write(const uint8_t *buffer, size_t size)
...
+// where the port is configured (lineState != 0) but not quite opened.  This
+// only happens when the line state transitions from ==0 to >0 otherwise you
+// get a nasty delay when one isn't needed.
 Serial_::operator bool() {
-	bool result = false;
-	if (_usbLineInfo.lineState > 0) 
-		result = true;
-	delay(10);
-	return result;
+    bool res = false;
```
NicoHood: I'd not rename it to res. Res sounds weird to me. Please keep result. The rest seems like a good idea to me.

PR #3625, opened 2015-08-04
Update notifications

ffissore:
This PR introduces an automatic check for libraries and cores updates. If some cores and/or libs can be updated, it will show a popup. See screenshot

update-notification


VaclavSal:
Great, nice work guys. 
I am switching to Linux and have little time to test this  release ( project is  due ( sic!)) and would appreciate if some "tester" would post comment on this: 
Does the editor still indicate matching brackets / code block  " {....}" only by putting blue frame around them or is the ENTIRE code block HIGHLIGHTED as is was prior ( 1.6.5?) to this annoying change ?

PR #3647, opened 2015-08-09
Do not expose internal SD.root to sketches with SD.open("/")

PaulStoffregen:
Discussion here:
https://groups.google.com/a/arduino.cc/d/msg/developers/jWsi7A_g-Mk/Q2sFpzDDBQAJ


cmaglie:
Moved to: https://github.com/arduino-libraries/SD/pull/40

PR #3648, opened 2015-08-09
Rename howto.txt to readme.md

Ramoonus:


matthijskooijman:
I wonder if it would be better to instead move the instructions into the repository, or perhaps the github wiki. @cmaglie, @ffissore, any suggestion for this?

ffissore:
We have ported the old instructions in the wiki https://github.com/arduino/Arduino/wiki/Building-Arduino but they so old to be plain wrong. I'll ask @agdl to review them. Merging and changing url in file

Issue #3614, opened 2015-08-03
Print not aborting when Write(char) returns Error


stickbreaker:
the `Print::write(const uint8_t *buffer, size_t size)` and `Print::print(const __FlashStringHelper *ifsh)` continue calling `write(char)` after `write(char)` has returned a 0 (zero) to mark a write failure. This creates a unexpected result when a block of characters is printed.  
  I implemented cts/rts handshaking in `HardwareSerial`, if !cts and the tx buffer is full and a timeout has occured, the `HardwareSerial::write(char)` returns 0(zero) to mark the failure. The Current `Print` object continues sending until the buffer has been send to `write(char)` if buffer space becomes free or cts clears, the `write()` failed characters are lost, but the rest of the buffer is sent.  Since `Print` functions returns the number of characters sent, the expectation is that if len!=sent, the last (len-sent) characters were not sent.  This expectation is incorrect.  With the current `Print` library if len!=sent, there is no way to identify which character were not sent!
  Included is the simple fix for this problem.
~~~C++
// from Arduino\hardware\avr\cores\arduino\Print.cpp 

// Public Methods //////////////////////////////////////////////////////////////

/* default implementation: may be overridden */
size_t Print::write(const uint8_t *buffer, size_t size)
{
  size_t n = 0;
  while (size--) {
// abort on write() failure, Chuck Todd
    if(write(*buffer++))n++; 
    else break;  
 //   n += write(*buffer++); original code 1.6.5
  }
  return n;
}

size_t Print::print(const __FlashStringHelper *ifsh)
{
  PGM_P p = reinterpret_cast<PGM_P>(ifsh);
  size_t n = 0;
  while (1) {
    unsigned char c = pgm_read_byte(p++);
    if (c == 0) break;
// abort on write() failure, Chuck Todd
    if(write(c)) n++; 
    else break;    
//   n += write(c); original code 1.6.5
  }
  return n;
}
~~~

Chuck Todd

Chris--A:
That looks like a nice simple fix. However is it compatible with what is already happening.

I can think of one case: `EthernetServer.println()` will write to all clients. If one client fails to write, all clients will only get a half finished message.

Maybe the class needs a toggle to 'turn on' this behavior if the return value is needed to find where the stream stopped. 

But the return value really only has meaning when using `write()` directly, the formatting print() functions return value has very little resemblance to what was passed in (might not even be text before formatting).

matthijskooijman:
[...]

PR #3651, opened 2015-08-09
Print not Aborting on Write() failure

stickbreaker:
The current 1.6.5 Print library report the total count of characters send during operations.  The assumption is that this count is of contiguous characters.  If you printed a buffer of 45 characters and the print command returned 44 as the number of characters sent, the assumption is that the last character was not sent.  But, this is not a valid assumption.  The two low level functions through which all of the helper print funtions work (`size_t Print::write(const uint8_t *buffer, size_t size)` and `size_t Print::print(const __FlashStringHelper *ifsh)`) do not exit when a write failure is detected.  They just attempt to write the next character from the input buffer.  This renders any positional transmit information useless. 
  I propose to change the behavior of these two routines.  When a write error is encountered these function return with the count of sent characters.  If the calling procedure wants to retry sending then ~buffer[sentCount] is the next character needing to be send.

see arduino/Arduino issue #3614

Chuck Todd

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -35,7 +35,7 @@ size_t Print::write(const uint8_t *buffer, size_t size)
 {
   size_t n = 0;
   while (size--) {
-    if(write(*buffer++) n++;
+    if(write(*buffer++)) n++;
     else break;
   }
   return n;
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -17,6 +17,7 @@
  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  
  Modified 23 November 2006 by David A. Mellis
+ Modified 03 August 2015 by Chuck Todd
  */
 
 #include <stdlib.h>
@@ -34,7 +35,8 @@ size_t Print::write(const uint8_t *buffer, size_t size)
 {
   size_t n = 0;
   while (size--) {
-    n += write(*buffer++);
+    if(write(*buffer++) n++;
+    else break;
   }
   return n;
 }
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @stickbreaker , your proposal is interesting but the code won't work with cdc serials (like Leonardo/Micro) as `write` could return more than `1`.

stickbreaker:
@facchinm , this fix is for the ~Print::~ object, all of the higher level calls to ~Print::print(types)~ or~Print::println(types)~ funnel thru the  ~size_t Print::write(const uint8_t *buffer, size_t size)~

What are _cdc serials_? reference please :) I'll check it out.

chuck

matthijskooijman:
@stickbreaker, the CDC serial (virtual serial over USB) for AVR is here: https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/CDC.cpp#L164-L191 (there is also one for SAM).

@facchinm, I'm not sure what you mean with `write()` returning more than one? AFAICS, `write(char)` should always return 0 or 1, and `write(uint8_t*, size_t)` should return 0 up to size. Both of the calls in this PR use the `write(char)` version. Note that it is not technically impossible for `write(char)` to return more than 1, but I'd consider this a bug (`Ethernet` currently does this, see #1699).

Note that this PR modifies the virtual `write(uint8_t*, size_t)` in `Print`, which is just the default implementation for subclasses that did not implement this virtual method themselves. This means that subclasses can still override this method and provide an implementation for which the assumptions about the return value do not hold. This would probably need to be checked against the known subclasses, and these assumptions should be written down somewhere (reference page for `Print` on arduino.cc seems to make sense).

You might want to modify your commit message to say "Closes #3614" in the last line, so the issue will be [automatically closed](https://help.github.com/articles/closing-issues-via-commit-messages/) when merging the PR.

facchinm:
Sorry @stickbreaker , your solution is ok, I was blinded by the difference between `n +=` and `n++` (trusting `write()` to return 1) and while searching for a counter-example I totally forgot the method signature :smile: 
Anyway, you missed a parentheses on `if(write(*buffer++) n++;`, so I'll merge after the fix

stickbreaker:
@facchinm , You mean every little punctuation must be Correct !  :blush: 
sorry about that, I thought I could copy 20 character without error, but I was wrong!

Chuck.

facchinm:
Pushed as single commit as 25d81c98e

PR #3673, opened 2015-08-14
Added missing break.

onovy:


matthijskooijman:
The break change you made looks good. AFAICS it wasn't broken without it, it just continued to iterate over a loop when the only valid result in the array was already found, so with the break it should be slightly more efficient.

I don't think the comment change you made is correct, though. The comment "leave timer 0 for last" applies to the original array contents, where multiple timers were available for tone playback. All but 1 timer have been commented out, so there isn't really any leaving to last, so the comment is confusing, but I don't think your change actually improves this.

Could you change your PR to remove the comment change? You can `git commit --amend` to change your commit locally and then force push it to the same branch to update this PR. While you're at it, perhaps you could improve the commit message a bit too. The trailing dot in the first summary line should be removed and it might be good to mention that this change applies to the Tone library.

Thanks for your contribution!

onovy:
all done, thanks.

matthijskooijman:
Looks good to merge to me.

Issue #3675, opened 2015-08-14
build.extra_flags -U undef option not working


NicoHood:
I want to deactivate the USB core of Arduino. Therefore USBCON needs to be undefined. Passing options with -D works, -U not.

2nd issue related to this: the USB-Core should not rely on such a USBCON definition, there should be a wrapper

cc @facchinm

matthijskooijman:
@NicoHood AFAIU, -U only undefines things previously defined with -D on the same commandline, it does not prevent #defining them in the source. To support what you need, you would need an #ifnded NO_USB (or something better named) in the source.

NicoHood:
Okay so I changed the issue.

We could:
* integrate a #ifndef NO_USB_CORE as you said
* or we could make the whole USB-Core a library which can be included.

2nd option would save us any compiler flags and new boards.txt submenus.
cc @facchinm

matthijskooijman:
Problem with the second option is that any existing sketches will break, since they won't have the right include...

NicoHood:
See the PR I referenced. This would solve it properly.

The only question is if we should add a workaround or update the bootloader.

PR #3679, opened 2015-08-15
Made USP-API included flexible

NicoHood:
Resolves https://github.com/arduino/Arduino/issues/3675

This also saves us tons of flash and ram for the Leonardo if the CDC Serial/HID is not used.
Example: 3724 flash, 135 ram (with Serial.begin(0)) vs 578 flash, 9 ram without
```cpp
void setup() { Serial.begin(0); }
void loop() {}
```

The "downside" is that you need to reset the Leonardo if no USB-Core was compiled. But with the new Serial touch waiting this is not a problem. Maybe an additional not like "If your Leonardo board does not show up, reset it NOW." could easily solve this.

This PR would save us a lot of flash + ram AND USB caused problems (like ISRs).

There is also a problem that if no USB-Core is used, the Led blink example wont work correctly. This is caused by a bad coded bootloader which does not fully reset the MCU via watchdog and leaves the USB-Clock configured. You need a [workaround](https://github.com/NicoHood/HoodLoader2/tree/f0d50015890a50cfe154b5ec4eda7fbeda12c1e6/avr/variants/leonardo_NoUSB) to fix this (not included in this PR yet). HoodLoader2 already includes this (if you want to see it in action, v2.0.5 available for 32u4 as well!).

The workaround is easy to add, but you might want to think about updating the bootloader now.
Updating the bootloader can fix:
* This PR
* Magic Bootkey https://github.com/arduino/Arduino/issues/2474
* Change the reset behavior [similar to this](https://learn.sparkfun.com/tutorials/pro-micro--fio-v3-hookup-guide/all#ts-reset) (but with different implementation)
* Maybe also include the Fast USB-Serial if you like to from [HoodLoader2](https://github.com/NicoHood/HoodLoader2).


NicoHood:
May anyone review this PR?
With https://github.com/arduino/arduino-builder/issues/15 one could dynamically disable the USB-Core if needed. This keeps the USB-Core for beginners but can be still disabled also for leonardo boards. (The problem with the unfixed USB Clock bootloader is still there, but professional users should be aware of that)

PR #3689, opened 2015-08-16
Provide for support for ZeroConf/Bonjour

lathoub:
Provide for support for ZeroConf/Bonjour, by allowing derived classes
access to _sock and _remaining. See
https://github.com/lathoub/EthernetBonjour

### DIFF
--- a/libraries/Ethernet/src/EthernetUdp.h
+++ b/libraries/Ethernet/src/EthernetUdp.h
@@ -43,11 +43,13 @@
 
 class EthernetUDP : public UDP {
 private:
-  uint8_t _sock;  // socket ID for Wiz5100
   uint16_t _port; // local port to listen on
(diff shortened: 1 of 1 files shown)


cmaglie:
I see that you use _sock only in `beginMulti` method:

https://github.com/lathoub/EthernetBonjour/blob/master/EthernetBonjour.cpp#L167

but this method has been recently merged upstream and named `beginMulticast`:

https://github.com/arduino/Arduino/commit/e5805926d0f02c26522de7c8c9c4e21ae589bd80

can you use this method so we can avoid to expose the `_sock` field?

lathoub:
e580592 looks good from my perspective, avoids exposing both _sock and _remaining. This is great! I shall close this issue, delete my pull request and await 1.6.6.

Issue #2800, opened 2015-03-20
.a linkage for libraries via library.properties


NicoHood:
It would give developers a lot of more possibilities if libraries could be linked via an .a file.
So only the needed .cpp files get compiled like the core or the .ino file directory.

It would also break existing libraries. So my suggestion is just to only compile through .a if a special property was set.

You could add this perfect into such a file with an entry "linkage"
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/libraries/SoftwareSerial/library.properties

Also this link sais if i put the library into an "src/" folder it gets linked in the .ino path. This option is not working anymore. Also the import function is under "sketch" not "tools".
https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification#source-code
it would be cool to make it work like this though.

matthijskooijman:
I like this idea of opting in to linking through a .a file through library.properties. It might even be a good idea if _some_ source files can be linked through .a files and some directly included in the link, for maximum flexibility. The reason for wanting to go through a .a file, is so you can leave out a source file from the link when it is unused, _including_ any ISRs (which are never removed by `gc-sections` once they're included, due to the `used` attribute).

> Also this link sais if i put the library into an "src/" folder it gets linked in the .ino path. This option is not working anymore. Also the import function is under "sketch" not "tools".

I'm not sure what you are saying here, but the "also" suggests they are separate issues. Please don't complicate discussion by chunking together unrelated issues, better open multiple issues instead (or, for things like what option goes where, collect opinions on the developer mailing list first).

NicoHood:
Has `* Upgraded toolchains: statically compiled, ship with Link Time Optimization plugin (disabled by default)` something to do with this issue? Any chance to see this in the future? With the new library system this would really make sense to include the .a linkage.

And I also like the idea of mathijs to link different files "normal" or via .a file.

NicoHood:
Any chance to see this soon? It would fix a lot of issues with 2 of my libraries and newer libs can be optimized better.

I'd do it myself but I dont know where to start with the whole Arduino IDE source and I also dont have enough knowledge about the linking process.

NicoHood:
I am currently trying to implement this feature myself.
After some time of research and the help of @matthijskooijman I finally got a working version that compiles the library via .a file instead. And it seems to work.

One quick question:
[...]

Issue #3649, opened 2015-08-09
[1.6.6 Pluggable HID] Compiling issues for custom HID libraries


NicoHood:
This PR added Pluggable HID with IDE 1.6.6
https://github.com/arduino/Arduino/pull/3304

I started developing some custom HID devices.
See this commit as reference: https://github.com/NicoHood/HID/tree/4df1f30427730c43216336b94673f66b546fc6d4/src

It turns out that because of static linkage all files are compiled. This means Mouse, Gamepad, System and Consumer files are all compiled **and linked**. Now try to open the Mouse example. It wont work because the singletons (Consumer, Gampad, System instances) add their descriptor in the constructor. I dont know why this still happens, **even if you do not include the other files** they will cause trouble.

First I did not notice this, because not all HID devices cause problems when used together. The mouse button was working but the moving did not. The problem is a mix of incompatible USB devices, cause by accidentialy compiled files.

So I am not sure if this is expected to be this way, but this feature request could possibly help us:
https://github.com/arduino/Arduino/issues/2800

The solution for now is to delete the other .h/cpp files (if you want to try it yourself with the commit above). The long time solution would be to create 4 different libraries. This is bad for the boards manager. It'd be better if you can add a whole HID package instead of 4 (and later even more).

@facchinm

facchinm:
Hi Nico, the library compilation method (all .cpp files get compiled) plus the linker strategy (all .o file become part of the .a) plus the pluggable method (using singletons) do, in fact, cause your issue. 
I can see two solutions: 

1 -  if you drop your cpp files (which are almost all empty), declare the singleton in the .h without `extern` and include only what you need, no extra descriptor should be included

2 - standalone libraries; I can't see the problem with the library manager, quite the opposite in fact. From the dev point of view, you could have a development repo with all the libraries together and some handy scripts using `git-filterbranch` that, once you want to release, will take care to extract the history of every standalone library, tag it and push it

3 - (this in not a real solution :smile: ) you can also drop the pluggable-HID feature and base everything on pluggable-USB (and give your users infos on how to enable/disable the library features), somehow like [CompleteHID] (https://github.com/facchinm/Arduino/tree/8ebc5d0347ba3c5e3e672d4d2144c947cb6c03fb/libraries/CompleteHID) library. In this way you can have full control over what is included and what is not, better handle coexistence between different functions (which you can't do with the standalone libraries) and reduce the memory footprint.

NicoHood:
[...]

PR #3697, opened 2015-08-18
Optional .a linkage for libraries

NicoHood:
Fixes:
https://github.com/arduino/Arduino/issues/2800
https://github.com/arduino/Arduino/issues/3649
And adds the possibility to create more enhanced libraries with smaller flash size. **This feature is optional and won't break any existing library.**

All you need to do is add a `alinkage=true` to your library.properties file. By default alinkage is off, to not break older libraries with ISRs. A linkage is only possible for newer libraries with a /src folder.

Most of the code was adapted from here, so it should work fine, I did not reinvent the wheel here.
https://github.com/arduino/Arduino/blob/master/arduino-core/src/processing/app/debug/Compiler.java#L1130-L1182

The code now moved in its own function (2nd commit) to reduce code duplication.

I am still not sure if we need `recursiveCompileFilesInFolder()` or if we can use `compileFiles()` with `true` as paramter. If desired I can add a new `recursiveCompileFilesInFolder()` which returns a List of Files instead of void, if that helps in any case. What was the thought behind this? For this PR it is not critical since .a linkage is optional and won't break anything, and the core compiling stays the same.

### DIFF
--- a/arduino-core/src/processing/app/debug/Compiler.java
+++ b/arduino-core/src/processing/app/debug/Compiler.java
@@ -1070,8 +1070,76 @@ public class Compiler implements MessageConsumer {
     if (lib.useRecursion()) {
       // libBuildFolder == {build.path}/LibName
       // libFolder      == {lib.path}/src
-      recursiveCompileFilesInFolder(libBuildFolder, libFolder, includeFolders);
-      
+
+      // Compile the library with .a linkage if a flag was set in library.properties
+      if(lib.alinkage()){
+
+          File afile = new File(libBuildFolder, lib.getName() + ".a");
+
+          createFolder(libBuildFolder);
+          List<File> libraryObjectFiles = compileFiles(libBuildFolder, libFolder, true, includeFolders);
+
+          // See if the .a file is already uptodate
+          if (afile.exists()) {
+            boolean changed = false;
+            for (File file : libraryObjectFiles) {
+              if (file.lastModified() > afile.lastModified()) {
+                changed = true;
+                break;
+              }
+            }
+
+            // If none of the object files is newer than the .a file, don't
+            // bother rebuilding the .a file. There is a small corner case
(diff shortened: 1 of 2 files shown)


matthijskooijman:
@ArduinoBot, Compile this please.

matthijskooijman:
@ArduinoBot, Build this please.

matthijskooijman:
@ArduinoBot build this please.

NicoHood:
No you need to write `@ArduinoBot, Build this please` without the dot

cmaglie:
@ArduinoBot build this please

cmaglie:
BTW the proposed patch looks good to me (y).

NicoHood:
The easiest thing would be to merge it ;)
I've been using this for quite a while now and it works perfectly with and without .a linkage.

matthijskooijman:
I haven't gotten around to looking at the patch in detail yet, should be able to do that in 1, max 2 weeks from now.

matthijskooijman:
Seems this patch should be re-done for the new arduino-builder, now. @ffissore, any plans to include this? @NicoHood, care to learn the Go language too? :-)

NicoHood:
If you tell me the files I need to have a look at after the change :S I dont care if its go, java or c.

But it should be easy for @ffissore to include this since he just wrote the compiler. So he knows where to search and how to add it best.

NicoHood:
I've looked through the sources now and beside I do not fully get the Go language there is not a single commentary. So I am pretty much helpless here. I linked the useful files above but I am currently unable to understand how the .a linkage was used here. I also dont know how I should compile the files  (the go files) anyways.

A bit feedback from @ffissore would be appreciated. Thanks.

ffissore:
Closing as I'm working on it. Tracking issue is https://github.com/arduino/arduino-builder/issues/10

Issue #613, opened 2012-11-15
Wire library TwoWire.end() needed by Firmata [imported]


cmaglie:
This is [Issue 613](http://code.google.com/p/arduino/issues/detail?id=613) moved from a Google Code project.
Added by 2011-08-30T13:51:00.000Z by [paul.sto...@gmail.com](http://code.google.com/u/102276914943260257800/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Enhancement, Priority-Medium, Component-Core

### Original description

**What change would like to see?**

TwoWire.end() function that disables I2C and returns both pins to normal operation.

**Why?**

So the pins can be reused

**Would this cause any incompatibilities with previous versions?  If so, how**
**can these be mitigated?**

should not cause any compatibility problems

cmaglie:
Fixed with #3724

sandeepmistry:
@cmaglie re-opening as the SAM core ``Wire.end()`` function hasn't been implemented yet.

cmaglie:
closed by #3894

PR #3730, opened 2015-08-26
Implement Wire.end() for SAM core

sandeepmistry:
For the SAM side of #613.

Also moved common reset and disable code in libsam ``TWI_ConfigureMaster``
and ``TWI_ConfigureSlave`` functions to new ``TWI_Disable`` function, which is
used in ``TwoWire::end()``

@cmaglie needs to be run and verified on a Due board (both Wire master and slave), as I don't have one to test on yet.


sandeepmistry:
@cmaglie I received a Due board to test with, unfortunately the patch doesn't work, the voltage on the pins were still high after ``end`` was called. I will take another look at it this week.

sandeepmistry:
I'll open another pull request with my updated changes for this shortly.

Issue #1477, opened 2013-06-20
I2C/Wire library: Timing issue between Master-receiver and Slave-transmitter


ghost:
This happens only with an Arduino as I2C Slave (connected to an Arduino as I2C Master).
If the Master requests a few byte with Wire.requestFrom(), and the Slave writes a few bytes in the onRequest handler, the Master sometimes gets 0 bytes.

Arduino version 1.0.5

This issue came in view with a slave at 8Mhz.
With two 16MHz Arduino boards, this problem still occurs at an average of 1 in 100 times.
With other boards it could happen only 1 in 1000 times, but also all the time.

A workaround seems to be a "delay(1)" between the "Wire.endTransmission()" and the "Wire.requestFrom()" in the Master. The Wire.endTransmission is part of writing the (virtual) registers address to the Slave, since I programmed the Slave to act like a sensor or memory chip.

After some testing I found that 6 micro seconds delay was enough.

I have made a test sketch and more explanation here, http://forum.arduino.cc/index.php?topic=172296.0

sandeepmistry:
This is reproducible with @PaulStoffregen's example sketches from http://forum.arduino.cc/index.php?topic=172296.msg1289288#msg1289288 and two Uno boards

sandeepmistry:
Fixed by https://github.com/arduino/Arduino/pull/3735

PR #66, opened 2012-03-07
Repeated Start fix I2C slave

helenarobotics:
See issue 848:

http://code.google.com/p/arduino/issues/detail?id=848&q=Wire

PR #3735, opened 2015-08-27
Release TWI bus right after stop or repeated start condition

sandeepmistry:
This resolves timing issues seen in #1477.

Merging #66 made timing errors from #1477 more frequent, as TWI bus was no longer released earlier (via ``twi_stop``). This change releases the TWI bus right after a stop or repeated start condition is received.

Example sketches from https://github.com/arduino/Arduino/issues/1477#issuecomment-135459738 can be used to reproduce, and verify the fix with 2 Arduino Uno boards. The delay in the master can also be removed.


cmaglie:
rebased and merged: https://github.com/arduino/Arduino/commit/48ff3a10ad465afc08a76874cf0e9ef25a6da86a

PR #3737, opened 2015-08-27
Add isFlushed() method to HardwareSerial class for AVR and UARTClass for SAM

thatcadguy:
Needed a non-blocking way to check if the transmit buffer was flushed/clear. This is a re-request because I stink at Git, original request was #3589. Sorry!

### DIFF
--- a/build/shared/lib/keywords.txt
+++ b/build/shared/lib/keywords.txt
@@ -152,6 +152,7 @@ println	KEYWORD2	Serial_Println
 available	KEYWORD2	Serial_Available
 availableForWrite	KEYWORD2
 flush	KEYWORD2	Serial_Flush
+isFlushed	KEYWORD2
 setTimeout	KEYWORD2
 find	KEYWORD2
 findUntil	KEYWORD2
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
@@ -210,6 +210,12 @@ void HardwareSerial::flush()
   // the hardware finished tranmission (TXC is set).
 }
 
+bool HardwareSerial::isFlushed(void)
+{
+  if (!_written) return true;
+  return (_tx_buffer_head == _tx_buffer_tail) && bit_is_clear(*_ucsrb, UDRIE0) && bit_is_set(*_ucsra, TXC0);
+}
+
 size_t HardwareSerial::write(uint8_t c)
 {
   _written = true;
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.h
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.h
@@ -126,6 +126,7 @@ class HardwareSerial : public Stream
     virtual int read(void);
     int availableForWrite(void);
     virtual void flush(void);
+    virtual bool isFlushed(void);
     virtual size_t write(uint8_t);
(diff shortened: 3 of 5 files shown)


thatcadguy:
@PaulStoffregen @matthijskooijman Matt, you were right about the AVR code, should be fixed now. Everything is in one nice commit now as well.

matthijskooijman:
I believe the AVR code still has a race condition. Note that `flush()` checks the `UDRIE0` (uart data register empty interrupt enable) bit, not `UDRE0`.

thatcadguy:
At this point I've spent way too much time on this. If someone wants to
jump in and correct the AVR portion then be my guest, otherwise this pull
will die.

On Thu, Aug 27, 2015 at 9:24 AM, Matthijs Kooijman <notifications@github.com

PaulStoffregen:
If you do work on this more, I highly recommend connecting a MAX483 or similar chip.  Just a little code after writing data should poll isFlushed() and de-assert the driver enable pin.  If you monitor the chip's output, you'll be able to very easily tell if you're truncating the data.  Or the same could be accomplished with a logic OR gate, or with a digital oscilloscope and no extra circuits at all.

PaulStoffregen:
Also, if you want some other code to look at, my HardwareSerial code for Teensy has an extension called transmitterEnable(pin), which causes it to generate the driver enable control signal for those common RS-485 chips.  Teensy has provided this feature for years.  I have personally used it in projects and so have hundreds of other people, so I can say with very good confidence it has correct timing on controlling that pin on both AVR and ARM chips.

thatcadguy:
Have a Due and a SP3485, going to do this within the hour actually. Don't
have an AVR based board to test anything out on. Honestly only care about
the SAM part because that's what I need it for. Someone more familiar with
the AVR side should re-write isFlushed() for that.

matthijskooijman:
@thatcadguy, regarding the AVR code, I think (based on the current `flush()` function, which IIRC I wrote a while ago) replacing the return line in your function with this should be sufficient:

    return bit_is_clear(*_ucsrb, UDRIE0) && bit_is_set(*_ucsra, TXC0);

I don't have time right now to actually test this, but perhaps you can put the above line in your PR for anyone else to test?

thatcadguy:
@matthijskooijman Updated. I tested the SAM code on the Due today, works fine.

thatcadguy:
Just rebased this in case anyone merges it soon.

thatcadguy:
@cmaglie Hi, is there something else my branch needs to get merged? I'd like to see this get merged soon.

Chris--A:
I know its only being picky, but what about simply `flushed()` as the function name.

thatcadguy:
@Chris--A If that's what it takes to get this merged, then sure.

facchinm:
Partially moved to arduino/ArduinoCore-sam#7

thatcadguy:
Closing because this hasn't been acted upon in 2 years.

budulinek:
Please reopen and merge this commit. We need isFlushed() method for non-blocking Serial.write() on half duplex communication (such as RS485).

Thanks.

matthijskooijman:
This is no longer the right place for this PR. To have it merged, it should be resubmitted to the ArduinoCore-avr repo (if not already done, for ArduinoCore-sam there seems to be https://github.com/arduino/ArduinoCore-sam/pull/7). Also, any discussion about adding this method more generally into the Stream class (which might be appropriate), would need to happen in the ArduinoCore-API repo probably.

Issue #2173, opened 2014-07-09
I2C/TWI library: Race Condition with Repeated STARTs


earlyprogrammer:
I believe I've identified a race condition with repeated STARTS in the TWI library.
(As Master, doesn't seem related to the TWI slave reports I found searching online)
I'm using a modified version of Wire and TWI, I'll discuss my changes later.

The Flow:
Instead of sending a stop, send a repeated start.  Note: TWIE is set to 0, so the ISR should not get called when the hardware flags that sending the START has completed.
(ISR)
```
if (twi_sendStop)
    twi_stop();
else {
  twi_inRepStart = true;	// we're gonna send the START
  // don't enable the interrupt. We'll generate the start, but we 
  // avoid handling the interrupt until we're in the next transaction,
  // at the point where we would normally issue the start.
  TWCR = _BV(TWINT) | _BV(TWSTA)| _BV(TWEN) ;
  twi_state = TWI_READY;
}
```
Next, the flow skips the wait loop, believing that the state is READY.
(readFrom or writeTo)
```
while(TWI_READY != twi_state){
    continue;
  }
//various prep variable assignments
```
Finally, put the address in TWDR, then initiate the transmission and reenable the interrupts
(readFrom or writeTo)
```
if (true == twi_inRepStart) {
    // if we're in the repeated start state, then we've already sent the start,
    // (@@@ we hope), and the TWI statemachine is just waiting for the address byte.
    // We need to remove ourselves from the repeated start state before we enable interrupts,
    // since the ISR is ASYNC, and we could get confused if we hit the ISR before cleaning
    // up. Also, don't enable the START interrupt. There may be one pending from the 
    // repeated start that we sent outselves, and that would really confuse things.
    twi_inRepStart = false;			// remember, we're dealing with an ASYNC ISR
    TWDR = twi_slarw;				
    TWCR = _BV(TWINT) | _BV(TWEA) | _BV(TWEN) | _BV(TWIE);	// enable INTs, but not START
  }
```

CASE 1: intended
#The hardware finishes the repeated start before `TWDR = twi_slarw;`
Hardware sets TWINT to 0, but because TWIE is disabled, ISR will not be called.
Address will be put in the TWDR, then the hardware will be initiated by setting TWINT to 1.

CASE 2: works
#The hardware finishes the repeated start after `TWCR = _BV(TWINT) | _BV(TWEA) | _BV(TWEN) | _BV(TWIE);`
[...]

PR #3742, opened 2015-08-28
Check TWWC (Write Collision Flag) bit after setting TWDR

sandeepmistry:
As suggested by @earlyprogrammer in #2173, to ensure TWDR value is set if there is a write collision. See #2173 for more details.


cmaglie:
rebased and merged: https://github.com/arduino/Arduino/commit/c71f5e9f883bac00f4013e4c94579bbd81002c58

PR #3747, opened 2015-08-31
Added EXTERNAL_NUM_INTERRUPTS for AVR xxU2 series

cmaglie:
@NicoHood 
are you sure that the correct number is 8? it sounds weird since the 32U4 has 5.


NicoHood:
Yes I am. u2 and u4 are VERY different. u2 doesnt even have analog pins.
http://www.atmel.com/Images/doc7799.pdf
Page 64

PR looking good.

PR #3728, opened 2015-08-25
Examples: Add DNS to Ethernet.begin

eddyst:
Set the DNS-Server as third parameter of Ethernet.begin if at least gateway is specified. Reference is Updated before.

In BarometricPressureWebServer gateway and subnet wasn't used.

Signed-off-by: Eddy <est.git@online.de>

PR #3764, opened 2015-09-03
Correct Ethernet.begin()

PaoloP74:
The correct inizialization of Ethernet.begin() is:
- Ethernet.begin(mac);
- Ethernet.begin(mac, ip);
- Ethernet.begin(mac, ip, dns);
- Ethernet.begin(mac, ip, dns, gateway);
- Ethernet.begin(mac, ip, dns, gateway, subnet);

Ethernet.begin(mac, ip, gateway, subnet); is a common mistake.
--> https://www.arduino.cc/en/Reference/EthernetBegin

Will need to change the examples in the Reference.

### DIFF
--- a/libraries/Ethernet/examples/AdvancedChatServer/AdvancedChatServer.ino
+++ b/libraries/Ethernet/examples/AdvancedChatServer/AdvancedChatServer.ino
@@ -41,7 +41,7 @@ EthernetClient clients[4];
 
 void setup() {
   // initialize the ethernet device
-  Ethernet.begin(mac, ip, gateway, subnet);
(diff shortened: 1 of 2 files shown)


RobAtticus:
Would it be clearer to have two vars, `dns` and `gateway`, that point to the same `IPAddress` rather than list `gateway` twice?

PaoloP74:
Like this? #3728.

RobAtticus:
Yes, I think that's a good example.

cmaglie:
I've merged the changes in #3728 that are slightly more clear.

Thank you!

PR #3780, opened 2015-09-11
DNSClient::inet_aton fix and new IPAddress::fromString method

cmaglie:
- fixed a parsing bug in `DNSClient::inet_aton` as reported in #2500 (the solution proposed in #2500 is only a partial fix)
- Since it's not DNS related, the same functionality has been added to IPAddress.

At this point `DNSClient::inet_aton` can be transformed to just an inline call to `IPAddress::fromString` but this will make the Ethernet library incompatible with old cores, so the temporary compromise is that the code is duplicated until the new core becomes more widespread.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/IPAddress.cpp
+++ b/hardware/arduino/avr/cores/arduino/IPAddress.cpp
@@ -43,6 +43,48 @@ IPAddress::IPAddress(const uint8_t *address)
     memcpy(_address.bytes, address, sizeof(_address.bytes));
 }
 
+bool IPAddress::fromString(const char *address)
+{
+    // TODO: add support for "a", "a.b", "a.b.c" formats
+
+    uint16_t acc = 0; // Accumulator
+    uint8_t dots = 0;
+
+    while (*address)
+    {
+        char c = *address++;
+        if (c >= '0' && c <= '9')
+        {
+            acc = acc * 10 + (c - '0');
+            if (acc > 255) {
+                // Value out of [0..255] range
+                return false;
+            }
+        }
+        else if (c == '.')
+        {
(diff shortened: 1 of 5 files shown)


jhorck:
I proposed #2500. This fix looks good to me too, though I didn't test it. 

I don't understand the comment "// TODO: add support for "a", "a.b", "a.b.c" formats"
IP addresses always have 4 bytes (IPv4). I hope fewer bytes are NEVER accepted as a valid IP address?

cmaglie:
Hi @jhorck

thanks for the comment.

> I don't understand the comment "// TODO: add support for "a", "a.b", "a.b.c" formats"

believe it or not you can specify the ip address using the following format:

`A.A.A.A`
`A.A.B`
`A.C`
`D`

where:
- A is an 8-bit unsigned
- B is a 16-bit unsigned
- C is a 24-bit unsigned
- D is a 32-bit unsigned

for example on my linux box I can do:

```
:~$ ping 2130706433
PING 2130706433 (127.0.0.1) 56(84) bytes of data.
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.056 ms
```

or better.

```
~$ ping 127.1
PING 127.1 (127.0.0.1) 56(84) bytes of data.
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.043 ms
```

I didn't know that until I looked to some implementations of inet_aton (for example http://www.comp.nus.edu.sg/~cs3223/doxygen-postgres9.4.0/html/dd/dbe/inet__aton_8c_source.html)

BTW it's a so rarely used feature that I guess we'll leave it unsupported.

jhorck:
@cmaglie, You're right! I didn't know that!

PR #3811, opened 2015-09-17
Add iSerial field to Native USB devices

facchinm:
Windows PCs always believe that an USB device with a particular triplet of VID/PID/SerialNumber will never change during its existence, thus saving the device's "nature" on the Registry. 
With the introduction of PluggableUSB, a device can change its "composition" between a sketch and another if we include different headers. But the new device does not necessarily implement all the capabilities of the one that has been saved, leading to drivers not loading etc.

To solve this problem, two different approaches where tested:
- creating a S/N based on the descriptor length (collisions still possible, less overhead)
- craft a S/N string from the pluggable modules properties (no collision, no difference between different HID submodules) (this seems better)

A call has been added (`getShortName`) and it MUST be implemented by any library based on PluggableUSB. No effect on PluggableHID-based libraries.

The patchset is composed by two patches (which will be squashed afterwards) to show the first approach on AVR (first patch)

@NicoHood  @matthijskooijman @PaulStoffregen @cmaglie

### DIFF
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.h
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.h
@@ -35,6 +35,7 @@ protected:
   virtual bool setup(USBSetup& setup) = 0;
   virtual int getInterface(uint8_t* interfaceCount) = 0;
   virtual int getDescriptor(USBSetup& setup) = 0;
+  virtual uint8_t getShortName(char *name) { name[0] = 'A'+pluggedInterface; return 1; }
 
   uint8_t pluggedInterface;
   uint8_t pluggedEndpoint;
@@ -55,6 +56,7 @@ public:
   int getInterface(uint8_t* interfaceCount);
   int getDescriptor(USBSetup& setup);
   bool setup(USBSetup& setup);
+  void getShortName(char *iSerialNum);
 
 private:
   uint8_t lastIf;
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -50,6 +50,15 @@ int PluggableUSB_::getDescriptor(USBSetup& setup)
 	return 0;
 }
 
+void PluggableUSB_::getShortName(char *iSerialNum)
+{
+	PluggableUSBModule* node;
+	for (node = rootNode; node; node = node->next) {
+		iSerialNum += node->getShortName(iSerialNum);
+	}
+	*iSerialNum = 0;
+}
+
 bool PluggableUSB_::setup(USBSetup& setup)
 {
 	PluggableUSBModule* node;
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -18,6 +18,7 @@
 
 #include "USBAPI.h"
 #include "PluggableUSB.h"
+#include <stdlib.h>
 
 #if defined(USBCON)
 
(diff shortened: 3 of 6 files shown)


facchinm:
Companion branches:
https://github.com/arduino-libraries/MIDIUSB/tree/nativeUSB_serialNumber
https://github.com/facchinm/ArduinoCore-samd/tree/nativeUSB_serialNumber

NicoHood:
Thx for the CC.

Does it really make Windows see this as different device? I never knew about this ISERIAL feature, is there any docs where I can read this up?

I guess the lufa equivalent is this:
https://github.com/abcminiuser/lufa/blob/master/Projects/USBtoSerial/Descriptors.c#L63

But you know, that changing an HID device to another device also causes problems. For example if you add a mouse and the next time its a gamepad. Sometimes windows then recognizes the gamepad axis as mouse and other strange things. Luckly I dont use windows anymore. How do you want to fix this issue?

facchinm:
The solution is simply to tell Windows that any combination of composite device is another device by changing the iSerial field (LUFA equivalent is perfect).

Anyway, I thought that every HID-compliant device would be compatible, as it appears only one on `Device manager`. 
If you tell me that Win has strange behaviour also with different HID submodules I'd combine the two approaches so any device is "unique" using a signature composed by modules name + descriptor length; in this way `HI56` (HID with Mouse) will be different from `HI128` (HID with Keyboard and Joystick) and `HI56` will also be different from `MI56` (MIDI).

What do you think about that?

NicoHood:
Doesnt this lower the number of possible devices, if the string is longer? Does it has a max length?
I just said what I remember. I have not tested the pluggable HID with windows and I wont be able to do this again. If it bugs someone, we can fix it later again (upgrade this patch). For now I'd leave it like that. Also for windows 8 you can always uninstall the HID drivers in "Devices and printers".

I suggest to keep it like that (havent tested yet, but use linux anyways), and if someone reports a problem to my HID project, I will come back here again. For the normal Arduino purposes (keyboard + mouse) this should be fine.

Before we can use the HID-Project the .a linkage for libraries needs to be implemented for 1.6.6. My PR is out of date because the Arduino builder is now coded in go which I dont get. and not a single line was documented. So I dont know how to go on with this anyways. If this is all fixed and wokring we can test if the ISerial solution is good enough or not.

facchinm:
Patches rebased on newer PluggableUSB API

NicoHood:
How much chars may the ISERIAL have? Does it have to be X chars or X-Y? (With up to 6 custom endpoints this may exceed the limits)

facchinm:
I've arbitrarily chosen 20 chars as limit, since
* CDC doesn't add anything
* HID can add a max of 6 bytes ("HID" + length_descriptor -> max "999")
* MIDI adds 4 bytes ("MIDI")

So we are plenty of spare space with 20bytes (I believe)
Since I'm doing a lot of testing on Win I believe this issue should be addressed now and not after the first bunch of random issues appear :wink:

NicoHood:
Just asking for the USB limitation (if any).

How big is the overhead? Is it really needed? I had no problems with linux, but well its linux *G*

Are you going to make the CDC Serial pluggable (or anyone else)?

Have you already updated the MIDI? Does it work with an endpoint OUT or via control EP out? I am wondering how one would get the interrupt.

facchinm:
I wanted to answer you two days ago about making CDC pluggable but I totally forgot :smile: 
The main issue there is that some hook in irq callback must be added, with a (probably) excessive overhead . 
This is why MIDIUSB library uses an OUT endpoint to receive data but it polls for that data instead than relying on the interrupt-based structure.

To be back in topic, Linux and OSX users do not suffer the issue since their drivers structure does not use any non-volatile backing storage

NicoHood:
You mean the flush IRQ? Maybe this can be added as special weak callback. If the Serial is not used it just calls nothing? Havent looked deeper into that, but deactivating the Serial would be nice.

**[Or we would deactivate the USB core](https://github.com/arduino/Arduino/pull/3679)**. If we do this we can get rid of everything or even implement another USB core with very special features. I also though of adding the Lufa to Arduino as USB core. Not sure if the building system would work that simple.

It would be nice to add this feature via a #define so I can save this overhead. The more HID devices you have, the more overhead it will give. Or maybe there is another solution? I'd like to avoid this overhead.

facchinm:
squashed into a single commit + memory improvements 
/cc @cmaglie

NicoHood:
Meh for the overhead, thumbsup for windowsusers.
Have you added this to the wiki docs?

hardware/arduino/avr/cores/arduino/PluggableUSB.h
```diff
@@ -35,6 +35,7 @@ class PluggableUSBModule {
   virtual bool setup(USBSetup& setup) = 0;
   virtual int getInterface(uint8_t* interfaceCount) = 0;
   virtual int getDescriptor(USBSetup& setup) = 0;
+  virtual uint8_t getShortName(char *name) { name[0] = 'A'+pluggedInterface; return 1; }
```
NicoHood: Isnt this line consuming a lot memory? I havent tested it but I think the conversion from int to char could take some overhead. Okay its just the default implementation, but maybe we can improve this?

cmaglie: Yes we can surely improve, are you suggesting to do something like:

`name[0] = 'A' + ((uint8_t)pluggedInterface)`

?

NicoHood: Nevermind I missunderstood the line. It is correct.

I thought it does something similar as in java, where it merges them together as string (with 2 chars). But this is just fine like that. However its not really save, but for this reason users should overwrite this function of course.

But I'd still like to be able to remove this ISERIAL if possible. Its just so useless for linux users. Has someone measured the added overhead?

PR #3815, opened 2015-09-17
Expose optional DHCP timeout parameters to Ethernet::begin()

chaveiro:
A no brainer update.
Exposes optional timeout parameter values of dhcp for user to adjust on
their code.
Currently if dhcp server is not available, code will block inside
DhcpClass::request_DHCP_lease() for 60s. Having this parameters exposed
user can specify their timeout requirements without changing the lib
code.
Change is backward compatible as parameters are optional and defaults to
existing values.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -8,7 +8,7 @@ uint8_t EthernetClass::_state[MAX_SOCK_NUM] = {
 uint16_t EthernetClass::_server_port[MAX_SOCK_NUM] = { 
   0, 0, 0, 0 };
 
-int EthernetClass::begin(uint8_t *mac_address)
+int EthernetClass::begin(uint8_t *mac_address, unsigned long timeout, unsigned long responseTimeout)
 {
   static DhcpClass s_dhcp;
   _dhcp = &s_dhcp;
@@ -22,7 +22,7 @@ int EthernetClass::begin(uint8_t *mac_address)
   SPI.endTransaction();
 
   // Now try to get our config info from a DHCP server
-  int ret = _dhcp->beginWithDHCP(mac_address);
+  int ret = _dhcp->beginWithDHCP(mac_address, timeout, responseTimeout);
   if(ret == 1)
   {
     // We've successfully found a DHCP server and got our configuration info, so set things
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Change looks ok to me. I do wonder if the API would not be better with a separate setTimeout() function instead of adding more parameters to `begin()`, since the latter has the tendency to get messy quickly. Perhaps @ffissore or @cmaglie has an idea about that?

chaveiro:
matthijskooijman a separate setTimeout() function would increase memory footprint, i personally prefer as on this commit.

matthijskooijman:
@chaveiro, how would this add footprint? The timeouts are already stored in parameters inside the dhcp class (or however it is called exactly). So you won't need to add any variables inside the Ethernet classes, but instead modify the Dhcp classes so you can set the timeout before initialization happens (this does need some more invasive changes to the dhcp code, that's true).

chaveiro:
But you would then have to include "Dhcp.h" on the arduino project or modify Ethernet.h to expose that dhcp related function. This way is cleaner.
Anyway if you have other solution, please modify the commit as desired. The requirement is that the values can be specified from arduino ide without much trouble.

matthijskooijman:
My suggestion would be to add a function to Dhcp.h and call that function from another function in Ethernet.h. Like I said, it's a bit more complex, but might result in a cleaner API on the long run. However, I'm not sure which of the options is the best one, nor am I authoritive here - @cmaglie or @ffissore might better respond here.

cmaglie:
This options are already saved in Dhcp class members:
https://github.com/arduino/Arduino/blob/master/libraries/Ethernet/src/Dhcp.cpp#L17
so the fix is good as it is IMHO.

matthijskooijman:
> This options are already saved in Dhcp class members:

My point was not that these things should be saved somewhere, my point was that they should *not* be constructor arguments (since passing options to the constructor does not scale very well if more and more options are added). Having said that, just this option isn't a problem directly, so doing it like this could be ok.

q2dg:
This should be documented in web (https://www.arduino.cc/en/Reference/EthernetBegin)

chaveiro:
Asked, see : http://forum.arduino.cc/index.php?topic=352476.0

RonniePoulsen:
Can someone please tell me what is the difference between "timeout" and "responseTimeout". I can't seem to find any documentation on this.

PR #3825, opened 2015-09-20
Remove Raw HID definitions at wrong place

NicoHood:
Raw HID is not used. And if someone (like me) want to create a library for this he can and should define those definitions himself. This code is a leftover from the old USB-Core where the raw HID also never got used.


facchinm:
Correct, it was the last leftover of the porting. Thanks!

PR #3839, opened 2015-09-22
Fix CDC write() when DTR is not set

NicoHood:
Fix https://github.com/arduino/Arduino/issues/1182

### DIFF
--- a/hardware/arduino/avr/cores/arduino/CDC.cpp
+++ b/hardware/arduino/avr/cores/arduino/CDC.cpp
@@ -206,7 +206,7 @@ size_t Serial_::write(const uint8_t *buffer, size_t size)
 	// TODO - ZE - check behavior on different OSes and test what happens if an
 	// open connection isn't broken cleanly (cable is yanked out, host dies
 	// or locks up, or host virtual serial port hangs)
-	if (_usbLineInfo.lineState > 0)	{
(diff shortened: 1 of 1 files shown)


NicoHood:
@matthijskooijman build this please

cmaglie:
@ArduinoBot build this please

NicoHood:
The PR like this does not work for me I've noticed. If the host is not listening my RawHID dev sketch will block for a long time. I tried to check for baud() but this also didnt help (set the value default to zero of course).

But adding ` && availableForWrite()` helped.

I will add this to the patch. The only thing we could also add would be baud() but that was not needed in this case. Tested under linux 64 bit elementary os.

PeterVH:
I think testing for availableForWrite() inside write() is not correct. It looses the automatic flow control  USB provides: if the process on the host temporarily does not catch up with the data you send it, the write method should block till data is sent. If this is not acceptable for an application, it should call availableForWrite() by itself.

>It is not cleared on disconnect? Where do we have to handle this? (this is very important)

Just consulted the datasheet. The VBUSTI irq fires upon connect/disconnect. The VBUS status bit in USBSTA tells you the current state. Did not try this though.

NicoHood:
Without the writeavailable my device locked up. Have to check this again. Feel free to also test it, I will not have time to debug this in the next days.

PR #3840, opened 2015-09-22
[PluggableHID] API simplification proposal

cmaglie:
Since the PluggableHID is not yet officially released maybe we could make some makeup on the API?
This is a **breaking change for the early adopters** of the Pluggable HID, but if we decide to merge it better doing it now that the library is not published.

The most ugly part for me is how the HID "modules" are defined, for example:

```C++
Keyboard_::Keyboard_(void) 
{
	static HID_Descriptor cb = {
		.length = sizeof(_hidReportDescriptor),
		.descriptor = _hidReportDescriptor,
	};
	static HIDDescriptorListNode node(&cb);
	HID.AppendDescriptor(&node);
}
```

IMHO the HID_Descriptor is a useless level of abstraction and can be eliminated.
After this PR the initialization becomes:

```C++
Mouse_::Mouse_(void) : _buttons(0)
{
    static HIDDescriptorListNode node(_hidReportDescriptor, sizeof(_hidReportDescriptor));
    HID.AppendDescriptor(&node);
}
```

This give us also a small gain in performance and in flash and ram usage as well.

/cc @NicoHood @facchinm @matthijskooijman

### DIFF
--- a/hardware/arduino/sam/libraries/HID/HID.cpp
+++ b/hardware/arduino/sam/libraries/HID/HID.cpp
@@ -68,7 +68,7 @@ int HID_GetDescriptor(int8_t t)
 		HIDDescriptorListNode* current = rootNode;
 		int total = 0;
 		while(current != NULL) {
-			total += USBD_SendControl(0,current->cb->descriptor,current->cb->length);
+			total += USBD_SendControl(0,current->descriptor->data,current->descriptor->length);
 			current = current->next;
 		}
 		return total;
@@ -89,7 +89,7 @@ void HID_::AppendDescriptor(HIDDescriptorListNode *node)
 		current->next = node;
 	}
 	modules_count++;
-	sizeof_hidReportDescriptor += node->cb->length;
+	sizeof_hidReportDescriptor += node->descriptor->length;
 }
 
 void HID_::SendReport(uint8_t id, const void* data, int len)
--- a/hardware/arduino/sam/libraries/HID/HID.cpp
+++ b/hardware/arduino/sam/libraries/HID/HID.cpp
@@ -68,7 +68,7 @@ int HID_GetDescriptor(int8_t t)
 		HIDDescriptorListNode* current = rootNode;
 		int total = 0;
 		while(current != NULL) {
-			total += USBD_SendControl(0,current->descriptor->data,current->descriptor->length);
+			total += USBD_SendControl(0,current->data,current->length);
 			current = current->next;
 		}
 		return total;
@@ -89,7 +89,7 @@ void HID_::AppendDescriptor(HIDDescriptorListNode *node)
 		current->next = node;
 	}
 	modules_count++;
-	sizeof_hidReportDescriptor += node->descriptor->length;
+	sizeof_hidReportDescriptor += node->length;
 }
 
 void HID_::SendReport(uint8_t id, const void* data, int len)
--- a/hardware/arduino/sam/libraries/HID/HID.h
+++ b/hardware/arduino/sam/libraries/HID/HID.h
@@ -42,16 +42,12 @@
 #define HID_REPORT_DESCRIPTOR_TYPE      0x22
 #define HID_PHYSICAL_DESCRIPTOR_TYPE    0x23
 
-typedef struct __attribute__((packed)) {
-  uint8_t length;
-  const void* data;
-} HID_Descriptor;
-
 class HIDDescriptorListNode {
 public:
   HIDDescriptorListNode *next = NULL;
-  const HID_Descriptor *descriptor;
-  HIDDescriptorListNode(const HID_Descriptor *d) : descriptor(d) { }
+  HIDDescriptorListNode(const void *d, uint16_t l) : data(d), length(l) { }
+  uint8_t length;
+  const void* data;
 };
 
 class HID_
--- a/hardware/arduino/sam/libraries/HID/HID.h
+++ b/hardware/arduino/sam/libraries/HID/HID.h
@@ -44,14 +44,14 @@
 
 typedef struct __attribute__((packed)) {
   uint8_t length;
-  const void* descriptor;
+  const void* data;
 } HID_Descriptor;
 
 class HIDDescriptorListNode {
 public:
   HIDDescriptorListNode *next = NULL;
-  const HID_Descriptor * cb;
-  HIDDescriptorListNode(const HID_Descriptor *ncb) {cb = ncb;}
+  const HID_Descriptor *descriptor;
+  HIDDescriptorListNode(const HID_Descriptor *d) : descriptor(d) { }
 };
 
 class HID_
--- a/hardware/arduino/avr/libraries/HID/HID.cpp
+++ b/hardware/arduino/avr/libraries/HID/HID.cpp
@@ -61,7 +61,7 @@ int HID_GetDescriptor(int8_t t)
 		HIDDescriptorListNode* current = rootNode;
 		int total = 0;
 		while(current != NULL) {
-			total += USB_SendControl(TRANSFER_PGM,current->cb->descriptor,current->cb->length);
+			total += USB_SendControl(TRANSFER_PGM,current->descriptor->data,current->descriptor->length);
 			current = current->next;
 		}
 		return total;
@@ -82,7 +82,7 @@ void HID_::AppendDescriptor(HIDDescriptorListNode *node)
 		current->next = node;
 	}
 	modules_count++;
-	sizeof_hidReportDescriptor += (uint16_t)node->cb->length;
+	sizeof_hidReportDescriptor += (uint16_t)node->descriptor->length;
 }
 
 void HID_::SendReport(u8 id, const void* data, int len)
--- a/hardware/arduino/avr/libraries/HID/HID.cpp
+++ b/hardware/arduino/avr/libraries/HID/HID.cpp
@@ -61,7 +61,7 @@ int HID_GetDescriptor(int8_t t)
 		HIDDescriptorListNode* current = rootNode;
 		int total = 0;
 		while(current != NULL) {
-			total += USB_SendControl(TRANSFER_PGM,current->descriptor->data,current->descriptor->length);
+			total += USB_SendControl(TRANSFER_PGM,current->data,current->length);
 			current = current->next;
 		}
 		return total;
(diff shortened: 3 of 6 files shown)


NicoHood:
If you really want to change the API I recommend you to do it right and add all the functions I've added in my extended version. Maybe not all, but the key features. Then people can create cool stuff even with the Arduino Core and not only with my patched HID-Project.

I also suggest to make the HID Interface itself "multi pluggable". Right now you have a single HID Interface for all the HID Devices. But you can also create an Interface her HID Device. This avoids lots of problems and makes it easier to add a "non workaround" boot keyboard and an option for things like raw HID.

I was currently working on raw HID and I got it setup properly under linux. Windows wont work right know, I am pretty sure. But I can make this work if I want to.

If you have more time for applying changes I'd like to help out here. But I need more time to some up with a clever and useful plan. If we change this over and over again, noone is happy. If we think clearly about it, we probably can make this API pretty fancy. And the way the pluggable HID works right now is pretty impressive I think.

How do you feel about this?

NicoHood:
Thinking about this some more it would probably make sense to build the pluggable USB the same way I improved the pluggable HID. If you create a mean class USBInterface that every other Interface can inherit (HID, Serial) its way simpler to code multiple HID Interfaces. Because then you can add a `this` pointer automatically via inheritation etc. This also gives us the option to avoid static functions inside or outside the class and probably saves a lot of doubled pointers (ram + flash again).

Once this is done it would be possible to create a general HID Interface class with a similar API that can be used for different HID Devices (if you want each device on a separate endpoint. Reasons why described in the post above).

This requires even more changes but helps to make the Pluggable USB as best as possible. I think I am not able to improve the HID unless the USB itself is patched (there we have this problem again).

Comments?

cmaglie:
Hi @NicoHood 

well, when I said "make some makeup on the API" I meant "let's see if we can do some **small** adjustments on the API" before the release :-)

With this PR my aim is just to remove the ugliness of the `HID_Report` structure because it seems more like a refactoring leftover than something put there on purpose and this is confirmed also by the presence, on the same `HID.h`, of another struct called `HIDReport` (without the underscore!).

I understand your goodwill to change the API as best as possible but, please, consider also release timings in the mix: making **substantial** changes in the HID API now implies doing another round of porting/coding/testing on various OS (winXP/7/8/10, Mac 10.7/8/9/10, Linux 32/64) and various platform (AVR, SAM, SAMD) that may take long time.
I don't want to release a sub-optimal API, but on the other hand I don't want to postpone forever waiting for perfection.

So to move forward my question is: all the missing feature you mentioned may be added, not now, but in a second release of the HID library? Adding new features in a second release without breaking the current API is generally fine, so my question is intended to understand if all the new features you have in mind could be added in a second time without breaking the current API.

NicoHood:
All features I have in mind are not possible with the current API. At least for HID. I can still write my own HID API, no problem. But it would be easier if that would be in the Arduino code.

Maybe I'd also change some stuff in the USB API too, (similar flexible changes as in the HID). The USB API is more important than the HID, because I always can replace the HID API with my own. But you are right, we cant postpone this forever. I think I will go with my own version and Arduino with its own.

Its all about backwards compatibility, I know. That holds back the progress.

I have to see what is most annoying in the USB API that needs to be changed. I think I will hold back my HID changes since I am still developing much on this and dont have a clear Idea whats the best solution for everything.

When is the next release planned? So I have some time to look at the USB Core or just give up since its not enough time.

NicoHood:
The PR itself looks good to me though. That should be changed at minimum ;)

hardware/arduino/sam/libraries/HID/HID.h
```diff
@@ -42,16 +42,12 @@
...
-  const void* descriptor;
-} HID_Descriptor;
-
 class HIDDescriptorListNode {
 public:
   HIDDescriptorListNode *next = NULL;
-  const HID_Descriptor * cb;
-  HIDDescriptorListNode(const HID_Descriptor *ncb) {cb = ncb;}
+  HIDDescriptorListNode(const void *d, uint16_t l) : data(d), length(l) { }
+  uint8_t length;
```
NicoHood: uint16_t? Also this could be const (avr as well)?

hardware/arduino/sam/libraries/HID/HID.cpp
```diff
@@ -68,7 +68,7 @@ int HID_GetDescriptor(int8_t t)
 		HIDDescriptorListNode* current = rootNode;
 		int total = 0;
```
NicoHood: Is there a reason why you always use int in the Arduino programs? uint16_t would make life much easier in most cases.

NicoHood:
Yeah I checked that again.

It would be nice to pack this stuff inside a real class that can be inherited:
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/PluggableUSB.h#L28-L49

This way we can generalize and USB Device and inherit it further. For example I can inherit the usb device, create global HID class and then specify different HID Device classes (keyboard etc) with different endpoints.

Similar to this:
https://github.com/NicoHood/HID/blob/dev_2_4/src/PluggableHID/HIDDevice.h
https://github.com/NicoHood/HID/blob/dev_2_4/src/ImprovedKeyboard.h#L53

At the moment I can only duplicate the code with a lot of messy stuff.

hardware/arduino/sam/libraries/HID/HID.h
cmaglie: > uint16_t?

It should allow for descriptors >256.

> Also this could be const (avr as well)?

Mbah, does it really matter? `l` is passed by copy so the method can't change the "original" parameter in any way.

NicoHood: yes. but its uint8_t. look ;D

Its not about changing permissions, its about code style, debugging and compiler optimization.

hardware/arduino/sam/libraries/HID/HID.cpp
cmaglie: I'm not sure, since I'm not the author of the USB stack. I guess the reason is that `USBD_SendControl` returns `-1` in case of failure.

(well, for the sake of precision, it seems that `USBD_SendControl` in the SAM core will never returns `-1` but the equivalent function in the AVR core, `USB_SendControl` does, probably this is something that has been inherited when the USB stack has been ported from AVR to SAM).

To fix this we probably should handle the error condition, and eventually return -1 if one call to `USB(D)_SendControl` fails.

hardware/arduino/sam/libraries/HID/HID.h
cmaglie: Ahhh, right good catch! ;-)

Teiwaz83:
As one of the "early adopters" of PluggableHID, I'd rather have to do a bit of rework now than deal with a mess and/or limited functionality later. Here's some minor sore spots for me as I've been working with it:

It might be smart to rename (from HID.h)

HID_ENDPOINT_INT
HID_INTERFACE

These guys used to be constants defined by macros before PluggableUSB. Now they just look like they are. I found this very confusing when reading the constructor for HID and HID_GetInterface, as it looked at first brush like there was a weird circular dependency between the constructor and HID_GetInterface. It all makes sense when you dig into it, but starting out with wrong assumptions about what these are doesn't help. Also, it leads to painful things like this:

int HID_::HID_GetInterface(u8* interfaceNum)
{
	interfaceNum[0] += 1;	// uses 1
	_hidInterface.hid =
#if defined(USE_BOOT_KEYBOARD_PROTOCOL)
		D_INTERFACE(HID_INTERFACE,1,3,1,1);
#elif defined(USE_BOOT_MOUSE_PROTOCOL)
		D_INTERFACE(HID_INTERFACE,1,3,1,2);
#else
		D_INTERFACE(HID_INTERFACE,1,3,0,0);
#endif
	_hidInterface.desc = D_HIDREPORT(sizeof_hidReportDescriptor);
	_hidInterface.in = D_ENDPOINT(USB_ENDPOINT_IN (HID_ENDPOINT_INT),USB_ENDPOINT_TYPE_INTERRUPT,USB_EP_SIZE,0x01);
	return USB_SendControl(0,&_hidInterface,sizeof(_hidInterface));
}

Half of the constant/macro-looking things in there are actually variables. More than that, they're variables which *don't actually have their values set yet* at the only point this function is referenced in the HID class.

Also, I'd suggest doing a quick pass on the various descriptor variable names and structure names:

typedef struct
{
  u8 len;     // 9
  u8 dtype;   // 0x21
  u8 addr;
  u8  versionL; // 0x101
  u8  versionH; // 0x101
  u8  country;
  u8  desctype; // 0x22 report
  u8  descLenL;
  u8  descLenH;
} HIDDescDescriptor;

typedef struct 
{
  InterfaceDescriptor     hid;
  HIDDescDescriptor     desc;
  EndpointDescriptor      in;
} HIDDescriptor;

The... HID Descriptor contains a HID... Descriptor Descriptor? So the HIDDescriptor _hidInterface contains a HID Descriptor Descriptor called desc. Not to be confused with the InterfaceDescriptor contained by the HID Descriptor called hid. Not the HID Descriptor, mind you, as we already established, the HIDDescriptor is called _hidInterface. Headache.

I find myself having to constantly refer back to the USB spec, trying to figure out what various almost identically named macros and structures are by counting the number of bytes in them. I'd suggest making the names of these structures and their members to precisely match their equivalents in the USB spec for easy comparison and less ambiguity. (Hell, I'd also stick in comments with hyperlinks to the relevant parts of the USB spec. But I assume that'd be frowned upon here. Go team JavaDoc!)

facchinm:
Hi @Teiwaz83 and @NicoHood , you are obviously pointing in the right direction. 
Keep in mind that the whole Pluggable* effort was not a rewrite but a porting of existing code, so no struct was rationalized and existing macros which needed to be dynamic kept the same name (my fault).

Anyway, thanks for sharing your ideas and improvements, the whole HID library needs some love now but it's quite easier than before since it's outside the core :smile:

NicoHood:
First progress can be found here:
https://github.com/arduino/Arduino/compare/master...NicoHood:pluggableUSB?expand=1

I now try to get an actual HID device running. Seems like the change (for the other APIs like HID) is not that huge. And this gives us a whole new more flexible options.

cmaglie:
Hi Nico,

I'm also trying some simplification, most looks similar to yours, have a look here:

https://github.com/arduino/Arduino/compare/master...cmaglie:phid-class?expand=1

It seems that we started at the same time, so let's try to converge the efforts!

Issue #2996, opened 2015-04-18
Not reloading keywords.txt for new boards


NicoHood:
If a new board definition via sketchbook/hardware/myboard has intern libraries with keywords.txt they are not loaded after a board switch.

Example:
Select Arduino Leonardo. Type Gamepad. It will not mark orange. Thats fine.
Now switch to the HID Project Leonardo Board (which has a library with Gamepad as keyword).
You see its still not highlighted.
Close the IDE and restart it. Since the setting is now the HID Project at startup it will load its keywords.

Example file:
https://github.com/NicoHood/HID/tree/master/avr/libraries/HIDBridge

tested with 1.6.3 under Windows

cmaglie:
Fixed with #3847

PR #3847, opened 2015-09-23
Allow per-platform keywords.txt

PaulStoffregen:
This patch allows platforms to define an keywords.txt file, to add their unique keywords.  When a board change also changes platforms, the keywords are reloaded and all editor windows updates with the new keywords.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -117,7 +117,7 @@ public class Base {
   private List<JMenu> boardsCustomMenus;
   private List<JMenuItem> programmerMenus;
 
-  private final PdeKeywords pdeKeywords;
+  private PdeKeywords pdeKeywords;
   private final List<JMenuItem> recentSketchesMenuItems;
 
   static public void main(String args[]) throws Exception {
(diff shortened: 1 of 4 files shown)


NicoHood:
Would this also fix https://github.com/arduino/Arduino/issues/2996 ?

PaulStoffregen:
Yes, I believe it should partially fix #2996.

The keywords are only re-read when changing architecture folders.  If you have a AVR board which uses hardware/arduino/avr for its boards.txt and platform.txt and you switch to a different board, keywords will **not** be re-read if the new board also uses hardware/arduino/avr.  But if you switch to a board that uses hardware/HoodLoader2/avr as its boards.txt and platform.txt, then keywords will be re-read.

Any chance you could give the arduinobot build a quick try?  Would be nice if someone (other than me) could at least confirm it works in at least 1 quick test...

NicoHood:
The specific issue I linked is fixed with this.

PaulStoffregen:
Good :)

Is the assumption correct that keywords only need to be re-read when a board change switches to a new architecture folder?  Are there any cases where this would be needed changing between boards that share the same architecture folder?

NicoHood:
I think the behavior is correct like this. I havent looked at the patch itself but from the logic you described it is correct, no need for a further change.

lmihalkovic:
[LMNTS]

PR #3853, opened 2015-09-23
ability to set a stepper's rpm via float

ccoenen:
Stepper can currently only set rotations per minute as integer. This is not a problem for fast motors, as 1200 rpm is not much different from 1200.5 rpm. For slow motors it's a big deal.

Having no way to go slower than 1 rpm or nothing in between 1 rpm and 2 rpm means people have to resort to workarounds like setting a different (wrong) number of steps instead. To illustrate this more clearly: right now, i have to choose if a single revolution takes 60 seconds or 30 seconds or 20 seconds. There is nothing in between.

I added `setSpeed(float)` as a second signature to not break backwards compatibility.

### DIFF
--- a/libraries/Stepper/src/Stepper.cpp
+++ b/libraries/Stepper/src/Stepper.cpp
@@ -175,7 +175,7 @@ Stepper::Stepper(int number_of_steps, int motor_pin_1, int motor_pin_2,
 /*
  * Sets the speed in revs per minute
  */
-void Stepper::setSpeed(long whatSpeed)
+void Stepper::setSpeed(float whatSpeed)
 {
   this->step_delay = 60L * 1000L * 1000L / this->number_of_steps / whatSpeed;
 }
--- a/libraries/Stepper/src/Stepper.h
+++ b/libraries/Stepper/src/Stepper.h
@@ -91,7 +91,7 @@ class Stepper {
                                  int motor_pin_5);
 
     // speed setter method:
-    void setSpeed(long whatSpeed);
+    void setSpeed(float whatSpeed);
 
     // mover method:
     void step(int number_of_steps);


cmaglie:
Adding this method breaks the old sketches that uses `int` for setting the speed:

```
stepper_oneRevolution.ino: In function 'void setup()':
stepper_oneRevolution.ino:28:24: error: call of overloaded 'setSpeed(int)' is ambiguous
stepper_oneRevolution.ino:28:24: note: candidates are:
In file included from stepper_oneRevolution.ino:18:0:
libraries/Stepper/src/Stepper.h:94:10: note: void Stepper::setSpeed(long int)
     void setSpeed(long whatSpeed);
          ^
libraries/Stepper/src/Stepper.h:95:10: note: void Stepper::setSpeed(float)
     void setSpeed(float whatSpeed);
          ^
exit status 1
call of overloaded 'setSpeed(int)' is ambiguous
```

I don't know what the proper fix can be here. Changing the function name seems odd, also adding 10 different overloads. Maybe method templates can be of help here?

ccoenen:
Awww damn, I thought I had prevented that with the overloaded version. I'll take a closer look.

ccoenen:
Apparently it can't decide which typecast is better. So it's probably better to just offer one thing (I'd opt for the float). I'm preparing another commit, that will replace the previous one.

ccoenen:
I think it would also be another improvement to guard this against very small values. But i haven't included it, yet. I would love to discuss this with someone first.

cmaglie:
Using float force the compiler to always link floating-point math subroutines that are huge and slow: someone may want to use only integer math.

My suggestion is to use Method Templates:

http://www.codeproject.com/Articles/257589/An-Idiots-Guide-to-Cplusplus-Templates-Part#MethodTemplates

the setSpeed function may be implemented as:

```C++
class Stepper {
  public:
    [...]

    template<typename T>
    void setSpeed(T whatSpeed) {
      step_delay = 60L * 1000L * 1000L / this->number_of_steps / whatSpeed;
    }
```
(this is just a draft, I've not tried it)

so the setSpeed methods with the correct type are instantiated "on-demand" only when used.

ccoenen:
That's intersting. It will be a while before i can try that. My suggestion is to leave this pull request open, i'll provide another commit in a few days. Would that be alright with you?

Thanks also for the hint regarding floats problems. I was not aware of that!

ccoenen:
turns out "a while" could be almost five years. I'll close this, i will not get back to it for the next five years, either :-/

PR #3854, opened 2015-09-24
Fix compiler warnings in USBCore.cpp

NicoHood:
Removes a compiler warning and saves 14 bytes of flash. Also adds `static` to variables only found in the file.

As you can see an u8 value is created and returned as int, which is not needed.
If you also apply this correct u8 value to the `bool SendConfiguration(int maxlen)` function all warnings are removed and the code doesnt need an extra 2 byte return value.

Related resources to ensure all functions are u8:
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/USBCore.cpp#L457-L470
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/USBCore.cpp#L441-L452
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/USBCore.h#L146-L156
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/CDC.cpp
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp

### DIFF
--- a/hardware/arduino/avr/cores/arduino/CDC.cpp
+++ b/hardware/arduino/avr/cores/arduino/CDC.cpp
@@ -236,9 +236,11 @@ Serial_::operator bool() {
 
 unsigned long Serial_::baud() {
 	// Disable interrupts while reading a multi-byte value
+	uint32_t baudrate;
 	ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
-		return _usbLineInfo.dwDTERate;
+		baudrate =  _usbLineInfo.dwDTERate;
 	}
+	return baudrate;
 }
 
 uint8_t Serial_::stopbits() {
--- a/hardware/arduino/avr/cores/arduino/Tone.cpp
+++ b/hardware/arduino/avr/cores/arduino/Tone.cpp
@@ -209,7 +209,7 @@ static int8_t toneBegin(uint8_t _pin)
         #if defined(WGM42)
           bitWrite(TCCR4B, WGM42, 1);
         #elif defined(CS43)
-          #warning this may not be correct
+          // TODO this may not be correct
           // atmega32u4
           bitWrite(TCCR4B, CS43, 1);
         #endif
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -372,8 +372,8 @@ bool ClassInterfaceRequest(USBSetup& setup)
 	return false;
 }
 
-int _cmark;
-int _cend;
+static int _cmark;
+static int _cend;
 void InitControl(int end)
 {
 	SetEP(0);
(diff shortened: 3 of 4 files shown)


facchinm:
@ArduinoBot build this please

NicoHood:
May we also remove the Timer warnings?
`#warning Timer 2 not finished (may not be present on this CPU)`

matthijskooijman:
Changes look good to me. Ideally, the warning fix and the static changes would have been a separate commit (or does static fix a warning too?), but that's not really a big deal. However, you should mention the static changes in the commit message, and mention the warnings that you're actually fixing (since they're not so obvious, I think).

> cores\arduino\CDC.cpp:242:1: warning: control reaches end of non-void function [-Wreturn-type]

Weird, I would think `ATOMIC_BLOCK` is actually intended to support exactly this usecase. Apparently the compiler isn't smart enough to figure out that the loop that `ATOMIC_BLOCK` expands to always executes at least once. Oh well, decent fix, I think.

> May we also remove the Timer warnings?
> And in Tone.cpp

Sure, go ahead! I think there might already be a github issue or even PR for the timer warnings>

NicoHood:
Isnt this warning not needed anyways?
https://github.com/NicoHood/Arduino/blob/patch-9/hardware/arduino/avr/cores/arduino/wiring.c#L307

The definition was set above here:
https://github.com/NicoHood/Arduino/blob/patch-9/hardware/arduino/avr/cores/arduino/wiring.c#L297
Attiny 85 uses this timer for example.

Maybe someone with more timer knowledge can tell us if this timer setup is finished or not. And if not, maybe he can finish it.

matthijskooijman:
@NicoHood your changes look good. You should probably remove the commented code entirely in the first new comment, leaving the TODO in the second one might be useful.

As for the wiring.c warning, I agree that the warning for TCCR1 can be removed, since it is already initialized above. In general, the timer initialization is a bit messy I guess, ideally it would only initialize timers actually used in the variants file, but that will be tricky to get right.

NicoHood:
I would leave the commented code to see that not all MCUs actually have this Timer and maybe for further debugging. It wont hurt.

This should remove all warnings now.

cmaglie:
Rebased and merged, thanks!

NicoHood:
With a 32u2 board I get:
```
Arduino/Arduino/build/linux/work/hardware/arduino/avr/cores/arduino/Tone.cpp:538:5: warning: 'TIMER2_COMPA_vect' appears to be a misspelled signal handler [enabled by default]
 ISR(TIMER2_COMPA_vect)
```

No idea what this means

matthijskooijman:
I suspect it means that 32u2 does not have a TIMER2_COMPA ISR, but the code defines it anway. The definition of this ISR should be protected by the proper guards, then (rather, I guess the Tone library should be configured for the 32u2 chip, to select a different timer on that MCU).

Issue #3745, opened 2015-08-30
HardwareSerial library sometimes sends the same data twice


johnholman:
Running a Arduino Mega 2560 program that makes heavy use of interrupts and that reports by sending lines of text (about 150 characters) in the main loop via the serial port at 115200 baud, I noiticed that additional spurious characters were sent on average every 2000 lines or so. Looking at the problem in more detail, what was transmitted was a line of exactly 66 characters consisting of the first few characters of the new line followed by last part of the line that had already been sent. The next line missed the first two characters, and then following ones were normal until the next occurrence.

These symptoms seemed consistent with void HardwareSerial::_tx_udr_empty_irq(void) occasionally being invoked at a time when the transmit buffer is in fact empty (i.e. _tx_buffer_head == _tx_buffer_tail), even though this should never happen. So I added some instrumentation:

void HardwareSerial::_tx_udr_empty_irq(void)
{
  // the buffer shouldn't be empty if we get here, but ...
  if (_tx_buffer_head == _tx_buffer_tail) {
    serial_errors++;

where serial_errors is defined in HardwareSerial.h as a private member

	volatile uint8_t serial_errors;

and added a method to read and print the error count as part of the main loop.

This confirmed that the problem is indeed associated with the ISR being entered at a time when the buffer is empty, as the error count increments by one each time the problem occurs.

My original code is quite complicated, but I've been able to replicate the problem with a simple testcase in which the loop prints as before while a timer interrupt runs every 0.5ms with an associated ISR running for 0.3ms. The error seems to occur more often at higher baud rates - at 500,000 baud it occurs every few seconds. I can provide the testcase if required.

As a workaround, I've added this section to the start of _tx_udr_empty_irq:

void HardwareSerial::_tx_udr_empty_irq(void)
{
  // the buffer shouldn't be empty if we get here, but ...
  if (_tx_buffer_head == _tx_buffer_tail) {
    serial_errors++;
    // empty buffer so the interrupt should be disabled. Make it so and return.
    cbi(*_ucsrb, UDRIE0);
	return;
  }
...

This fix seems to work for me, but I'm certainly not convinced that it is correct in all circumstances.

[...]

PR #3865, opened 2015-09-25
Fix Hardwareserial sending a byte twice

NicoHood:
https://github.com/arduino/Arduino/issues/3745

This also solves a flush() lockup.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
@@ -218,8 +218,11 @@ size_t HardwareSerial::write(uint8_t c)
   // significantly improve the effective datarate at high (>
   // 500kbit/s) bitrates, where interrupt overhead becomes a slowdown.
   if (_tx_buffer_head == _tx_buffer_tail && bit_is_set(*_ucsra, UDRE0)) {
+    uint8_t oldSREG = SREG;
+    cli();
     *_udr = c;
     sbi(*_ucsra, TXC0);
+    SREG = oldSREG;
     return 1;
   }
   tx_buffer_index_t i = (_tx_buffer_head + 1) % SERIAL_TX_BUFFER_SIZE;
@@ -240,9 +243,12 @@ size_t HardwareSerial::write(uint8_t c)
   }
 
   _tx_buffer[_tx_buffer_head] = c;
+  uint8_t oldSREG = SREG;
+  cli();
   _tx_buffer_head = i;
 	
   sbi(*_ucsrb, UDRIE0);
+  SREG = oldSREG;
   
   return 1;
 }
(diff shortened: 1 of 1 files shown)


matthijskooijman:
This code still has the `sbi()` for clearing `TXC0`, which is incorrect in theory (though we haven't properly deduced if it actually causes problems in practice too, and in theory making it atomic also prevents issues).

Also, perhaps you could use `ATOMIC_BLOCK`? That makes things more readable and should be used for new code IMHO.

NicoHood:
Added, also applied for the one left cli() that was there before.

matthijskooijman:
Looking at this PR, it actually seems like it applies good fixes. See https://github.com/arduino/Arduino/issues/3745#issuecomment-331627824 for a summary of the related problems.

This PR makes the following changes:
 1. Replace cli()/SREG with `ATOMIC_BLOCK`
 2. Makes `_tx_buffer_head == _tx_buffer_tail` in `write()` atomic.
 3. Makes `UDR` write and `TXC` clear atomic.
 4. Makes `_tx_buffer_head` and `UDRIE` writes atomic.

All of which I think are useful and relevant. However, I think this PR could still be improved:
 - By putting the `ATOMIC_BLOCK` commit first, so the rest  can just use `ATOMIC_BLOCK` from the start.
 - By, instead of using `ATOMIC_BLOCK` directly for the tx buffer head/tail accesses, introduce a `TX_BUFFER_ATOMIC` macro or similar. When the TX buffer is <= 256 bytes, this macro would just be empty, otherwise it would resolve to `ATOMIC_BLOCK(ATOMIC_RESTORESTATE). Doing so removes the need for the `#if (SERIAL_TX_BUFFER_SIZE>256)` checks all over the code and improves readability.
 - By fixing the way `TXC0` is set (e.g. do not use `sbi()`, see issue 1 described at https://github.com/arduino/Arduino/issues/3745#issuecomment-331627824).
 - By splitting 9aa902ef9d272e12306eb9ff7ffae4cdf47c6227 into two separate commits, and describing the problem in a bit more detail in the commit message.

@NicoHood, would you be interested in making these changes?

NicoHood:
@matthijskooijman I've granted arduino maintainers access to my patch branch via github. I currently have no time to do and test the changes. Feel free to modify or do a new PR, the important stuff is already there (how to fix the issue).

johnholman:
Are either of you working on this? If not I'm happy to make a new PR based on @NicoHood's work and incorporating changes suggested by @matthijskooijman.

@matthijskooijman I've defined the TX_BUFFER_ATOMIC macro in HardwareSerial.h, but would you prefer it in HardwareSerial.cpp? Also is the include of <util/atomic.h> needed for ATOMIC_BLOCK better kept local in HardwareSerial.h or in Arduino.h to support use elsewhere?

matthijskooijman:
@johnholman, I'm not finding the time for this, so I'm not working on this. If you would, that would be great.

As for the `TX_BUFFER_ATOMIC` macro, if it's only used inside the .cpp file, I'd define it in the .cpp file, to prevent polluting the global scope of other source files. The same goes for the `util/atomic.h` include.

johnholman:
@matthijskooijman, @NicoHood I've created pull request #6855 for this.

matthijskooijman:
I'm closing this one in favor of #6855.

PR #3882, opened 2015-09-28
Fixed typo in Keyboard API

NicoHood:


NicoHood:
Beside this typo you could consider to add UK keyboard support via a #define:
https://github.com/NicoHood/HID/commit/b28d83b1601a0566844bc884131ceff6b886b30e

(German etc is not possible since the API does not support ALT GR ascii lookup tables)

PR #3883, opened 2015-09-28
USB: fixed compiler warning "control reach end of non-void function"

cmaglie:
If we enable all warnings, for some reason the compiler is unable to detect the atomic block as always-executed and produce a "control reach end of non-void function".

The solution is to cache the result out of the atomic block where the dataflow analisys is handled correctly.

/cc @matthijskooijman


cmaglie:
Duplicate of #3854

Issue #3874, opened 2015-09-27
PUSBCallbacks - is it necessary to require static functions? (PluggableUSB in 1.6.6)


Teiwaz83:
The PUSBCallbacks struct (in PluggableUSB.h) is defined like so:
```
typedef struct __attribute__((packed))
{
  bool (*setup)(USBSetup& setup, u8 i);
  int (*getInterface)(u8* interfaceNum);
  int (*getDescriptor)(int8_t t);
  int8_t numEndpoints;
  int8_t numInterfaces;
  uint8_t *endpointType;
} PUSBCallbacks;
```
This requires the methods setup, getInterface, and getDescriptor to all be static. I haven't dug very far past this (the whole point of PluggableUSB is to keep me as far form the inner horrors of USB as possible!) - but is this necessary? Requiring these methods to be static really complicates setting up classes which can manage their own interfaces and endpoints. If you want each instance of a class to have its own interface/endpoint to make Windows see it as a separate device, you need to be using unique information when building the interface descriptor, which generally you want to be doing in getInterface. GetInterface having to be static prevents you from easily correlating a created interface with a specific instance of your class.

matthijskooijman:
IIUC you're saying that these callbacks should be method pointers instead of function pointers? The downside of that is that you will need to create a full object whenever you want to plug an USB backend, even if you would just have one instance anyway.

An alternative could be to let these callback functions accept a `void *userdata` parameter, that you then store as a new member inside the PUSBCallbacks. The function called can then cast the userdata pointer to the right object type and call the method you need. This is a common way to solve this problem (and the userdata can also be used for other things, like an integer if that's all you need to distinguish instances, or a pointer to a struct with some parameters).

Instead of passing the userdata, you could also pass a pointer to the `PUSBCallbacks` struct, which might be useful in some cases.

Adding such a userdata parameter to the callbacks will break the API, but I guess the pluggable USB API is new enough to still allow some changes.

@facchinm, @NicoHood, IIRC you were working on this pluggable USB stuff, right? Care to have a look?

Teiwaz83:
I was thinking that they should just be method pointers instead of function pointers as you said first, since it's fairly trivial to redirect to a function if it calls a method, but your userdata pointer parameter seems like a better option and would be potentially useful for other things (like passing instance, endpoint, and report IDs directly).

NicoHood:
I tried this here, but I cant get it working
[...]

PR #3884, opened 2015-09-28
[HID] Another small cleanup

cmaglie:
Applying some of the suggestion from @NicoHood and other clean-up as well.

  * HID.SendReport now use an unsigned parameter for lenght
  * Removed all weird `u8` datatype (from HID only, USBCore not yet)
  * Removed unused PUSBReturn structure (is that used somewhere?)
  * Some other minor fix

/cc @facchinm @NicoHood

### DIFF
--- a/hardware/arduino/avr/libraries/HID/HID.cpp
+++ b/hardware/arduino/avr/libraries/HID/HID.cpp
@@ -152,6 +152,7 @@ HID_::HID_(void)
 
 int HID_::begin(void)
 {
+	return 0;
 }
 
 #endif /* if defined(USBCON) */
--- a/hardware/arduino/avr/libraries/HID/HID.cpp
+++ b/hardware/arduino/avr/libraries/HID/HID.cpp
@@ -85,7 +85,7 @@ void HID_::AppendDescriptor(HIDDescriptorListNode *node)
 	sizeof_hidReportDescriptor += (uint16_t)node->length;
 }
 
-void HID_::SendReport(u8 id, const void* data, int len)
+void HID_::SendReport(uint8_t id, const void* data, const uint16_t len)
 {
 	USB_Send(HID_TX, &id, 1);
 	USB_Send(HID_TX | TRANSFER_RELEASE,data,len);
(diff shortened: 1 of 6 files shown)


hardware/arduino/avr/libraries/HID/HID.cpp
```diff
@@ -85,19 +85,19 @@ void HID_::AppendDescriptor(HIDDescriptorListNode *node)
 	sizeof_hidReportDescriptor += (uint16_t)node->length;
 }
 
-void HID_::SendReport(u8 id, const void* data, int len)
+void HID_::SendReport(uint8_t id, const void* data, const uint16_t len)
```
NicoHood: This should be int or you should change all underlying APIs as well. I know that this is sometimes good or bad but in this case we should keep the int. 32k bytes is far enough anyways.

hardware/arduino/avr/libraries/HID/HID.h
```diff
@@ -58,28 +58,28 @@ class HID_
 public:
   HID_(void);
   int begin(void);
-  void SendReport(uint8_t id, const void* data, int len);
+  void SendReport(uint8_t id, const void* data, const uint16_t len);
```
NicoHood: int here as well

NicoHood:
Looks good to me except the int vs uin16_t.

cmaglie:
> This should be int or you should change all underlying APIs as well. I know that this is sometimes good or bad but in this case we should keep the int. 32k bytes is far enough anyways.

my preference was to change it, but you're right, it's not worth doing it now.

I've rebased and merged everything except int -> uint16

NicoHood:
Maybe we will change this stuff later on but I dont think so.

Could you please try to get this working?
https://github.com/arduino/Arduino/issues/3874#issuecomment-143685335

Otherwise I really need a pluggable USB API change to get stuff working efficiently.

Similar to your last PR I'd merge those two also together:
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/PluggableUSB.h#L28-L43

But before you do that I need to know if the code issue above can be solved. If not I'd like to change the Pluggable USB a bit, similar to my HID change.

PR #3892, opened 2015-09-29
Removed not used PUSB_Begin()

NicoHood:


cmaglie:
I've removed it also from SAM. Thanks.

PR #3893, opened 2015-09-29
Fix wrong buffer read from Pluggable USB

NicoHood:
I just stumbled about this. I havent tested this, since I am reqriting the Pluggable USB.
And in my rewrite I replaced the useless pointer to a real uint8_t and noticed that this reads out of bounds, since the endpoint will always be 1 byte array and not longer.

This worked so far since Serial is not pluggable and HID the only device used.

This PR is more though as note for you, because I try to rewrite the API itself to make it even more pluggable.


NicoHood:
I was wrong, sorry

NicoHood:
However those names and the function itself is totally confusing. there is probably still something wrong with it

PR #3894, opened 2015-09-29
Implement Wire.end() for SAM core

sandeepmistry:
For the SAM side of #613.

Also moved common reset and disable code in libsam ``TWI_ConfigureMaster`` and ``TWI_ConfigureSlave`` functions to new ``TWI_Disable`` function, which is used in ``TwoWire::end()``

The ``TwoWire`` constructor now has a new end callback parameter that is used to undo any actions that were done in the begin callback. 

Updated version of https://github.com/arduino/Arduino/pull/3730


hardware/arduino/sam/libraries/Wire/Wire.cpp
```diff
@@ -408,7 +433,21 @@ static void Wire1_Init(void) {
...
-TwoWire Wire1 = TwoWire(WIRE1_INTERFACE, Wire1_Init);
+static void Wire1_Deinit(void) {
+	NVIC_DisableIRQ(WIRE1_ISR_ID);
+	NVIC_ClearPendingIRQ(WIRE1_ISR_ID);
+
+	pmc_disable_periph_clk(WIRE1_INTERFACE_ID);
+
+	// disable pull ups
+	pinMode(PIN_WIRE1_SDA, OUTPUT);
+	pinMode(PIN_WIRE1_SCL, OUTPUT);
```
cmaglie: Mhhh, why OUTPUT? I'm expecting to see them set as INPUT

hardware/arduino/sam/system/libsam/source/twi.c
```diff
@@ -179,6 +166,27 @@ void TWI_ConfigureSlave(Twi *pTwi, uint8_t slaveAddress)
...
+    assert( pTwi ) ;
+
+    uint32_t i;
+
+    /* TWI software reset */
+    pTwi->TWI_CR = TWI_CR_SWRST;
+    pTwi->TWI_RHR;
+
+    /* Wait at least 10 ms */
+    for (i=0; i < 1000000; i++);
```
cmaglie: Is this loop really needed?
Probably the compiler is optimizing it away...

sandeepmistry: @cmaglie I would agree. When I try the following sketch, ``millis`` does not increase at all.

```arduino
#include <Wire.h>

void setup() {
  unsigned long start;
  unsigned long afterBegin;
  unsigned long afterEnd;
  
  Serial.begin(9600);

  start = millis();
  Wire.begin(42);
  afterBegin = millis();
  Wire.end();
  afterEnd = millis();

  Serial.println(start);
  Serial.println(afterBegin);
  Serial.println(afterEnd);
}

void loop() {
  unsigned long l = millis();

  Serial.println(l);
  delay(1000);              // wait for a second
}
```

Output was:
```
2
2
2
2
1002
2002
...
```

There are two "delay" loops which are run, this one and one in ```TWI_ConfigureSlave```.

hardware/arduino/sam/libraries/Wire/Wire.cpp
sandeepmistry: @cmaglie upon further investigation, this code isn't needed, as the ``PIO_Configure`` calls  in the ```Wire1_Init``` do not enable pull ups. I've updated this in: https://github.com/sandeepmistry/Arduino/commit/dd7e90802e960904f281c8fb6ac24b7fae29e863

I was confused my the external SDA0 and SCL0 pull ups on the Due.

PR #3896, opened 2015-09-29
[PluggableHID] 2nd round of API improvement (WIP)

cmaglie:
This is a follow up to #3840, an attempt to make the PluggableHID API more flexible before the release.

@NicoHood, please, try to find 10 minutes to look at the commits one at a time. From what I can see they "match" your latest attempts very closely.

This is implemented on AVR only for now (SAM and SAMD will follow).

### DIFF
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -23,9 +23,6 @@
 #if defined(USBCON)	
 #ifdef PLUGGABLE_USB_ENABLED
 
-// TODO: set correct value for different CPUs
-#define MAX_EP 6
-
 extern uint8_t _initEndpoints[];
 
 PluggableUSB_ PluggableUSB;
@@ -64,7 +61,7 @@ bool PluggableUSB_::setup(USBSetup& setup, uint8_t j)
 
 bool PluggableUSB_::plug(PUSBListNode *node)
 {
-	if ((lastEp + node->numEndpoints) >= MAX_EP) {
+	if ((lastEp + node->numEndpoints) > USB_ENDPOINTS) {
 		return false;
 	}
 
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -25,8 +25,6 @@
 
 extern uint8_t _initEndpoints[];
 
-PluggableUSB_ PluggableUSB;
-
 int PluggableUSB_::getInterface(uint8_t* interfaceNum)
 {
 	int sent = 0;
@@ -88,6 +86,12 @@ bool PluggableUSB_::plug(PUSBListNode *node)
 	}
 	return true;
 	// restart USB layer???
+}
+
+PluggableUSB_& PluggableUSB()
+{
+	static PluggableUSB_ obj;
+	return obj;
 }
 
 PluggableUSB_::PluggableUSB_() : lastIf(CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT),
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -25,21 +25,21 @@
 
 #define MAX_MODULES	6
 
-static u8 lastIf = CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT;
-static u8 lastEp = CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT;
+static uint8_t lastIf = CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT;
+static uint8_t lastEp = CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT;
 
-extern u8 _initEndpoints[];
+extern uint8_t _initEndpoints[];
 
 //PUSBCallbacks cbs[MAX_MODULES];
-static u8 modules_count = 0;
+static uint8_t modules_count = 0;
 
 static PUSBListNode* rootNode = NULL;
 
-int PluggableUSB_::getInterface(u8* interfaceNum)
+int PluggableUSB_::getInterface(uint8_t* interfaceNum)
 {
 	int ret = 0;
 	PUSBListNode* node = rootNode;
-	for (u8 i=0; i<modules_count; i++) {
+	for (uint8_t i=0; i<modules_count; i++) {
 		ret = node->getInterface(interfaceNum);
 		node = node->next;
 	}
@@ -50,18 +50,18 @@ int PluggableUSB_::getDescriptor(int8_t t)
 {
 	int ret = 0;
 	PUSBListNode* node = rootNode;
-	for (u8 i=0; i<modules_count && ret == 0; i++) {
+	for (uint8_t i=0; i<modules_count && ret == 0; i++) {
 		ret = node->getDescriptor(t);
 		node = node->next;
 	}
 	return ret;
 }
 
-bool PluggableUSB_::setup(USBSetup& setup, u8 j)
+bool PluggableUSB_::setup(USBSetup& setup, uint8_t j)
 {
 	bool ret = false;
 	PUSBListNode* node = rootNode;
-	for (u8 i=0; i<modules_count && ret == false; i++) {
+	for (uint8_t i=0; i<modules_count && ret == false; i++) {
 		ret = node->setup(setup, j);
 		node = node->next;
 	}
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -23,58 +23,56 @@
 #if defined(USBCON)	
 #ifdef PLUGGABLE_USB_ENABLED
 
-#define MAX_MODULES	6
+// TODO: set correct value for different CPUs
+#define MAX_EP 6
 
 extern uint8_t _initEndpoints[];
-
-//PUSBCallbacks cbs[MAX_MODULES];
 
 PluggableUSB_ PluggableUSB;
 
 int PluggableUSB_::getInterface(uint8_t* interfaceNum)
 {
 	int ret = 0;
-	PUSBListNode* node = rootNode;
-	for (uint8_t i=0; i<modulesCount; i++) {
+	PUSBListNode* node;
+	for (node = rootNode; node; node = node->next) {
 		ret = node->getInterface(interfaceNum);
-		node = node->next;
 	}
 	return ret;
 }
 
 int PluggableUSB_::getDescriptor(int8_t t)
 {
-	int ret = 0;
-	PUSBListNode* node = rootNode;
-	for (uint8_t i=0; i<modulesCount && ret == 0; i++) {
-		ret = node->getDescriptor(t);
-		node = node->next;
+	PUSBListNode* node;
+	for (node = rootNode; node; node = node->next) {
+		int ret = node->getDescriptor(t);
+		if (ret)
+			return ret;
 	}
-	return ret;
+	return 0;
 }
 
 bool PluggableUSB_::setup(USBSetup& setup, uint8_t j)
 {
-	bool ret = false;
-	PUSBListNode* node = rootNode;
-	for (uint8_t i=0; i<modulesCount && ret == false; i++) {
-		ret = node->setup(setup, j);
-		node = node->next;
+	PUSBListNode* node;
+	for (node = rootNode; node; node = node->next) {
+		if (node->setup(setup, j)) {
+			return true;
+		}
 	}
-	return ret;
+	return false;
 }
 
 bool PluggableUSB_::plug(PUSBListNode *node)
 {
-	if (modulesCount >= MAX_MODULES) {
+	if ((lastEp + node->numEndpoints) >= MAX_EP) {
 		return false;
 	}
 
-	if (modulesCount == 0) {
+	if (!rootNode) {
 		rootNode = node;
 	} else {
 		PUSBListNode *current = rootNode;
-		while(current->next != NULL) {
+		while (current->next) {
 			current = current->next;
 		}
 		current->next = node;
@@ -83,18 +81,17 @@ bool PluggableUSB_::plug(PUSBListNode *node)
 	node->pluggedInterface = lastIf;
 	node->pluggedEndpoint = lastEp;
 	lastIf += node->numInterfaces;
-	for (uint8_t i=0; i<node->numEndpoints; i++) {
+	for (uint8_t i = 0; i < node->numEndpoints; i++) {
 		_initEndpoints[lastEp] = node->endpointType[i];
 		lastEp++;
 	}
-	modulesCount++;
 	return true;
 	// restart USB layer???
 }
 
 PluggableUSB_::PluggableUSB_() : lastIf(CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT),
                                  lastEp(CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT),
-                                 modulesCount(0), rootNode(NULL)
+                                 rootNode(NULL)
 {
 	// Empty
 }
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -25,12 +25,12 @@
 
 extern uint8_t _initEndpoints[];
 
-int PluggableUSB_::getInterface(uint8_t* interfaceNum)
+int PluggableUSB_::getInterface(uint8_t* interfaceCount)
 {
 	int sent = 0;
 	PUSBListNode* node;
 	for (node = rootNode; node; node = node->next) {
-		int res = node->getInterface(interfaceNum);
+		int res = node->getInterface(interfaceCount);
 		if (res < 0)
 			return -1;
 		sent += res;
@@ -50,11 +50,11 @@ int PluggableUSB_::getDescriptor(int8_t type)
 	return 0;
 }
 
-bool PluggableUSB_::setup(USBSetup& setup, uint8_t j)
+bool PluggableUSB_::setup(USBSetup& setup, uint8_t interfaceNum)
 {
 	PUSBListNode* node;
 	for (node = rootNode; node; node = node->next) {
-		if (node->setup(setup, j)) {
+		if (node->setup(setup, interfaceNum)) {
 			return true;
 		}
 	}
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -31,7 +31,7 @@ int PluggableUSB_::getInterface(uint8_t* interfaceNum)
 	PUSBListNode* node;
 	for (node = rootNode; node; node = node->next) {
 		int res = node->getInterface(interfaceNum);
-		if (res == -1)
+		if (res < 0)
 			return -1;
 		sent += res;
 	}
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -68,10 +68,10 @@ bool PluggableUSB_::setup(USBSetup& setup, u8 j)
 	return ret;
 }
 
-int8_t PluggableUSB_::addFunction(PUSBListNode *node, u8* interface)
+bool PluggableUSB_::plug(PUSBListNode *node)
 {
 	if (modules_count >= MAX_MODULES) {
-		return 0;
+		return false;
 	}
 
 	if (modules_count == 0) {
@@ -84,14 +84,15 @@ int8_t PluggableUSB_::addFunction(PUSBListNode *node, u8* interface)
 		current->next = node;
 	}
 
-	*interface = lastIf;
+	node->pluggedInterface = lastIf;
+	node->pluggedEndpoint = lastEp;
 	lastIf += node->numInterfaces;
-	for ( u8 i = 0; i< node->numEndpoints; i++) {
+	for (uint8_t i=0; i<node->numEndpoints; i++) {
 		_initEndpoints[lastEp] = node->endpointType[i];
 		lastEp++;
 	}
 	modules_count++;
-	return lastEp - node->numEndpoints;
+	return true;
 	// restart USB layer???
 }
 
--- a/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
+++ b/hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
@@ -40,7 +40,7 @@ int PUSB_GetInterface(u8* interfaceNum)
 	int ret = 0;
 	PUSBListNode* node = rootNode;
 	for (u8 i=0; i<modules_count; i++) {
-		ret = node->cb->getInterface(interfaceNum);
+		ret = node->getInterface(interfaceNum);
 		node = node->next;
 	}
 	return ret;
@@ -51,7 +51,7 @@ int PUSB_GetDescriptor(int8_t t)
 	int ret = 0;
 	PUSBListNode* node = rootNode;
 	for (u8 i=0; i<modules_count && ret == 0; i++) {
-		ret = node->cb->getDescriptor(t);
+		ret = node->getDescriptor(t);
 		node = node->next;
 	}
 	return ret;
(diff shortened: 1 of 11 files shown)


NicoHood:
You were right. Your commits all look good to me. I will compare with my version and look if this can be used like this. But its looking very good to me. Sorry that i missed that first.

Have you tested this with real hardware?

afk for some time, excuse me.

cmaglie:
I haven't tested yet (all those changes should be "equivalent" but... only testing will tell).

> afk for some time, excuse me.

np, I'm going away too, will continue tomorrow.

NicoHood:
* The code works (tested KeyboardSerial example)
* The u8 should be replaced later on all places, also the typedefs
* There are some compiler warnings that needs to be solved.
* I will try yo fork your repo and post some fixes over that now. The API looks better to me now.
* Other stuff might be edited later on.

cmaglie:
> The code works (tested KeyboardSerial example)

great, thanks for testing!

> The u8 should be replaced later on all places, also the typedefs

I'm going to apply this change.

> There are some compiler warnings that needs to be solved.

I'll leave this one as the last task

> Why dont we make the 3 functions virtual. I think the virtual overhead is then the same but the same function can be used more flexible?

Yes

> numEndpoints should be const and declared via constructor. Without its more flexible but can contain more errors if they are not initialized.

Yes

> why is interface unsigned and endpoint signed?

because I've just moved code around and they was defined like that. I'll check if everything can be changed to unsigned.

> why do we need those wrapper function for the IF and EP? Doesnt inheriting them via protected work good enough?

Well the reasoning is that IF and EP are assigned *after* the module has been plugged, they are not structure-like that are assigned directly from the user. Moreover if we make the other fields constant and initialized through constructor we are going to remove this difference. Let's see how it looks I'm going to add more commits in the next few hours.

NicoHood:
I forgot: You can get rid of the static functions and variables now. Since the Class is now a singleton there is no need for static functions and variables.

I tried to fix that in a local branch, but before editing twice (you and me) I suggest you just add this.
Please initialize the variables via a proper constructor
Class::Class : var(defaultvalue), nextval(default) etc. You know the drill ;D

So I wait for further commits from you. Let me know on any update. 
This is getting better and better now :)

cmaglie:
Ok, just pushed another 4 commits, this should solve 90% of the problems so far.

Still to be done:

* check unsigned vs signed
* fix compiler warnings
* rename fields/class to a more appropriate/comfortable name

hardware/arduino/avr/cores/arduino/PluggableUSB.h
```diff
@@ -25,30 +25,42 @@
...
+  friend class PluggableUSB_;
+};
 
-int PUSB_GetDescriptor(int8_t t);
+class PluggableUSB_ {
+public:
+  static bool plug(PUSBListNode *node);
+  static int getInterface(uint8_t* interfaceNum);
+  static int getDescriptor(int8_t t);
+  static bool setup(USBSetup& setup, uint8_t i);
```
NicoHood: Those static functions can be removed as well I think.

hardware/arduino/avr/cores/arduino/PluggableUSB.cpp
```diff
@@ -25,53 +25,53 @@
 
 #define MAX_MODULES	6
 
-static u8 lastIf = CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT;
-static u8 lastEp = CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT;
+static uint8_t lastIf = CDC_ACM_INTERFACE + CDC_INTERFACE_COUNT;
+static uint8_t lastEp = CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT;
```
NicoHood: Those variables can be moved inside the pluggable class, without being static.

hardware/arduino/avr/libraries/HID/HID.cpp
```diff
@@ -130,24 +110,16 @@ bool HID_Setup(USBSetup& setup, uint8_t i)
...
+// XXX: I've found no way to pass literal value directly in
+//      the PUSBListNode constructor
+static uint8_t epTypeDef[] = { EP_TYPE_INTERRUPT_IN };
 
-	HID_ENDPOINT_INT = PUSB_AddFunction(&node, &HID_INTERFACE);
+HID_::HID_(void) : PUSBListNode(1, 1, epTypeDef),
+                   rootNode(NULL), sizeof_hidReportDescriptor(0),
+                   modules_count(0), protocol(1), idle(1)
+{
+	// XXX: Shall this be done in PUSBListNode(...) constructor?
```
NicoHood: No, this should not be done in the constructor. That's how I tried it before.

The problem is, that epTypeDef is a pointer/array. When you inherit the PUSBListNode and call the plug() function at the same moment it initializes the endpoint with this array value. But if the array is not initialized before, it just ignores its content. Because of this reason we need to manually call the add function ourselves.

So the problem is to initialize an array inside a constructor.
There was another possible workaround with a static array length member etc but I think this all makes it more complicated.

If we miss this feature and only accept one endpoint this could reduce the flexibility (I think Midi needs 1 Interface + 2 endpoints).

My suggestion:
Add an array with size 1 to the HID class and initzialize it inside the constructor function, before the plug() function. That should work.

NicoHood:
you could make those int8_t and use the interface as a check if the modul exists. Maybe this check can be done in the pluggable class so we have less code duplication/error with bad library coding. (set to -1 if all modules are used)

And we also need to change the modules used limit to the endpoint number, not the modules itself. On a 16u2 there is only one more free endpoint. If your interface (midi usb I suspect) uses 2 endpoints it is not able to work correctly. Thatswhy the check should be done via endpoint count and set the interface to -1 if it doesnt work correct.

If we implement the -1 inside the pluggable class we can remove the bool return value of plug() otherwise we could use this to set the interface to -1.

The 16u2 needs to be patched with the modul/ep limit anyways, but later.

cmaglie:
Pushed other 2 commits.

> And we also need to change the modules used limit to the endpoint number, not the modules itself. On a 16u2 there is only one more free endpoint. If your interface (midi usb I suspect) uses 2 endpoints it is not able to work correctly. Thatswhy the check should be done via endpoint count and set the interface to -1 if it doesnt work correct.

this means that the check:

```C++
bool PluggableUSB_::plug(PUSBListNode *node)
{
	if (modulesCount >= MAX_MODULES) {
		return false;
	}
```

should really be:

```C++
bool PluggableUSB_::plug(PUSBListNode *node)
{
	if ((lastEp + node->node->numEndpoints) >= MAX_MODULES) {
		return false;
	}
```

and `MAX_MODULES` should be renamed to `MAX_EP`, right? In this case I think that we can completely drop `modulesCount`

NicoHood:
Yes.

You need to define the value then with 6 (since 3 are used by the CDC which is initialized to `lastep` already.

For the u2 Series please add 4 enpoints.

You also need to edit this file:
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/USBCore.cpp#L320-L323

Or (maybe even better) you can use `sizeof(_initEndpoints)` (minus one for the control ep) and define the MAX_EP somewhere in USBCore.h

To initialize `_initEndpoints` to zero i'd use something similar to the PinInterrupt PR (if you remember).

Please use `#ifndef MAX_EP` and then define the values for 32u4 and u2 series.

cmaglie:
Wait wait wait, the `_initEndpoints[]` array has `7` elements not `6`, so `MAX_EP` should be `7`.

From USBCore:
```C++
u8 _initEndpoints[] =
{
        0,

        EP_TYPE_INTERRUPT_IN,           // CDC_ENDPOINT_ACM
        EP_TYPE_BULK_OUT,                       // CDC_ENDPOINT_OUT
        EP_TYPE_BULK_IN,                        // CDC_ENDPOINT_IN

#ifdef PLUGGABLE_USB_ENABLED
        //allocate 3 endpoints and remove const so they can be changed by the user
        0,
        0,
        0,
#endif
};
```

in PluggableUSB lastEp is set to `CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT`where

```
#define CDC_FIRST_ENDPOINT	1
#define CDC_ENPOINT_COUNT	3
```

so lastEp is `4`. With `3` extra endpoint it goes to `7`.
BTW I guess that I'll follow your suggestion to use sizeof.

NicoHood:
@matthijskooijman HID will be initialzed multiple times. At least if I need to do a fork. This is needed to create multiple gamepads on multiple endpoints (and other cool stuff). I am just wondering about what variable you are talking. Whatever variable this is, if the first adding works (enough endpoints) and the 2nd doesnt, with a shared variable this can cause problems.

Pluggable USB probably does not need to be a singleton/class at all. So we could maybe change this back. As you said the size should be compared.

@cmaglie good catch with the 1 offset. sizeof is still the best solution i guess. Please conside to commend the first entry of `_initEndpoints` with `// Control Endpoint`

matthijskooijman:
I looked casually through all commits, looking good to me!

> To initialize _initEndpoints to zero i'd use something similar to the PinInterrupt PR (if you remember).

No need - global variables are zeroed by default (the attachInterrupt stuff need to initialize to `&empty` or something, not 0).

> Yes. We can get rid of the modules count variable and overhead with this check.

I was going to say that you still need the count to interate the modules, but you can of course just loop until `next` is `NULL`.

> in PluggableUSB lastEp is set to CDC_FIRST_ENDPOINT + CDC_ENPOINT_COUNT

I haven't looked at the code for this, but perhaps it would be worth considering to convert the CDC class to a pluggable class as well? This might simplify some of the code? Even if it is always enabled, the code structure might become more regular? OTOH it could add size to the compiled code, of course.

cmaglie:
Pushed other 3 commits.

Please have a closer review and also test if possible, because those commits have some code change too.

I'm done for today :-), be back tomorrow.

matthijskooijman:
I see only 2 commits? Those look good to me.

cmaglie:
Github orders the history of the pull request based on the commit timestamp (not the push timestamp), so you can find the commit if you reload the page and scroll back in the discussion, or maybe it's more simple to change to the "Commits" tab, btw this is the "missing" one:

https://github.com/cmaglie/Arduino/commit/ad368331bb975082690c31b22a10644647c26646

NicoHood:
> No need - global variables are zeroed by default (the attachInterrupt stuff need to initialize to &empty or something, not 0).
Yeah but we only want to initialize some of them. I just want to be sure the endpoint is not (wrong) configured if it shouldnt be at all.

> I was going to say that you still need the count to interate the modules, but you can of course just loop until next is NULL.
We need this check anyways so the modul count is really not needed (like in pluggable HID we also do not count the devices)

> I haven't looked at the code for this, but perhaps it would be worth considering to convert the CDC class to a pluggable class as well?
I was thinking of this as well. I also would like to see a whole deactivation of the USB-Core at all. This removes a lot of overhead for the 32u4 which can be used for non usb stuff very good. The u2 Series profits even more here.

matthijskooijman:
One more thing I realized wrt the static/non-static thing: AFAIU constructors for global objects are run in arbitrary order. This means that inside a constructor, other objects might not be fully initialized yet. This applies in particular to non-static members, which are always initialized in the constructor, whereas static members and global variables can be initialized by zeroing or copying the .data section (both of which happen *before* any constructors are called).

In this case, this can cause problems if a module is initialized before the PluggableUSB class, or a HID "submodule" is initialized before the main HID object. Not sure what the exact problems or solutions are, just wanted to throw this out there.

cmaglie:
omg @matthijskooijman you're right, we hit exactly in the static initialization order fiasco, totally missed that :-(

https://isocpp.org/wiki/faq/ctors#static-init-order

The FAQ propose also doing the initialization using lazy construction through an helper function, for example:

```C++
HID_& HID()
{
  static HID_* obj = new HID_();
  return *obj;
}
```

and replace every occurrence of HID with the call to HID() so:

```C++
HID.SendReport(.....);
```

becomes

```C++
HID().SendReport(....);
```

Very ugly :-( don't know if there are other easier solutions.

NicoHood:
I got the idea mainly, but not the connection to this API. Is it more a problem to have static members or to not have static members. This is above my c++ knowledge at the moment.

You mean the problem is, if the HID.cpp tried to call the PUSB.plug() function if it does not exist yet? Wouldnt this give a compile error or so? May you give me a real example related to our API?

cmaglie:
Exactly that, and the compiler doesn't give any warning.

The problem is that static objects are constructed at the very beginning of the sketch (even before main()!) in an undefined order.

So the solution is to use a lazy construction like the one I've posted above, or to not call other (static) object's method from the constructor.

NicoHood:
A singleton HID_ HID is also a static object?
We have mainly removed the static functions, so isnt this good now?

cmaglie:
> A singleton HID_ HID is also a static object?

yes

NicoHood:
And what if we just add the hid this pointer to the root node linled list and call the init endpoint functions *that are now inside plug() with the pluggable constructor? Or there the same problem, whether all notes are added or not?

I have no real idea how to fix this. Is there no compiler attribute we can add?

matthijskooijman:
In general, I guess there are two ways to fix this problem:

 1. Make sure that the HID/PUSB constructors get called before the constructors that rely on them.
 2. Make sure that it is not a problem for HID/PUSB to be not initialized yet.

> You mean the problem is, if the HID.cpp tried to call the PUSB.plug() function if it does not exist yet? Wouldnt this give a compile error or so? May you give me a real example related to our API?

It's not so much that the `plug()` method doesn't exist yet, but the PUSB constructor might not have been called yet, so any non-static member variables (which reside in the memory pointed to by `this`) have random values still (and even if you just write to them - they will be overwritten when the PUSB constructor does run). I believe this means that your program won't blow up per se if you call a method on a non-initialized object, but things will break if that method uses non-static member variables (calling a virtual method *will* likely blow up, since I expect the vtable isn't set up yet). Also, I suspect that the C++ specification says it is undefined behaviour to call methods on non-initialized objects, so I guess that implementing 2. above, but also staying within the limits of the C++ spec might be impossible (without resorting to static methods / functions). Might be good to look closer at the spec to verify this.


>      static HID_* obj = new HID_();

This implements option 1 aboved. This uses dynamic memory, which I'd rather avoid. Using a direct object instead of a pointer should also work. See the next FAQ item in your link (it has problems with destruction, but that's not relevant for Arduino).

Still, using `HID()` everywhere is ugly, so I don't really like this approach.

> And what if we just add the hid this pointer to the root node linled list and call the init endpoint functions *that are now inside plug() with the pluggable constructor? Or there the same problem, whether all notes are added or not?

I don't actually think that `plug` calls any functions/methods? In general, it is a good idea to do as little as possible inside code that runs during static initialization, but in this case I think `plug()` only access its own non-static member variables (which is a problem, but reducing this to just accessing 1 non-static member variable, `rootNode`, won't really improve things).

What would help, is to make `rootNode` and `plug()` static again, so that `rootNode` can be zero-initialized and only a static method needs to be called, which is safe. This does really enforce single-ton-ness of the HID/PUSB classes, since there is only one static linked list, then (this implements option 2 above).

Another option for implementing 1 above might be const-initialization. In some cases, a C++11 compiler can evaluate constructors at compiletime, so at runtime it is just a matter of copying some bytes from the .data section into RAM, which is guaranteed to happen before any other constructors are run. Here's a good writeup about this: https://akrzemi1.wordpress.com/2012/05/27/constant-initialization/ One problem with this is is that it's impossible to really guarantee that const-initialization actually takes place - if the code is modified, it might no longer do this (though I believe it is not a matter of compiler optimization, the C++11 spec actually specifies when const-initialization *must* be applied). I described some ways to partially check that const-initialization happens here: http://www.stderr.nl/Blog/Software/Cpp/ConstantInitialization.html

A variant on this might be to use in-line member initializers, e.g. specify the initial value of a non-static member right where it is declared inside the `class { ... }`. I'm not sure if this allows (partial?) const-initialization as well, might be interesting to check.

I think that const-initialization might be our best bet here, since it is clean and non-obtrusive. In general, the Arduino code could benefit from allowing const-initialization too, shrinking the compiled code size by replacing constructors with some data in the .data section.

One caveat of const-initialization is that, of course, only constant expressions can be used in a `constexpr` constructor. Registering yourself as a PUSB handler using `plug()` will certainly not be a constant expression, meaning that the `HID()` constructor cannot be `constexpr`, which leaves this problem around for HID submodules. This could be fixed by splitting HID into a part that registers with PUSB (with a normal constructor) and a part that receives the registrations of HID submodules (with a `constexpr` constructor).

cmaglie:
Thanks @matthijskooijman, interesting read, I'll give it a try.

In the meantime I've pushed a solution with the lazy-construction trick, it doesn't look so bad in the end.

https://github.com/cmaglie/Arduino/commit/65b8430fece7a708be740089864835a862678cff

cmaglie:
@matthijskooijman, I've read about `constexpr`, very interesting stuff, but I'm not confident enough to propose a solution based on that right now...

I've tried to revert back the various `rootList` as static: https://github.com/cmaglie/Arduino/commit/07c7fdeb4a668c280ad5e0acb9d30a498dee8c5a
this solves the problem without the need to use `HID()` instead of `HID`.
The cons is that the code now is a bit more complex because it we need an `init()` function that does all the non-static initialization after the construction. Moreover this solution doesn't provide any significant gain in Flash/RAM usage compared to the other one based on the `HID()` function.

NicoHood:
I'd vote against the static HID class, since I want to inherit from this and creat multiple HID devices. (with a single module). This is required since not all HID features can be merged in a multireport and needs more than a single endpoint to work properly.
Maybe I need to dupe the code anyways but then I still do have the problem...

I have no idea to solve it though. But I will test the new build now.

facchinm:
Hi Nico, you can specify multiple endpoints in an alternative implementation of HID library; I am still convinced that that class should remain as simple as possible.
Anyway, `Mouse` and `Keyboard` can still be singletons because they are the last link in the chain so no one can call their methods before their constructor is executed

NicoHood:
Then the solution sounds reasonable.
After burning a new bootloader to my new hardware the HID/USB Core itself seems to work. Only tested the SerialKeyboard example real quick.

I looked at all commits of the PR anyways, looking good so far. This weekend I hope I can test what I can do this these updates now. Anything that is still on the todo list? Is this static thing now 100% fixed?

You can remove this public statement:
https://github.com/cmaglie/Arduino/blob/65b8430fece7a708be740089864835a862678cff/hardware/arduino/avr/cores/arduino/PluggableUSB.h#L49

NicoHood:
I found a new problem:
If we have multiple HID devices `int PluggableUSB_::getDescriptor(int8_t type)` will only return the very first HID device since the type is always the same. We need to pass the interface number here instead I guess.

Originally I was trying to remove the interface check to the Pluggable USB instead of the HID class to avoid code duplication, passing a variable and coding errors.

NicoHood:
Next problem I found is that the API is currently not supporting multiple interfaces per plug. I am not sure if this is really needed to have but there are two options:
* Remove all Interface counts etc that just take flash for nothing
* Make it fully pluggable

It seems that the CDC Serial uses 2 interfaces. Not sure if this can be realized with 2 plugs or better with a real multiinterface API. We should defenitely consider to make the CDC class pluggable as well. This would save us some bytes. Not sure if we still have the static problem here as well.

Once again to the static problem:
Good that we solved it like this, but what If a user wants to create a singleton for a plugged interface? Then the API would look uglz with the Singleton(). I think If you port MidiUSB now to this API it will look ugly. Its a real problem...

I also made some changes to the API (on top of this PR). Please have a look:
(will be edited)

cmaglie:
>  If we have multiple HID devices int PluggableUSB_::getDescriptor(int8_t type) will only return the very first HID device since the type is always the same. We need to pass the interface number here instead I guess.

This is an interesting one, looking at the sequence of calls starting from the ISR we have:

```C++
//	Endpoint 0 interrupt
ISR(USB_COM_vect)
{
[...]
	USBSetup setup;
	Recv((u8*)&setup,8);
[...]
	u8 requestType = setup.bmRequestType;
[...]
	if (REQUEST_STANDARD == (requestType & REQUEST_TYPE))
	{
		u8 r = setup.bRequest;
		u16 wValue = setup.wValueL | (setup.wValueH << 8);
[...]
		else if (GET_DESCRIPTOR == r)
		{
			ok = SendDescriptor(setup);    <---- enters here
		}
[...]
}
```

```C++
static bool SendDescriptor(USBSetup& setup)
{
	int ret;
	u8 t = setup.wValueH;
	if (USB_CONFIGURATION_DESCRIPTOR_TYPE == t)
		return SendConfiguration(setup.wLength);

	InitControl(setup.wLength);
#ifdef PLUGGABLE_USB_ENABLED
	ret = PluggableUSB().getDescriptor(t);     <--- enters here
	if (ret != 0) {
		return (ret > 0 ? true : false);
	}
#endif
```

```C++
int PluggableUSB_::getDescriptor(int8_t type)
{
	PUSBListNode* node;
	for (node = rootNode; node; node = node->next) {
		int ret = node->getDescriptor(type);
		// ret!=0 -> request has been processed
		if (ret)
			return ret;
	}
	return 0;
}
```

so we don't have the interface number information anywhere in the chain.
Maybe the setup.wIndex contains something that could be of interest here?

> It seems that the CDC Serial uses 2 interfaces. Not sure if this can be realized with 2 plugs or better with a real multiinterface API. We should defenitely consider to make the CDC class pluggable as well. This would save us some bytes. Not sure if we still have the static problem here as well.

The pluggable should already handle multiple interfaces, you just need to instantiate the PlugNode with the value of 2 on `numIfs`.

NicoHood:
Yes you can set numIFs o 2. But how can we actually call a methode of interface 2? This was too long ago since I looked at the code, but I think there were some parts where not all interfaces can be used. The implementation seemed not 100% completed/compatible to me. Well I have to look again later.

cmaglie:
> But how can we actually call a methode of interface 2? This was too long ago since I looked at the code, but I think there were some parts where not all interfaces can be used. The implementation seemed not 100% completed/compatible to me. Well I have to look again later.

Thanks for looking into this. In the meantime I'm merging this one so we can align the other implementations (for SAM and SAMD) as well, and eventually add the latest bit coming from your analysis.

Looking at the CDC implementation:
  * `CDC_GetInterface`  sends the whole USB descriptor altogether (for all the interfaces), without worring too much about interface number.
  * `CDC_Setup` didn't check for iface number but this is an implementation detail, the information is available and we can add it when we'll made CDC itself pluggable.
  * all the communication happens inside specific endpoints.

So, in the end, the USB-CDC seems to not suffer the problem above.

NicoHood:
Meh you deleted the branch. Now I cannot compare my changes real quick...

The multiinterface would be "disabled" with this commit I made:
https://github.com/NicoHood/Arduino/commit/98f331b6dc7817830d2ffd834b670455f76645d4

So its better to check the value in the HID class itself.

And yes we should rather use the wIndex value to check the Interface number. So we a) dont need to pass a 2nd value and b) we may remove the interfacetype and use this instead.
https://github.com/abcminiuser/lufa/blob/master/Demos/Device/LowLevel/KeyboardMouse/KeyboardMouse.c#L146

I also got a new idea (pretending the multi interface thing works) is to make the pluggable HID to also plug different Interface or endpoints. Depending if you want to use a multireport device or now. I will try this out myself.

TODO:
* fix the get and send names. Over the whole USBCore and Pluggable, but dont mix
* Remove the 2nd param in `getInterface` and `setup` and use `wIndex` instead
* make the endpoint number `uint8_t`. I saw some mixing there and unsigned was the better choice.
* I am not sure if the static initialization problem is fixed like that. I am not sure if its still flexible enough with this different syntax. (e.g. USB MIDI). One should try this out.
* Did I miss anything?

I guess you go ahead and fix this yourself since I currently have less time and want to avoid doing the work twice. Your commits look really good and I am really happy to see the change :)

NicoHood:
Meow.
https://github.com/arduino/Arduino/compare/master...NicoHood:plugfix3?expand=1

To rename the functions is up to you.

NicoHood:
May somebody explain me the static fix once again?

I think you currently cannot add multiple instances of HID like that. Originally I wanted to create a main HID class and inherit from this multiple interfaces. I think this is impossible now.

However I think I can solve this with a multiinterface HID class. You then can plug the HIDDevice to any interface number and give them unique interfaces if desired.

But then it is required to create a flexible endpoint type array and interface number etc which is generated at runtime. The Keyboard/Mouse constructor needs to be called first and then it adds them to the HID class. However this requires those variables to be non const:
```
  const uint8_t numEndpoints;
  const uint8_t numInterfaces;
  const uint8_t *endpointType;
```

Is this correct? Can anyone follow?

Then we can have (all pluggable):
* Interface 1 RawHID
* Interface 2 BootKeyboard
* Interface 3 Multireport with:
  * Mouse
 * Consumer
  * etc
* Interface 4 Gamepad (works better as separate interface)
* (not that CDC Serial limits this. If this would be pluggable too we could remove this and create awesome 4 player gamepads etc)

PR #3948, opened 2015-10-09
[PUSB] Minor Style change

NicoHood:


hardware/arduino/avr/libraries/HID/HID.cpp
```diff
@@ -96,11 +96,16 @@ bool HID_::setup(USBSetup& setup)
 			// TODO: Send8(protocol);
 			return true;
 		}
+		if (request == HID_GET_IDLE) {
+			// TODO: Send8(idle);
```
NicoHood: Send8() cannot be accessed from here anyways. But its just a comment and the other requests above have the same syntax. I already mentioned that in the other PR, that should be fixed as well.

PR #3958, opened 2015-10-11
Port Scheduler library to Arduino Zero.

PeterVH:
The library from the Due does not work as is on the Zero because it loads and stores the processor state using ldmia/stmia. On cortex m0, these instructions only work with the 'low' registers (r0-r7).

I made an implementation that loads the registers in the low registers first, and then moves them into the high regs.
Tested with MultipleBlink sketch.
Some feedback review on the code is desirable.
See [forum](http://forum.arduino.cc/index.php?topic=352408.0)

### DIFF
--- a/libraries/Scheduler/library.properties
+++ b/libraries/Scheduler/library.properties
@@ -1,9 +1,9 @@
 name=Scheduler
-version=0.4.3
+version=0.4.4
 author=Arduino
 maintainer=Arduino <info@arduino.cc>
-sentence=Allows multiple tasks to run at the same time, without interrupting each other. For Arduino DUE only.
-paragraph=The Scheduler library enables the Arduino Due to run multiple functions at the same time. This allows tasks to happen without interrupting each other.</br>This is a cooperative scheduler in that the CPU switches from one task to another. The library includes methods for passing control between tasks.
+sentence=Allows multiple tasks to run at the same time, without interrupting each other. For Arduino sam and samd architectures only (Due, Zero...).
+paragraph=The Scheduler library enables the Arduino to run multiple functions at the same time. This allows tasks to happen without interrupting each other.</br>This is a cooperative scheduler in that the CPU switches from one task to another. The library includes methods for passing control between tasks.
 category=Other
 url=http://www.arduino.cc/en/Reference/Scheduler
-architectures=sam
+architectures=sam,samd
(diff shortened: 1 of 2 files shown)


libraries/Scheduler/src/Scheduler.cpp
```diff
@@ -64,11 +83,39 @@ static void __attribute__((naked)) __attribute__((noinline)) coopTaskStart(void)
 static void __attribute__((naked)) __attribute__((noinline)) coopDoYield(CoopTask* curTask) {
 	asm (
 		"mov   r12, sp;"
-		"stmia r0, {r4-r12, lr};"
+		/* store low registers */
```
cmaglie: Shall this be put inside ifdef too, so we can use the more efficient store instruction when available (without copying the high registers into the low registers)?

libraries/Scheduler/library.properties
```diff
@@ -1,9 +1,9 @@
...
+version=0.4.4
 author=Arduino
 maintainer=Arduino <info@arduino.cc>
-sentence=Allows multiple tasks to run at the same time, without interrupting each other. For Arduino DUE only.
-paragraph=The Scheduler library enables the Arduino Due to run multiple functions at the same time. This allows tasks to happen without interrupting each other.</br>This is a cooperative scheduler in that the CPU switches from one task to another. The library includes methods for passing control between tasks.
+sentence=Allows multiple tasks to run at the same time, without interrupting each other. For Arduino sam and samd cores only (Due, Zero...).
+paragraph=The Scheduler library enables the Arduino to run multiple functions at the same time. This allows tasks to happen without interrupting each other.</br>This is a cooperative scheduler in that the CPU switches from one task to another. The library includes methods for passing control between tasks.
 category=Other
 url=http://www.arduino.cc/en/Reference/Scheduler
 architectures=sam
```
cmaglie: You should append `,samd` to the `architectures` field.

PeterVH:
Yes indeed, that chunk of m0 specific code escaped my attention.
I fixed both issues you brought up and squashed everything in one commit again.

B.t.w. I had big trouble testing this on my Due. Is there something seriously wrong with sam package 1.6.4? I could upload sketches to the Due (prog port) but could not run any simple led blinking program. Thought for a moment my Due was toasted. I reverted to 1.6.3 and everything was back ok.

Anyway test passed on Due too.

q2dg:
Should https://www.arduino.cc/en/Reference/Libraries list be updated to reflect Scheduler is "Due and Zero only", then?

If it is, Zero board should be referenced too in first sentence of https://www.arduino.cc/en/Reference/Scheduler ...

PR #3962, opened 2015-10-12
Improve Korean translations

jasonnam:
Try to focus on typos and obviously incorrect translations. Avoid any
debatable because of unnecessary bad impacts.


facchinm:
Hi @jasonnam , thanks for the submission!
Unfortunately we cannot directly merge this on Github because we use Transiflex to generate the translation files.
So, if you wish to contribute here (https://www.transifex.com/mbanzi/arduino-ide-15/) the strings will be automatically included in the next IDE build!

PR #3973, opened 2015-10-13
Allow spaces in IDE install path on Linux

### LINKED ISSUE #3950: IDE does not launch
IDE Version:   9 October 2015 13:15:11 GMT Nightly
OS: Ubuntu 15.04
Issue: IDE does not launch, throws an error (but latest stable build runs fine).

    :~/Arduino IDE$ ./arduino
    Picked up JAVA_TOOL_OPTIONS:
    Error: Could not find or load main class IDE.lib.splash.png

matthijskooijman:
The startup bash script lacked quotes in some places, causing it to
interpret part of the path to the splash image as the main class and
fail to start:

arduino: line 29: [: /home/a/Arduino: binary operator expected
Error: Could not find or load main class IDE.lib.splash.png

For the -splash option, simply adding quotes was not sufficient. If no
splash screen was to be used (so when $SPLASH was empty), using
"$SPLASH" would result in an empty argument being passed to java, which
was then interpreted by java as the name of the base class.

To allow spaces to occur in the -splash option, but also allow it to be
omitted entirely, options to java are now passed through the
$JAVA_OPTIONS array. By using the special "${JAVA_OPTIONS[@]}" syntax,
each element in the array is expanded into a single argument, even when
spaces are present inside (this is identical to what happens with "$@").

This fixes #3950

### DIFF
--- a/build/linux/dist/arduino
+++ b/build/linux/dist/arduino
@@ -3,9 +3,9 @@
 APPDIR="$(dirname -- "$(readlink -f -- "${0}")" )"
 
 for LIB in \
-    $APPDIR/java/lib/rt.jar \
-    $APPDIR/java/lib/tools.jar \
-    $APPDIR/lib/*.jar \
+    "$APPDIR"/java/lib/rt.jar \
+    "$APPDIR"/java/lib/tools.jar \
+    "$APPDIR"/lib/*.jar \
     ;
 do
     CLASSPATH="${CLASSPATH}:${LIB}"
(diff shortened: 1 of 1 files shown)


matthijskooijman:
@ArduinoBot build this please

build/linux/dist/arduino
```diff
@@ -17,18 +17,19 @@ export LD_LIBRARY_PATH
...
 
-$JAVA -DAPP_DIR="$APPDIR" -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel $SPLASH processing.app.Base "$@"
+# Collect options to java in an array, to properly handle whitespace in options
+JAVA_OPTIONS=("-DAPP_DIR=$APPDIR" "-Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel")
+
+if [[ "$@" != *"--upload"* && "$@" != *"--verify"* && "$@" != *"--get-pref"* && "$@" != *"--install-board"* && "$@" != *"--install-library"* ]] ; then
+	JAVA_OPTIONS+=("-splash:$APPDIR/lib/splash.png")
+fi
+
+$JAVA "${JAVA_OPTIONS[@]}" processing.app.Base "$@"
```
cmaglie: The variable `JAVA` can be obtained from:

`JAVA=$APPDIR/java/bin/java`

when the if on line 24 is followed and, since `APPDIR` may contain spaces, the last line should be changed to:

`"$JAVA" "${JAVA_OPTIONS[@]}" processing.app.Base "$@"`

(adding quotes to the $JAVA call), do you agree?

matthijskooijman: Good catch, agreed! Will you change it?

cmaglie: here we go https://github.com/arduino/Arduino/commit/a6602cac096c9d589f3d9b8ab917d9b38fcb8332

PR #3982, opened 2015-10-15
Mouse and Keyboard are supported on the Micro

obra:
The new Mouse and Keyboard error messages were inaccurate and misleading.

I'm thrilled that we're now warning users when they try to make a non-32u4 into a HID device. But we shouldn't steer them only toward one specific board when there are multiple totally legitimate official Arduino/Genuino boards that will work.

### DIFF
--- a/arduino-core/src/cc/arduino/Compiler.java
+++ b/arduino-core/src/cc/arduino/Compiler.java
@@ -549,12 +549,12 @@ public class Compiler implements MessageConsumer {
       }
 
       if (error.trim().equals("'Mouse' was not declared in this scope")) {
-        error = tr("'Mouse' only supported on the Arduino Leonardo");
+        error = tr("'Mouse' only supported on the Arduino Leonardo, Micro and other ATmega32U4 based devices");
         //msg = _("\nThe 'Mouse' class is only supported on the Arduino Leonardo.\n\n");
       }
 
       if (error.trim().equals("'Keyboard' was not declared in this scope")) {
-        error = tr("'Keyboard' only supported on the Arduino Leonardo");
+        error = tr("'Keyboard' only supported on the Arduino Leonardo, Micro and other ATmega3U4 based devices");
         //msg = _("\nThe 'Keyboard' class is only supported on the Arduino Leonardo.\n\n");
       }
 


matthijskooijman:
Looks good to me.

facchinm:
@orba, `Mouse` and `Keyboard` are also supported on Due and Zero so, in my opinion, in the long run we should add a field to boards.txt to mark the board as usbdevice-ready.
Furthermore, with PluggableUSB inclusion, the error can also be displayed if you forgot to include "Mouse.h" which is yet another case

cmaglie:
A better message could be:

`'Mouse' not found. Have you included 'Mouse.h'? Does your board support USB Mouse?`

But my preference is to remove completely the reference to the boards and just leave:

`'Mouse' not found. Have you included 'Mouse.h'?`

NicoHood:
u2 board also support USB :P 
I found this error confusing in general. Instead of a normal error you get an IDE special branded error.

Since Mouse and Keyboard are just APIs to generate USB report I'd leave out the USB stuff at all. If you look at my current HID Project dev version I created those APIs modular, so one can use the API so send the data via a Serial pipe to an USB AVR for example. (The IDE hasnt got this, just want to mention it).

One of @cmaglie s suggestions should be okay.

NicoHood:
Just found out how clever I am. Compiled my code accidentally for the Uno.

```cpp
#if !defined(USBCON)
#error HID Project can only be used with an USB MCU.
#endif
```
Got: `error: #error HID Project can only be used with an USB MCU.`

So why do we need this check at all? We can fix this inside the code. (might differ for SAM)

obra:
On Thu, Oct 15, 2015 at 05:18:33AM -0700, Nico wrote:
>    u2 board also support USB :P
>    I found this error confusing in general. Instead of a normal error you get
>    an IDE special branded error.

I'd be perfectly happy to see it removed entirely.

I'm a little sad that the requirement placed on my initial dev branch
(that Mouse.begin() or Keyboard.begin() implicitly do these #requires,
so old code keeps working) got dropped, but I'm incredibly happy to see
the USB HID stack getting pluggable. I'm excited to try to port the
Keyboardio extended Keyboard and Mouse support to it.

obra:
> 
>    'Mouse' not found. Have you included 'Mouse.h'?

Perhaps:

'Mouse' not found. Does your sketch include the line '#include <Mouse.h>'?

NicoHood:
What Keyboardio extended support? The internal Keyboard Mouse APIs are limited and thats intended. Dont know if you followed the other PRs but we decided to keep it simple and functional.

I've been working quite a lot now in the HID Project and just dropped an improved Keyboard API in the last commit. Feel free to contribute there or suggest new stuff.
https://github.com/NicoHood/HID/commit/1a676a2d203bfa158f550fc9d6463aa533b1c3ec

obra:
On Thu, Oct 15, 2015 at 11:25:47AM -0700, Nico wrote:
>    What Keyboardio extended support? The internal Keyboard Mouse APIs are
>    limited and thats intended. Dont know if you followed the other PRs but we
>    decided to keep it simple and functional.

I'm absolutely thrilled with the core APIs being simple, functional and
pluggable. What I was referring to was the other features in the now
year+ old pull request where I aggregated and cleaned up all the other
old HID pull requests.

If I recall correctly, the big things we had that the core didn't yet
have:

* Consumer Control 
* System Control
* Ability to wake the host
* Absolute mouse positioning
* A full USB HID keyboard mapping, rather than the old ~PS2 emulation.
* Slightly cleaned up code to support >6KRO in a non-boot-protocol
  compatible way. (That was just changing the key report size to a
  constant)
* Ability to vary the amount of power requested from the host 

Within the next few months, I need to actually get to a proper NKRO
implementation. All of this stuff, we'd be 100% happy to see upstream in
Arduino IF it makes sense for the core. 

Of course, it sounds like you've made tremendous progress while I've
been away building hardware.

> 
>    I've been working quite a lot now in the HID Project and just dropped an
>    improved Keyboard API in the last commit. Feel free to contribute there or
>    suggest new stuff

Awesome. I'll check it out. Thank you so much for pushing on all this
stuff!
>    [1]NicoHood/HID@1a676a2
> 
>    —
>    Reply to this email directly or [2]view it on GitHub.
> 
>    Reverse link: [3]unknown
> 
> References
> 
>    Visible links
>    1. https://github.com/NicoHood/HID/commit/1a676a2d203bfa158f550fc9d6463aa533b1c3ec
>    2. https://github.com/arduino/Arduino/pull/3982#issuecomment-148480360
>    3. https://github.com/arduino/Arduino/pull/3982#issuecomment-148480360

--

NicoHood:
Everything you mentioned is already working. USB wakeup is inside the core now.
System, Consumer etc is on the HID Project. NKRO as well. Different keyboard was (is currently broken) but can be patched within minutes (if i really want to) with the teensy API. RAWHID is also working (currently testing). and possibly many more i forgot about that just works now. like single reports instead of multiple. keyboard leds. bios support. multiple gamepads. you will see.

The wiki just lacks most information yet. If you think any feature is missing, open an issue at the hid project.

obra:
On Thu, Oct 15, 2015 at 11:54:09AM -0700, Nico wrote:
>    Everything you mentioned is already working. USB wakeup is inside the core
>    now.

That's amazingly wonderful news. Thank you!

cmaglie:
@obra your last error sentence looks better, may you update your patch with that one?

obra:
Obsoleted by PR #4018

NicoHood:
And inside the Mouse.h file I suggest to check if the device is USB capable.

Also I am not sure if this check (in the PR) is still required then. I think this check is also a bit "unfair" to 3rd party libraries, because I cannot add such a check.

PR #4009, opened 2015-10-21
SAM: add iSerial USB field

facchinm:
Porting of #3811 for SAM architecture
/cc @cmaglie


NicoHood:
Just a small note here:
If the plugged interface adds more than the 20 available chars this will write to undefined memory. I think its not needed to fix this, since this should not happen on a limited number on endpoints. But its sure a risk if people feel like writing stories to the field.

It should be noted in the docs though:
https://github.com/arduino/Arduino/wiki/PluggableUSB-and-PluggableHID-howto

PR #4027, opened 2015-10-25
Added return value to HID SendReport

NicoHood:
This is required for a good API to check if the send failed or not (e.g. device disconnected from USB to not block with waiting for timeouts).

### DIFF
--- a/hardware/arduino/avr/libraries/HID/HID.cpp
+++ b/hardware/arduino/avr/libraries/HID/HID.cpp
@@ -88,10 +88,11 @@ void HID_::AppendDescriptor(HIDSubDescriptor *node)
 
 int HID_::SendReport(uint8_t id, const void* data, int len)
 {
-    int ret = 0;
-	ret += USB_Send(pluggedEndpoint, &id, 1);
-	ret += USB_Send(pluggedEndpoint | TRANSFER_RELEASE, data, len);
-	return ret;
+    auto ret = USB_Send(pluggedEndpoint, &id, 1);
+    if(ret >= 0){
+        ret += USB_Send(pluggedEndpoint | TRANSFER_RELEASE, data, len);
+    }
+    return ret;
 }
 
 bool HID_::setup(USBSetup& setup)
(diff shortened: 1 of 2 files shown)


NicoHood:
I want to note that the underlying Keyboard and mouse API is not affected, since they just dont use the return value.

This should be also ported to the DUE.

HID-Project currently relies on this PR, so people cannot use it at the moment if they dont recompile the IDE manually.

Added another commit to avoid returning -2.

It would make sense to refactor the sending function to return a size_t instead of int. This way we dont need to use a signed variable and 0 indicates a failure.

cmaglie:
LGTM

@facchinm ?

NicoHood:
Any reason why USB_Send() is int instead of size_t?

hardware/arduino/avr/libraries/HID/HID.cpp
```diff
@@ -86,10 +86,13 @@ void HID_::AppendDescriptor(HIDSubDescriptor *node)
...
 }
 
-void HID_::SendReport(uint8_t id, const void* data, int len)
+int HID_::SendReport(uint8_t id, const void* data, int len)
 {
-	USB_Send(pluggedEndpoint, &id, 1);
-	USB_Send(pluggedEndpoint | TRANSFER_RELEASE, data, len);
+    auto ret = USB_Send(pluggedEndpoint, &id, 1);
+    if(ret >= 0){
+        ret += USB_Send(pluggedEndpoint | TRANSFER_RELEASE, data, len);
```
cmaglie: the second call to `USB_Send` may fail too, we must check for `USB_Send` return value before accumulating to `ret`. I'm doing it...

NicoHood: Damn how clever. Will you just add a commit and merge it with this one?

The question is still not answered: Why dont we use size_t?

cmaglie: `size_t` is unsigned, if we use it we can't return `-1`.

cmaglie:
Rebased and merged, implemented also on samd

hardware/arduino/avr/libraries/HID/HID.cpp
NicoHood: Yes but 0 would also indicate an error? Well I dont really know whats best here since you always have this dilemma whit different return types across differen functions. But I think size_t would fit better.

PR #4047, opened 2015-10-29
Update README.md

devoRPa:


cmaglie:
http://jakubmarian.com/information-is-vs-informations-are-is-information-plural-or-singular-in-english/

Issue #4029, opened 2015-10-25
Flush is broken for SoftwareSerial, WiFiClient TCP and UDP and other classes


drmpf:
Flush is broken for SoftwareSerial, WiFiClient TCP and UDP and other classes
They does not honor the update description for Flush in the stream class

the current docs for Flush() https://www.arduino.cc/en/Serial/Flush say
"Waits for the transmission of outgoing serial data to complete. (Prior to Arduino 1.0, this instead removed any buffered incoming serial data.)
flush() inherits from the Stream utility class."

However for WiFiClient::flush(), WiFiUDP::flush(), SoftwareSerial::flush(), EthernetUDP::flush()
  the current code, as of V1.6.4, clears the RX buffer, which is unexpected and leads to loss of incoming data.

I found this while tracking down why my WiFi connection was losing incoming messages sometimes.

Some classes implementations of flush() do nothing so a simple fix would be to replace the non-complient methods with empty methods.

matthijskooijman:
Confirmed for SoftwareSerial. Fix is simply removing the flush() implementation, since SoftwareSerial does not do any TX buffering, so flush() is a no-op.

cmaglie:
Thank you!

PR #4133, opened 2015-11-11
Fix flush() behaviour on a bunch of classes/libraries

cmaglie:
As described in #4029 

/cc @drmpf @matthijskooijman


matthijskooijman:
Looks good to me.

NicoHood:
I think so too.

PR #4138, opened 2015-11-12
Track if a USB ZLP is needed on flush

### LINKED ISSUE #3946: Leonardo fails to send exactly 64 bytes.

    void setup() {

    }

    void loop() {
      if (Serial) {
        if (Serial.available() > 0) {
          Serial.read();
          Serial.write(buff,64);
          Serial.flush();
        }
      }
    }

sandeepmistry:
Resolves #3946. This patch tracks if a USB ZLP (zero length packet) is needed to flush data.

A ZLP is needed to indicate end of transfer, if the size of data sent is a multiple of ```USB_EP_SIZE``` (64 bytes on AVR).

If a ZLP is needed, ```USB_Flush``` waits for the current endpoint's FIFO bank to be available before doing a release.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -302,9 +302,15 @@ int USB_Send(u8 ep, const void* d, int len)
 				while (n--)
 					Send8(*data++);
 			}
-			_sendZlp[ep & USB_ENDPOINTS_MASK] = !ReadWriteAllowed() && (len == 0);
-			if (!ReadWriteAllowed() || ((len == 0) && (ep & TRANSFER_RELEASE)))	// Release full buffer
+			_sendZlp[ep & USB_ENDPOINTS_MASK] = !ReadWriteAllowed() && (len == 0) && !(ep & TRANSFER_RELEASE);
+
+			if (!ReadWriteAllowed())	// Release full buffer
 				ReleaseTX();
+
+			if ((len == 0) && (ep & TRANSFER_RELEASE)) {
+				while(!ReadWriteAllowed());
+				ReleaseTX();
+			}
 		}
 	}
 	TXLED1;					// light the TX LED
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -191,6 +191,8 @@ static inline u8 FrameNumber()
 //==================================================================
 //==================================================================
 
+bool _sendZlp[USB_ENDPOINTS];
+
 u8 USBGetConfiguration(void)
 {
 	return _usbConfiguration;
@@ -300,6 +302,7 @@ int USB_Send(u8 ep, const void* d, int len)
 				while (n--)
 					Send8(*data++);
 			}
+			_sendZlp[ep & USB_ENDPOINTS_MASK] = !ReadWriteAllowed() && (len == 0);
 			if (!ReadWriteAllowed() || ((len == 0) && (ep & TRANSFER_RELEASE)))	// Release full buffer
 				ReleaseTX();
 		}
@@ -625,8 +628,15 @@ ISR(USB_COM_vect)
 void USB_Flush(u8 ep)
 {
 	SetEP(ep);
-	if (FifoByteCount())
+
+	// wait for write access if a ZLP is needed
+	if (_sendZlp[ep])
+		while(!ReadWriteAllowed());
+
+	if (FifoByteCount() || _sendZlp[ep])
 		ReleaseTX();
+
+	_sendZlp[ep] = false;
 }
 
 static inline void USB_ClockDisable()
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -204,7 +204,7 @@ public:
 	LockEP(u8 ep) : _sreg(SREG)
 	{
 		cli();
-		SetEP(ep & 7);
+		SetEP(ep & USB_ENDPOINTS_MASK);
 	}
 	~LockEP()
 	{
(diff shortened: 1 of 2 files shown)


embmicro:
This looks good but I propose changing the line where you assign _sendZlp to the following so that if the TRANSFER_RELEASE flag is set you don't set _sendZlp.

			_sendZlp[ep & USB_ENDPOINTS_MASK] = !ReadWriteAllowed() && (len == 0) && !(ep & TRANSFER_RELEASE);

			if (!ReadWriteAllowed())	// Release full buffer
				ReleaseTX();

			if ((len == 0) && (ep & TRANSFER_RELEASE))
				ReleaseTX();

That is unless I have a misunderstanding on what TRANSFER_RELEASE is supposed to do.

sandeepmistry:
@embmicro thanks for reviewing!

As for your suggestion, if ```ReleaseTX``` is called twice in row, the second call must wait for a ```ReadWriteAllowed()```to be true again, otherwise it won't send a ZLP. Something like this:

```arduino
        bool sendZlp = !ReadWriteAllowed() && (len == 0);

        if (!ReadWriteAllowed())    // Release full buffer
            ReleaseTX();
       
        if (sendZlp && (ep & TRANSFER_RELEASE)) {
            sendZlp = false;
           while(!ReadWriteAllowed());
        }

        if ((len == 0) && (ep & TRANSFER_RELEASE))
            ReleaseTX();

        _sendZlp[ep & USB_ENDPOINTS_MASK] = _sendZlp;
```

NicoHood:
Can you build this? I want to try.

sandeepmistry:
@ArduinoBot build this please

embmicro:
Can't you just add the while then like the following?
```java
			_sendZlp[ep & USB_ENDPOINTS_MASK] = !ReadWriteAllowed() && (len == 0) && !(ep & TRANSFER_RELEASE);

			if (!ReadWriteAllowed())	// Release full buffer
				ReleaseTX();

			if ((len == 0) && (ep & TRANSFER_RELEASE)) {
				while(!ReadWriteAllowed());
				ReleaseTX();
			}
```

sandeepmistry:
@embmicro agreed, your suggestion is much cleaner. Initially I thought the r/w check was only needed if a ZLP needs to be sent when the ```TRANSFER_RELEASE``` flag is set, but it doesn't hurt to check it regardless.

I've pushed your suggestion in https://github.com/sandeepmistry/Arduino/commit/bc808248d75e4abd218df6a428ba046d7aa80cb4

sandeepmistry:
@ArduinoBot build this please

embmicro:
I've run into a problem with this patch. Take the following example.

```cpp
void setup() {

}

void loop() {
  if (Serial) {
    if (Serial.read() == 'h'){
      Serial.write("Hello ");
      Serial.flush();
      Serial.write("World!\r\n");
      Serial.flush();
    }
  }
}
```

With this sketch I expect "Hello World!" to be printed every time I send an 'h' to the board. However, on my Windows 10 machine, I typically only get "Hello ". Sending another 'h' will then show "Hello World!\r\nHello "

Removing the calls to flush makes it work as expected, but that isn't much of a solution.

On my Linux machine it works perfectly fine with and without flush.

NicoHood:
Sounds like an useful option if it works on linux. +1 from me ;)

For the windows users a better fix might be to not send USB_EP_SIZE bytes. Also as described above, I dont think every EP should use the ZLP mechanism, since other devices work different.
I suggest to count the bytes in buffer and if there are any, flush them. For the write() function we should send those bytes if they exceed USB_EP_SIZE - 1, so we dont have to deal with a ZLP at all. That is also what LUFA does in its usb-serial.

embmicro:
I don't know if this is ideal, but I modified USB_Send to never send a full packet and it seems to be working with all my tests.

```c++
int USB_Send(u8 ep, const void* d, int len)
{
	if (!_usbConfiguration)
		return -1;

	int r = len;
	const u8* data = (const u8*)d;
	u8 timeout = 250;		// 250ms timeout on send? TODO
	while (len)
	{
		u8 n = USB_SendSpace(ep);
		if (n == 0)
		{
			if (!(--timeout))
				return -1;
			delay(1);
			continue;
		}

		n--; // leave one byte free

		if (n > len)
			n = len;
		{
			LockEP lock(ep);
			// Frame may have been released by the SOF interrupt handler
			if (!ReadWriteAllowed())
				continue;
			len -= n;
			if (ep & TRANSFER_ZERO)
			{
				while (n--)
					Send8(0);
			}
			else if (ep & TRANSFER_PGM)
			{
				while (n--)
					Send8(pgm_read_byte(data++));
			}
			else
			{
				while (n--)
					Send8(*data++);
			}
			
			// Release full buffer
			if (!ReadWriteAllowed() || USB_SendSpace(ep) == 1 || ((len == 0) && (ep & TRANSFER_RELEASE)) )
				ReleaseTX();
		}
	}
```

I also reverted flush back to it's original form. Still calling flush a lot like in my previous example causes trouble. However, flush doesn't seem to be needed at all so maybe it should be an empty function.

BlackBrix:
does someone knows how this is going on ?
will this last suggestion of @embmicro be implemented somehow ?

I think I maybe have a related problem with 32u4 (?) 
--> [http://forum.arduino.cc/index.php?topic=360286.msg2686772#msg2686772](http://forum.arduino.cc/index.php?topic=360286.msg2686772#msg2686772)
--> [http://forum.arduino.cc/index.php?topic=360286.msg2686772#msg2686772](http://forum.arduino.cc/index.php?topic=360286.msg2689095#msg2689095)

can I just try to exchange the original USB_Send() with this last one from @embmicro without changing any other code segments ?

sandeepmistry:
I'm closing this in favour of #4864. Which is a similar approach to @embmicro's suggestion in https://github.com/arduino/Arduino/pull/4138#issuecomment-163766826.

However, it will spin with a delay and timeout if buffer space is EP size - 1.

PR #4153, opened 2015-11-15
Stepper library: Fixed counter clockwise turn for four wire

yoones:
Bug noticed when using a STEP MOTOR 28BYJ-48 5VDC.

When calling step(), the step motor always turned in the same direction regardless of the sign of the parameter given (a negative value wouldn't make it turn counter clockwise). This patch fixes the issue.

### DIFF
--- a/libraries/Stepper/src/Stepper.cpp
+++ b/libraries/Stepper/src/Stepper.cpp
@@ -1,13 +1,15 @@
 /*
  * Stepper.cpp - Stepper library for Wiring/Arduino - Version 1.1.0
  *
- * Original library        (0.1)   by Tom Igoe.
- * Two-wire modifications  (0.2)   by Sebastian Gassner
- * Combination version     (0.3)   by Tom Igoe and David Mellis
- * Bug fix for four-wire   (0.4)   by Tom Igoe, bug fix from Noah Shibley
- * High-speed stepping mod         by Eugene Kozlenko
- * Timer rollover fix              by Eugene Kozlenko
- * Five phase five wire    (1.1.0) by Ryan Orendorff
+ * Original library                    (0.1)   by Tom Igoe.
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Your pullrequest seems to only add 2 files, not modify anything. Was that the intention? I suspect you added the updated files outside of the src directory. This makes it impossible to see what changed and review your PR. If you modify this (git commit --amend should help), you can force push to your github branch to automatically update this PR.

Regarding the bug you describe, I've seen the same thing and fixed it by swapping wire 2 and 3, assuming that my stepper had a unintuitive wiring. Your PR suggests that either you have the same, or the problem wasn't in the pinout of my stepper after all (though IIRC the AccelStepper required the same swapping of wires).

yoones:
PR fixed.

This fix comes from my understanding of https://en.wikipedia.org/wiki/Stepper_motor. If you look at the gif animation (right side), you see that each electromagnet is turned on alone, one after the other. Applying this let me turn the stepper motor counter clockwise as well.

matthijskooijman:
@yoones, the PR still contains a lot of incorrect and certainly unrelated whitespace changes, making review this diff harder than needed. Please be sure to look at your own diff before submitting a PR, and remove any changes that you did not intend.

As for the actual fix, it seems that you are switching from full step drive to wave drive, which only has a single coil active, but delivers less torque and thus isn't commonly used and, I think, shouldn't be the default here. Both are explained further down in the article you linked.

I'm out of time for now to figure out why the original code doesn't work. Perhaps the number of pins is inconsistent between the stepper motors and the library? (e.g. one starts with both pins of 1 coil, while the other starts with the first pins of both coils?)

skniazev:
Take a look at #4250. The new version of the library and the example stepper_oneRevolution_ULN2003.ino must fix the issue.

PR #4171, opened 2015-11-17
Port Serial_::readBreak() API to SAM core

### LINKED ISSUE #4094: Due: SerialUSB port busy after disconnecting and trying to reconnect
This is a regression in 1.6.6. 1.5.5 works correctly. 

On Arduino Due, upload any sketch:

````
void setup() {
SerialUSB.begin(115200);
}

void loop() {
  SerialUSB.println("hello");
  delay(100);
}
````

1. Upload it to the board using Native USB port.
2. Run Serial monitor.
3. It should connect and display the output correctly.
4. Close Serial Monitor.
5. Open Serial Monitor again.

Expected: it connects correctly.

Actual: get a `Error opening serial port '/dev/cu.usbmodem1d111'. (Port busy)` error. This happens in other apps too, like "Serial Tools" for OS X. It seems that when you close the connection on the PC side, the Arduino is not properly cleaning up and freeing the connection.

Tested on Arduino Due, OS X 10.10.5, with Arduino 1.6.6 (Due core 1.6.5).

sandeepmistry:
Port of the AVR ```Serial_::readBreak()``` API (https://github.com/arduino/Arduino/commit/b58f2394005cd073d582b8f90f8ba222aa95b0ee) to the SAM core.

Resolves #4094, as SAM core was not correctly handling receiving a ```CDC_SEND_BREAK``` and sending a stall, causing OS X to report the serial port as "busy". (OS X sends the break when closing the serial port).

cc/ @matthijskooijman

### DIFF
--- a/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
+++ b/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
@@ -55,6 +55,8 @@ static volatile LineInfo _usbLineInfo = {
     0x00   // lineState
 };
 
+static volatile int32_t breakValue = -1;
+
 _Pragma("pack(1)")
 static const CDCDescriptor _cdcInterface =
 {
@@ -144,6 +146,7 @@ bool WEAK CDC_Setup(USBSetup& setup)
 
 		if (CDC_SEND_BREAK == r)
 		{
+			breakValue = ((uint16_t)setup.wValueH << 8) | setup.wValueL;
 			return true;
 		}
 	}
@@ -303,6 +306,28 @@ Serial_::operator bool()
 
 	delay(10);
 	return result;
+}
+
+int32_t Serial_::readBreak() {
+	uint8_t enableInterrupts = ((__get_PRIMASK() & 0x1) == 0 && 
+								(__get_FAULTMASK() & 0x1) == 0);
+
+	// disable interrupts,
+	// to avoid clearing a breakValue that might occur 
+	// while processing the current break value
+	__disable_irq();
+
+	int ret = breakValue;
+
+	breakValue = -1;
+
+	if (enableInterrupts)
+	{
+		// re-enable the interrupts
+		__enable_irq();
+	}
+
+	return ret;
 }
 
 unsigned long Serial_::baud() {
--- a/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
+++ b/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
@@ -141,6 +141,11 @@ bool WEAK CDC_Setup(USBSetup& setup)
 			}
 			return true;
 		}
+
+		if (CDC_SEND_BREAK == r)
+		{
+			return true;
+		}
 	}
 	return false;
 }
--- a/hardware/arduino/sam/cores/arduino/USB/USBAPI.h
+++ b/hardware/arduino/sam/cores/arduino/USB/USBAPI.h
@@ -63,6 +63,23 @@ public:
 	using Print::write; // pull in write(str) from Print
 	operator bool();
 
+	// This method allows processing "SEND_BREAK" requests sent by
+	// the USB host. Those requests indicate that the host wants to
+	// send a BREAK signal and are accompanied by a single uint16_t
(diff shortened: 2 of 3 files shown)


sandeepmistry:
@matthijskooijman I've pushed a change (https://github.com/sandeepmistry/Arduino/commit/2c1462d828ef739bf39cd27d6530c1b5b360f08a) to disable and re-enable USB OTG interrupts to preventing overwriting the ```breakValue``` with -1.

Any suggestions on how to test this condition, it's difficult to distinguish between missing a break from not reading frequently enough and overwriting break value.

matthijskooijman:
https://github.com/arduino/ArduinoCore-samd/pull/65#issuecomment-169943359 also applies here.

sandeepmistry:
@matthijskooijman I've addressed your feedback in https://github.com/sandeepmistry/Arduino/commit/6ffc96b8f60016a18ccb0898ceb0a478a792953b, and have taken a similar approach to ```Serial_::accept```, by using the ARM exclusive access instructions (as suggested by @cmaglie).

sandeepmistry:
@matthijskooijman this is inline with the SAMD changes now, would you mind reviewing again? Then I can squash the commits and rebase.

matthijskooijman:
Looks ok to me.

I had not seen your "exclusive access" approach here before, though (somehow missed that commit). Looking more closely now, that does also seem like it could work. I don't think your previous attempt is correct (since the exclusive store [could fail, requiring the entire load-store sequence to be retried](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/BABDEEJC.html)), though.

At first glance, this exclusive access stuff does not protect against modification by an interrupt handler, only other processors or cores in a multicore setup, but it [seems that an interrupt between exclusive load and store will also cause store to fail](http://electronics.stackexchange.com/a/25721), so that would also work. An advantage of using the exclusive access primitives would be that it also helps in multicore setups (disabling interrupts only helps with a single core), though that is not really interesting for the current hardware. Another advantage is that by retrying when an interrupt is received, you gain safety at the cost of (potential) slowdown in the main loop, without delaying any interrupt at all. (on a sidenote, I thought this would also be possible to implement in software, e.g. by manually setting an "an interrupt has occurred"-bit in interrupt handlers, but doing this requires a way to *atomically* store a value, but only if the "an interrupt has occured"-bit is not set (which on e.g. AVR requires disabling interrupts again, I believe).

I suspect that the M0 does not support these access primitives (and probably not multicore or multiprocessor setups either), so just disabling interrupts is the proper solution there? If so, it might be better to do the same here, in the interest of consistency.

sandeepmistry:
> I suspect that the M0 does not support these access primitives (and probably not multicore or multiprocessor setups either), so just disabling interrupts is the proper solution there? If so, it might be better to do the same here, in the interest of consistency.

Yes, this is correct the M0 doesn't support those instructions.

>  I don't think your previous attempt is correct (since the exclusive store could fail, requiring the entire load-store sequence to be retried), though

Yes, it was not working if any interrupt (not just a USB one) was triggered between the load and store. I was testing using ```delay``` as you had suggested earlier.

There is small loss when retrying the load-store sequence, you would miss the previous break value, if a break occurs while reading. I'm still open to making this change though, but you do bring up a good point of doing the same as the SAMD core here.

matthijskooijman:
> There is small loss when retrying the load-store sequence, you would miss the previous break value, if a break occurs while reading.
A, good point. On the other hand, that will probably (almost) never happen, while delaying an interrupt by a few clockcycles will happen more often (or less seldomly, perhaps).

I'm not sure what is the best route to take here, a point can be made for both. I'm inclined to keep the `__disable_irq()` approach, because it is simpler, and consistent (in code) with SAM and (in concept) with AVR as well.

sandeepmistry:
Ok, let's stick with the ```__disable_irq()``` approach then.

I've squashed my later commits and rebased.

matthijskooijman:
Looks good!

sandeepmistry:
@matthijskooijman indentation update is in https://github.com/arduino/Arduino/commit/dcfcbefc32f97dd508707751f99848e37f2b763c

Again, thanks for reviewing!

PR #4172, opened 2015-11-17
Send a ZLP if data size is multiple of the endpoint size for USB sends in SAM core

### LINKED ISSUE #3946: Leonardo fails to send exactly 64 bytes.

sandeepmistry:
Resolves #3946 for SAM/Due.

**Note**: The CDC endpoint size on SAM is 512 bytes (not 64 bytes).


facchinm:
Moved to  arduino/ArduinoCore-sam#2

PR #4186, opened 2015-11-19
Fix README.md and Resources_fr

armel:
Hope this help :)


ffissore:
Thank you @armel Translation files are generated out of Transifex, we don't edit them directly. Please join the translators community at https://www.transifex.com/mbanzi/arduino-ide-15/

PR #4192, opened 2015-11-20
Added SerialUSB alias for Serial on 32u4 boards

cmaglie:
Allows the use of `SerialUSB` on Micro/Leonardo and other 32u4 derivatives.

`SerialUSB` doesn't replace but is added as an alias of `Serial`.


cmaglie:
related to #3680

NicoHood:
Yeah why not. +1

matthijskooijman:
+1, improves compatibility with the samd core as well.

PR #4210, opened 2015-11-24
fixed "download" message value in Polish

MarcinGrabowiecki:


ffissore:
Thank you @MarcinGrabowiecki However translations files are not manually edited. We use transifex https://www.transifex.com/mbanzi/arduino-ide-15/ Please consider joining the Polish translations group

PR #4252, opened 2015-12-04
Ported Arduino-usbserial to version 151115 of LUFA.

AlexHolden:
Hi, over in the grbl project a couple of us were experiencing data corruption when communicating with the Arduino at 250,000 baud. I tracked the problem down to the old USB-Serial firmware on the 16u2 and was able to fix it by porting the Arduino-specific changes forward to the newer version of USB-Serial from LUFA  151115.


NicoHood:
I  suggest that IF you change the bootloader, then with full 2M full duplex.

Original implementation:
https://github.com/urjaman/fast-usbserial

Adapted with HoodLoader2 (16u2 itself is also reprogrammable)
https://github.com/NicoHood/HoodLoader2

So if the speed is the problem I suggest to use HoodLoader2, cause it is simple to flash and you can keep it. You are also able to upload the old usb-serial firmware if any problems occur with hoodloader2 (which wont happen, just showing the infinite possibilities)

AlexHolden:
Thanks Nico, I wish I had known fast-usbserial existed before, as it probably would have saved me a day or two of debugging. The problem with the official firmware is it appears to work at 250K, until you start pushing a lot of data through it, then it starts randomly corrupting the transfer (in the case of grbl this could easily lead to an expensive/dangerous machine crash).

After tracking the problem down to the USBtoSerial firmware I first tried to recompile the official Arduino fork under the recommended LUFA 100807. It wouldn't compile with a recent avr-gcc. I then located and tried USBtoSerial from the latest LUFA 15115 (I could see that it contained several bug fixes and optimisations). It compiled but of course it wouldn't work without changes on the Arduino because the clock speed is different and the LEDs are wired differently and it needs the reset out line connecting to the DTR event and a couple of other things.

Finally I looked at what had been changed in the Arduino fork and ported them forward to the latest LUFA USBtoSerial. I needed to change how the timer to turn the LEDs off works, but other than that it's a fairly simple forward-port, and in my testing it completely eliminated the data corruption I was getting at 250K baud.

NicoHood:
Thatswhy just use HoodLoader2. Maybe they will make it an official feature.

AlexHolden:
Sorry Nico, I don't see your point. Are you saying, "don't accept this patch, take HL2 instead"?

For my purposes, all I need is a simple direct replacement for the USBtoSerial firmware that works reliably at baud rates higher than 115200. My patch works for me so I'll stick with it on my own boards. It sounds like fast-usbserial would have done the job too, had I known about it beforehand. The advantage my patch has over fast-usbserial is that it is pretty much just a forward-port of the Arduino changes to the upstream source, whereas FU has a lot more deep optimisation including several blocks of uncommented inline assembly.

I really don't mind if my patch gets accepted or not, though I think it would be a good thing for everyone if future Arduinos shipped with a more reliable firmware on the 16u2.

NicoHood:
Yes I am saying this. If you patch it (a bootloader/firmware patch like this is something you dont do every day) then do it right. Not only 250K, why not full duplex 2M. The additional CDC bootloader would also be an enhancement. Fastusbserial itself would be enough, but at least better than 250M.

Those usb-serial patches have been around here before. I really agree with you to update it, but then with more features, so people dont complain next time about 1M not working, then 2M and then they want to use the CDC programming.

However the option of using HoodLoader2 or fast-usbserial is there. Everyone can install it, its just nothing official.

AlexHolden:
Well, I wrote 250K because that's what I've tested it at (full duplex obviously). I don't actually know the maximum it will support. Probably not as high as fast-usbserial, I would guess.

I don't have an opinion on your bootloader; all I needed for my application was a more reliable usb to serial firmware, and I didn't experience any problems using the standard dfu-programmer procedure to upload it.

PR #4279, opened 2015-12-09
(not so) trivial keyboard shortcut fixes and cleanup

matthijskooijman:
Here's a few cleanups and fixes to EditorListener, which handles some keyboard shortcuts.


matthijskooijman:
While testing, @cmaglie pointed out that the ctrl-tab/ctrl-shift-tab bindings would actually break in some circumstances. Further digging turned out that the accelerators in the tab popup menu would actually not work reliably (only until the menu was opened for the first time), so I dug in to find a way to make them more reliable.

After a few failed attempts, I ended up with a working solution. As part of this solution, I switched the menu items to use Action objects, and it made sense to move the ctrl-alt-left/right bindings into EditorHeader as well. This left EditorListener a nearly empty class, containing just one overly-broad, poorly documented hack, so I figured out a slightly better solution for that as well, so EditorListener could be entirely removed.

So, this PR ended up a fair bit more complex that the original, but at least things are even cleaner than before.

matthijskooijman:
I fixed one more commit to not introduce unused imports.

@ffissore, I see Arduinobot saying the build failed, also on another PR, but it builds completely locally. Is there a problem with the bot?

ffissore:
Yes, bandwidth problems :( It should be fixed now

ffissore:
@ArduinoBot build this please

matthijskooijman:
And one more push, I forgot to add license headers on two new java files, which I now fixed.

NicoHood:
I gave it a quick test. It works for short key presses. As long as I hold down the ctr alt right for a rew seconds it doesnt work anymore. However the editor doesnt crash, but the shortcut seems to break (for the whole opened instance).

Scrolling over the tabs is still not implemented, which would be the most intuitive switching. But I guess that wasnt aimed in this PR.

matthijskooijman:
> I gave it a quick test. It works for short key presses. As long as I hold down the ctr alt right for a rew seconds it doesnt work anymore. However the editor doesnt crash, but the shortcut seems to break (for the whole opened instance).

Weird, I'm not sure what's going on there, then.

> Scrolling over the tabs is still not implemented, which would be the most intuitive switching. But I guess that wasnt aimed in this PR.

Correct :-)

NicoHood:
The described bug from my last post still exists. However if you dont escalate on holding down those keys it works fine. It should be fixed though.

lmihalkovic:
IMHO merging this code was not a smart move, and even if you do not need the extra feature my version has, this one might still bite you in the future.

PR #4280, opened 2015-12-09
Rework CDC reset to discover newer bootloader at runtime

facchinm:
Tested with Hoodloader2, should work with every LUFA-derived bootloader released after 2014 (.apitable_signatures section must be placed at end of the flash)

```
BootloaderAPITable.S :

.global BootloaderAPI_Signatures
BootloaderAPI_Signatures:

	.long BOOT_START_ADDR ; Start address of the bootloader
	.word 0xDF00 ; Signature for the CDC class bootloader
	.word 0xDCFB ; Signature for a LUFA class bootloader

makefile:

BOOT_API_LD_FLAGS    += $(call BOOT_SECTION_LD_FLAG, .apitable_signatures,  BootloaderAPI_Signatures,  8)

```

### DIFF
--- a/hardware/arduino/avr/cores/arduino/CDC.cpp
+++ b/hardware/arduino/avr/cores/arduino/CDC.cpp
@@ -102,21 +102,43 @@ bool CDC_Setup(USBSetup& setup)
 #ifndef MAGIC_KEY
 #define MAGIC_KEY 0x7777
 #endif
+
+// Legacy magic key position. Looks for old and new boot key position.
+// Use RAMEND-1 as MAGIC_KEY_POS to save a few bytes of flash and force the new bootloader.
 #ifndef MAGIC_KEY_POS
 #define MAGIC_KEY_POS 0x0800
+#endif
+
+#ifndef NEW_LUFA_SIGNATURE
+#define NEW_LUFA_SIGNATURE 0xDCFB
+#endif
+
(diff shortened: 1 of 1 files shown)


NicoHood:
Looks like my patch 1:1? If so it should work. Have used this for quite some time now with HoodLoader2 32u4. Should not break anything.

The question is still not answered:
Who will patch the 32u4 bootloader and when? I just dont want to do work twice and as said I currently have not much time, so I try to spend it on other stuff. But I really want to support the new bootloader. 

With this patch basically all doors are open, so just merge it I'd say.

facchinm:
Indeed, this is exactly your patch (I checked through Hoodloader2 to understand why on some chips `MAGIC_KEY_POS != (RAMEND-1)` doesn't necessarily mean `LEGACY_BOOTLOADER`.
I'm currently testing it with a bunch of boards and I've seen no regression, so I'd merge it quickly.
/cc @cmaglie

NicoHood:
Okay cool!
But who will develop the new bootloader? Just to give me an idea of the priority here. Its not a pity if you say "probably in 3 month or later". Thats at least a more clear statement.

facchinm:
Merged "check once" variant: 4c901d3f9f347177b3f263c451885ddc3b1b6ab0

NicoHood:
Great to see this merged!
However this issue is still not fixed, as there is no fixed bootloader available yet.

PR #4288, opened 2015-12-10
arduino-builder uses log levels

ffissore:
Upgrading arduino-builder to 1.3.7, which changes the way messages are sent back to the IDE. Instead of having just stdour and stderr, stdout only is used, but each message has a log level: info, warn, debug, error.
Plain stdout/stderr are still used by child processes


ffissore:
@ArduinoBot build this please

ffissore:
@ArduinoBot build this please

PR #4312, opened 2015-12-16
Rework {runtime.tools.$tool.path} expansion if multiple versions found

facchinm:
Test case: Intel Edison packages sketchUploader version 1.6.2, while Arduino101 packages v.1.6.4 for the same tool.
Use semantic versioning to assign
{runtime.tools.sketchUploader.path} to 1.6.4, while keeping {runtime.tools.sketchUploader-1.6.4.path} and {runtime.tools.sketchUploader-1.6.2.path} untouched

The new logic is not affected by the order in which the board_package jsons are parsed

@cmaglie @sandeepmistry

### DIFF
--- a/arduino-core/src/processing/app/BaseNoGui.java
+++ b/arduino-core/src/processing/app/BaseNoGui.java
@@ -33,6 +33,8 @@ import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import  cc.arduino.contributions.VersionComparator;
+
 import static processing.app.I18n.tr;
 import static processing.app.helpers.filefilters.OnlyDirs.ONLY_DIRS;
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Wouldn't it instead make more sense to use the version packaged by the current board instead? E.g. us 1.6.2 while compiling for the edison and 1.6.4 for the 101? Automatically using a newer version just because some other platform uses it is bound to cause breakage at some point?

facchinm:
Right now the breakage occurs because 1.6.2 is picked :smile: 
If a new tool is unable to perform the operations the old one did, I believe the provider would change the name.
The board-specific tool resolution would be cool but I can't see a feasible way to do so (boards installed by Board manager would be ok because you could parse the json, while boards "installed" via sketchbook/hardware would be unresolvable (if not using the old shared method).

matthijskooijman:
> If a new tool is unable to perform the operations the old one did, I believe the provider would change the name.

Or it would increase the version number, to e.g. 2.0, (or perhaps 1.7, that might also break small things?). Or does this semver library already handle this?

> The board-specific tool resolution would be cool but I can't see a feasible way to do so (boards installed by Board manager would be ok because you could parse the json, while boards "installed" via sketchbook/hardware would be unresolvable (if not using the old shared method).

That should be ok - manually installed boards do not specify the tools they need, so they can only use whatever happens to be installed, so the newest version of any tools makes sense.

PR #4318, opened 2015-12-19
Stepper: avoid division by zero

kennethlimcp:
Check for inputs of 0 and set it to 1 as the minimum

Was added here for Particle devices: https://github.com/kennethlimcp/spark-stepper/pull/2

Signed-off-by: Kenneth Lim kennethlimcp@gmail.com


matthijskooijman:
This PR touches on the same library as #4257, and IIRC there was some change to prevent zero division there as well. Since merging this one probably complicates merging that one, perhaps this commit could be included in that PR, or at least merged after it.

Looking at the actual commit, I'm wondering what the motivation for it is? Passing a 0 speed or number of steps doesn't make sense, so it's not a valid input. On AVR, I think division by zero doesn't actually trigger any exception, so no really special handling is needed, but perhaps that's different on the Particle devices (or arm devices in general)?

PR #4325, opened 2015-12-21
[AVR] Fixed USB_SendControl for buffer size > 64

cmaglie:
This is the best implementation I've got so far, this add 52 bytes (of ~3900 total) to the empty sketch.

I have **not** tested with transfers > 64 bytes, @NicoHood may you give it a try?

related to #4317 
/cc @facchinm

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -425,31 +425,21 @@ static bool USB_SendStringDescriptor(const u8*string_P, u8 string_len, uint8_t f
 }
 
 //	Does not timeout or cross fifo boundaries
-//	Will only work for transfers <= 64 bytes
-//	Use USB_RecvControlLong for longer transfers
 int USB_RecvControl(void* d, int len)
 {
-	WaitOUT();
-	Recv((u8*)d,len);
-	ClearOUT();
-	return len;
-}
-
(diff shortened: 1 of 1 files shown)


NicoHood:
This was my fix:
https://github.com/NicoHood/Arduino/commit/71c88b6f8e0da338b3413699b74e57c11773f6f6

You might want to check if my version gives the same overhead.

@facchinm rejected it cause of the overhead. That's why I added the Long function, because I need the 64 byte fix. From theory it should work, but I cannot test right now.

cmaglie:
hi @NicoHood 

I remember that @facchinm has discussed this one briefly with me, probably at the time I didn't understood the relevance of this patch, so sorry to take so long to make up our mind!

I checked your patch it seems to be the best until now:

|  patch  | empty sketch size | delta  |
|-----|-----|------|
| no patch |  3916  |     0 |
| https://github.com/NicoHood/Arduino/commit/71c88b6f8e0da338b3413699b74e57c11773f6f6  |  3966  |   +50 |
| https://github.com/cmaglie/Arduino/commit/de5a13956ebb14b0db6ecceed797f693dd768311 |  3968  |  +52 |

So, I'm going to use you patch at https://github.com/NicoHood/Arduino/commit/71c88b6f8e0da338b3413699b74e57c11773f6f6

thanks!

NicoHood:
You might want to rename the variables, the names are horrible tbh by myself. But yeah, thx for adding this!

PR #4326, opened 2015-12-21
Fixed NPE in some rare combinations of JSON files

cmaglie:
The error triggered inside ContributioIndexer.mergeContributions() while trying to remove a platform:

    if (platform != null) {
      targetPackage.getPlatforms().remove(platform);
    }

the remove() method calls ContributedPlatform.equals() to find the element to remove but since the parentPackage fields are resolved *after* merging contributions, the equls() method will fail with
a NullPointerException.


sandeepmistry:
Looks good! It resolved an issue I was seeing this morning.

Issue #4320, opened 2015-12-19
Delay yield


vbextreme:
While writing a port official scheduler I realized that the delay() function has a bug.
this is official delay() function
```
void delay(unsigned long ms)
{
	uint16_t start = (uint16_t)micros();

	while (ms > 0) {
		yield();
		if (((uint16_t)micros() - start) >= 1000) {
			ms--;
			start += 1000;
		}
	}
}
```
This code assumes that the task called from delay lasts less than a millisecond , otherwise you can get up to an infinite loop.
simple solution:
```
void myDelay(unsigned long ms)
{
      if ( 0 == ms ) return;
      
      uint32_t sm = micros();
      uint32_t sM = millis();

      while( millis() - sM < ms - 1 ) yield();
      ms *= 1000;
      while( micros() - sm < ms - 1 );
}
```
It works with the same precision but should be free of bugs.

or 

```
void myDelay(unsigned long ms)
{
    uint32_t start = micros();

    while (ms > 0) {
        yield();
        while ( ms > 0 && (micros() - start) >= 1000) {
            ms--;
            start += 1000;
        }
    }
}
```

Testato:
+1

cmaglie:
good catch,

I prefer the second version (using `uint32` instead of `uint16` and the `while` instead of the `if`) because the sequence:

```C++
      uint32_t sm = micros();
      uint32_t sM = millis();
```

it's still subject to race conditions (millis() may advance when the execution is in the middle of the two instruction). Would you like to make a pull request for this one?

matthijskooijman:
I also prefer the second one (I think you edited your post to add it? I don't remember seeing it in the e-mail notification).

> it's still subject to race conditions (millis() may advance when the execution is in the middle of the two instruction). Would you like to make a pull request for this one?

What race condition exactly? You mention `millis()`, but the second version doesn't use `millis()` at all? I can't see any race condition here?

cmaglie:
> What race condition exactly? You mention millis(), but the second version doesn't use millis() at all? I can't see any race condition here?

The second version if fine, I'm referring to the first version.

If the millis() update ISR fires exactly in the middle of the two instruction:

```C++
      uint32_t sm = micros();
      uint32_t sM = millis();
```

you can get something like:
sm == 100999
sM == 101

while you really wanted:
sm == 100999
sM == 100

or:
sm == 101000
sM == 101

matthijskooijman:
[...]

PR #4349, opened 2015-12-27
fix delay/yield avr

vbextreme:
fix issues #4320 delay/yield, when the yield function takes more than a millisecond the delay fails.
Only for avr.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/wiring.c
+++ b/hardware/arduino/avr/cores/arduino/wiring.c
@@ -105,11 +105,11 @@ unsigned long micros() {
 
 void delay(unsigned long ms)
 {
-	uint16_t start = (uint16_t)micros();
+	uint32_t start = micros();
 
 	while (ms > 0) {
 		yield();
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Did you figure out the rebase? Or did you just start from scratch? :-)

Anyway, there's still two issues unresolved:
 - You have not yet improved the commit message
 - The indentation now uses, AFAICS, 8 spaces, while the original uses tabs. This means that a bunch of lines are marked as changed now, while they didn't actually change, and the indentation is no longer consistent in this file.

vbextreme:
ok?

matthijskooijman:
The diff looks perfect now, though the commit message shows the same message three times? I guess just the initial line might be sufficient here, combined with the diff it should be clear what the problem was and how it is fixed.

vbextreme:
I fight with "rebase" but finally I did it.

matthijskooijman:
Cool, looks good. @cmaglie, merge this?

Testato:
+1

PR #4358, opened 2015-12-28
Fix compile errors after renaming files, related to unsaved changes in the sketch

### LINKED ISSUE #4335: Renaming C to CPP gives errors
@matthijskooijman @testato @lmihalkovic 
https://github.com/arduino/Arduino/issues/4333#issuecomment-166728159

If you rename the c files to cpp (or vice versa) the other is kept in the temporary directly for compiling and sometimes leftover. Saving fixes this, editing again breaks it again. I have no test for this though.

matthijskooijman:
Fix #4335 by always deleting the temporary sketch copy after every build. With that, the name of that directory can be random again, instead of needing to be deterministic using a hash of the main sketch filename.

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -26,7 +26,6 @@ package processing.app;
 import cc.arduino.Compiler;
 import cc.arduino.CompilerProgressListener;
 import cc.arduino.UploaderUtils;
-import cc.arduino.files.DeleteFilesOnShutdown;
 import cc.arduino.packages.Uploader;
 import org.apache.commons.codec.digest.DigestUtils;
 import processing.app.debug.RunnerException;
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
If a sketch has many .cpp files, will this change force them to all be rebuild even if only the main .ino has changed?

matthijskooijman:
Nope, currently arduino-builder currently compares the contents of all files in this temporary directory with their copies in the build directory, and only copies/recompiles them if they actually changed (disregarding the timestamps on the files). See these comments for some more details:
 - https://github.com/arduino/Arduino/issues/4335#issuecomment-166876989
 - https://github.com/arduino/Arduino/issues/4335#issuecomment-166880856

This PR does not implement the additional improvement suggested in the first comment, it just fixes the problem without needing changes to arduino-builder.

cmaglie:
LGTM :+1:

Issue #4332, opened 2015-12-22
One wire library still in v2.3.0 - causes useless warning


vicnevicne:
The version of the OneWire library proposed by the library manager is still v2.3.0, which causes the IDE to raise the following warning on compile :

WARNING: Category '' in library OneWire is not valid. Setting to 'Uncategorized'

This fix (and others) is included in OneWire lib v2.3.1 which was released on Dec. 6th - https://github.com/PaulStoffregen/OneWire/releases - so the IDE should link to v2.3.1 now.

matthijskooijman:
I had a look at that repository, it looks like it has a proper tag, which also contains the proper version number in its library.properties file, so it should be picked up. I can't look into the server side system for errors, though.

vicnevicne:
Hi, Matthijs, and thanks for looking into this.
Do you know how the "standard" libraries are referenced in the IDE ? Is there a list of urls in a config file somewhere or is the list fetched online ? Could it be a caching issue ? 
Best regards. Vicne

cmaglie:
@vicnevicne 
it seems that the library has been updated in the library manager:

![onewire](https://cloud.githubusercontent.com/assets/423515/11978852/dd5f2268-a98e-11e5-823d-23e024b3081e.png)

maybe your PC is unable to update the `library_index.json` file?
(if you're on windows you can find it in `%LOCALAPPDATA%/Arduino15` folder)

vicnevicne:
Hi, and thanks for helping. 
That's weird : I can find the file you mention and gets updated by the IDE, but the contents still references only 2.3.0.
Here are the steps I just did :
1) Exit IDE
2) Delete the old library version. I admit I found no procedure about how to uninsall a lib so I just delete both :
"C:\Users\<me>\AppData\Local\Arduino15\staging\libraries\OneWire-2.3.0-github.zip"
and :
"C:\Users\<me>\Documents\Arduino\libraries\OneWire2.3.0\"
3) Restart IDE
4) Sketch > Include library > Manage libraries > Filter on "OneW" > Click on OneWire 
(I'm just getting an "Install" button with no choice of version)
5) Click "Install"
=> Shows version 3.2.0 :

![image](https://cloud.githubusercontent.com/assets/842233/11983691/a482de9a-a9b5-11e5-808b-ff24dfeef60b.png)

The library_index.json file has been updated (see timestamp) but only references v2.3.0 :

[...]

Issue #4272, opened 2015-12-07
Arduino15/library_index.json.tmp.gz is not refreshed 


MichaelJonker:
Hello, I see still an issue with the library manager. This was also reported in https://github.com/arduino/Arduino/issues/4139

I noticed the issue is still there after I released a new version of a contributed library, which not did show up in the library manager. I tried this with the latest nightly build (of 20151204 02:17).

The symptoms are still the same, the file library_index.json.tmp.gz is not refreshed and I do not see the new libraries versions.

After renaming the file library_index.json.tmp.gz to something different, the IDE loads a new library_index.json.tmp.gz version and I see the new library updates in the library manager.
When I delete the newly copied library_index.json.tmp.gz and rename the old one back again, then I only see the old versions in the library manager.

Michael Jonker

vicnevicne:
I checked the code. The issue is that the downloaded gz file is never deleted.
Consequently, when a new version is downloaded again, the "tmpFile.renameTo(gzipTmpFile)" fails (at cc/arduino/contributions/GZippedJsonDownloader.java:56 ).
The result (false) is not checked, so the process goes on and uncompresses the old version over and over again.

The fix is to delete the previous file by adding
```java
      gzipTmpFile.delete();
```

As the file is named "tmp", I guess the intention of the author was to delete it once uncompressed so I would add the delete right after the call to decompress(), as follows : 

```java
      new JsonDownloader(downloader, gzippedUrl).download(tmpFile, progress, statusText, progressListener);
      File gzipTmpFile = new File(tmpFile.getParentFile(), GzipUtils.getCompressedFilename(tmpFile.getName()));
      tmpFile.renameTo(gzipTmpFile);
      decompress(gzipTmpFile, tmpFile);
      gzipTmpFile.delete();
```

Alternately, to ease debugging, the delete could also be called just before the rename, to always keep one version on the disk, as follows :

```java
      new JsonDownloader(downloader, gzippedUrl).download(tmpFile, progress, statusText, progressListener);
      File gzipTmpFile = new File(tmpFile.getParentFile(), GzipUtils.getCompressedFilename(tmpFile.getName()));
      gzipTmpFile.delete();
      tmpFile.renameTo(gzipTmpFile);
      decompress(gzipTmpFile, tmpFile);
```

Can someone review and include that fix ?

KR. Vicne

PR #4361, opened 2015-12-29
Remove temporary files when updating library_index.json

cmaglie:
This is a tentatve fix for #4272 #4332, following @vicnevicne hint.

I'm not able to reproduce the bug, @vicnevicne @MichaelJonker if you can reproduce it, by any chanche, can you try the fix? (ArduinoBot should make a build in a moment...)

### DIFF
--- a/arduino-core/src/cc/arduino/contributions/GZippedJsonDownloader.java
+++ b/arduino-core/src/cc/arduino/contributions/GZippedJsonDownloader.java
@@ -53,8 +53,13 @@ public class GZippedJsonDownloader {
     try {
       new JsonDownloader(downloader, gzippedUrl).download(tmpFile, progress, statusText, progressListener);
       File gzipTmpFile = new File(tmpFile.getParentFile(), GzipUtils.getCompressedFilename(tmpFile.getName()));
+      // remove eventual leftovers from previous downloads
+      if (gzipTmpFile.exists()) {
+        gzipTmpFile.delete();
+      }
       tmpFile.renameTo(gzipTmpFile);
       decompress(gzipTmpFile, tmpFile);
+      gzipTmpFile.delete();
     } catch (Exception e) {
       new JsonDownloader(downloader, url).download(tmpFile, progress, statusText, progressListener);
     }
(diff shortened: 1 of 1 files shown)


vicnevicne:
@cmaglie I'm confirming the fix is OK. Thanks.
In fact, the issue is easy to reproduce : use the library manager in 1.6.7 (or a version without your fix) and a file "C:\Users\<me>\AppData\Local\Arduino15\library_index.json.tmp.gz" gets created but not deleted. Then exit, let one minute pass, restart the same procedure and you'll notice the timestamp of that .gz file does not change (the file is not replaced). So the same file gets unzipped over and over.
With your fixed version, the .gz file is correctly deleted.
Here is the video with 2 successive launches of 1.6.7 and then 2 with your 1.6.8 : http://screencast.com/t/a8hGl2sd1S
So it's all good for me.
@matthijskooijman Could be done indeed, but I guess the separate download/unzip process eases debugging (well, maybe debugging is the reason why the delete was forgotten)
For me anything goes as long as it works ;-)

cmaglie:
@vicnevicne 
thanks for testing and for the screencast!
It seems that the `File.renameTo(..)` is not able to replace the destination file on Windows (while it does on linux/unix, that's why I wasn't able to reproduce here).

@matthijskooijman 
right, I've added another commit and removed the extra rename. I'll check the result with @vicnevicne's procedure and merge straight away if everything is ok.

matthijskooijman:
@cmaglie The extra commit looks good, that's exactly what I meant indeed.

vicnevicne:
@cmaglie : I didn't know renameTo() does overwrite on Unix/Linux, but I know for sure it doesn't on Windows indeed.
The doc confirms that "Many aspects of the behavior of this method are inherently platform-dependent: [...] it might not succeed if a file with the destination abstract pathname already exists" - http://docs.oracle.com/javase/8/docs/api/java/io/File.html#renameTo-java.io.File-

And the improvement following @matthijskooijman 's suggestion is even better. I checked that the downloader also requires the target file to be deleted beforehand, and it's the case (otherwise it appends to it, which gets detected as a corrupt download and forces the fallback to download the non-gz version).

So +1 for the last version, avoiding renameTo() altogether.

PR #4362, opened 2015-12-29
Bug fix and rationalization of Boards/Libs installer GUI

cmaglie:
This PR fixes a deadlock experienced on some recent linux distros and rationalize the code of the Boards and Libraries installers GUI.

Tested on linux. To be tested on Mac & Windows.


sandeepmistry:
I tested the following on OS X 10.11.1, and observed no issues:
 * Library Manager: add, update
 * Board Manager: add, update, remove

cmaglie:
I've tested on Windows 10 and OS X 10.10 as well. I'm going to merge this one so it becomes available through hourly-builds.

The deadlock on linux seems related to the recent update of `libatk-wrapper-java-jni`, current user of Ubuntu 15.10 may be affected.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=798131
https://bugs.launchpad.net/ubuntu/+source/openjdk-8/+bug/1510009

Issue #3441, opened 2015-06-30
Save expand/collapse code folding!


MrKp91:
I've recently found under preference the ability to have code folding which is very helpful but every time I switch tabs everything is expanded.

This is very annoying specially when you have  collapse specific parts of your code and go check something on another tab just to have it all expanded.

ALSO, it would be nice if there's a shortcut to collapse/expand everything.

Thanks!

ffissore:
That will take a while to get solved. The root cause is that we use the same editor for every tab, replacing its content every time your switch one. This resets its previous settings, like folded code blocks.
This is a legacy behaviour, which forced us event to code a customized version of the new editor (RSyntaxTextArea)
Solution is to stop recycling the editor and create new ones for each tab. Once done, this issue should autosolve

lmihalkovic:
This is another example of a very low hanging fruits in this project (just added the feature to my branch despiteand it works fine), and blaming the editor-sharing is just a bad excuse for not implementing the feature.

There are fundamentally not technical justification for the customized version of RSTxtArea, and this one owes more to poorly structured code than to real requirements. That is one of the issues pleaguing this industry: "It can't be done" used liberally in place of "I don't know how to do it".

lmihalkovic:
just in case... this is to prove that it is indeed extremely simple to implement:

```
  @Override
  protected void configureFolding(SketchFileType type) {
    textarea.setCodeFoldingEnabled(true);
    textarea.getFoldManager().addPropertyChangeListener(new PropertyChangeListener() {      
      @Override
      public void propertyChange(PropertyChangeEvent evt) {
        if(evt.getPropertyName() == FoldManager.PROPERTY_FOLDS_UPDATED) {
          @SuppressWarnings("unchecked")
          List<Fold> list = (List<Fold>)evt.getNewValue();
          state().foldings = list;
        }        
      }
    });
  }  

  @Override
  public void restoreState() {
    List<Fold> list = state().foldings;
    textarea.getFoldManager().setFolds(list);
  }  
```

lmihalkovic:
@cmaglie @matthijskooijman   should you have the availability within the near future, it would likely be very appreciated if you'd look into this one. The code does work *as-is* in my fork, but will need some adapting for your codebase. I gave a short description of what it boils down to [here](https://forum.arduino.cc/index.php?topic=366459.0), there is really nothing more to it. (oops.. and yes, surprisingly it works despite the very stupid *==* which I will fix).

[...]

PR #4363, opened 2015-12-30
Editor cleanup

matthijskooijman:
This pullrequest contains a large number of cleanups and code refactorings concerning the editor part of the IDE, which takes care of loading files, showing an editor and tabs, saving files, modifying files, etc.

The original goal of this PR was to use separate RSyntaxTextArea instances for all tabs, preventing some issues resulting from sharing a single instance, and cleaning up the code around this change. The second goal was to clean up the sketch classes (e.g. Sketch, SketchData, SketchCode and SketchCodeDocument, of which it was not really clear what purpose each served). Now, there is SketchController, Sketch and SketchFile, with better defined purposes.

The commits in this PR jump around the code base a bit, since often one cleanup either required another one to happen before, or allowed one after. Also, some commits are simply small cleanups of things I came across while going through the code. It might be possible to reorder the commits to group related commits, but there are a lot of dependencies between them (both conceptual or just textual), which would make this take a lot of time, which does not seem worth the investment.

Some commits do not seem to make much of an improvement in themselves (or even introduce new ugly code), but these usually prepare for a later cleanup (which often removes the ugly code again).

This PR should mostly leave behavior unchanged, only changing the code. In some cases there are some corner cases which are handled better or differently, and some small things have improved as result of some cleanups as well.

A large part of this code could be further improved and decoupled by applying the Observable pattern, or some other event handling mechanism (e.g. to let the EditorHeader get notified when a file gets renamed, so it can update, instead of having the rename code explicitely refresh the EditorHeader). This is something for a future PR, though, since I haven't yet figured out what, if any, framework would best for doing this easily.

The first two commits of this PR come from #4358, so use that PR for comments about these commits.


lmihalkovic:
It is sad to see some of the best things in the original codebase going away without even having been noticed for how useful they were. To be clear, many of the improvements i made, at times with minimal amount of work, would become the bigger efforts that some of the core team mentionned in previous comments. The positive news is that you will no longer have to deal with my comments as this codebase is no longer worth my time.

matthijskooijman:
I just got the test suite working, which showed a few tests were failing due to a problem with the caret position saving and undo menu items. I added two commits to fix that, which I'll later merge into the appropriate earlier commits (but leaving them separate is easier to review).

PaulStoffregen:
@lmihalkovic - You speak of "many of the improvements i made", but all your public activity appears to only rhetoric, without *any* real code contributions.

Chris--A:
@PaulStoffregen, that is a job for other people, apparently
http://forum.arduino.cc/index.php?topic=366459.msg2531345#msg2531345

lmihalkovic:
@PaulStoffregen  regarding the debugger you are writting, the impact these changes will have on your work (once you've evaluated it), I thought you'd be one of the people understanding
```
Eventually, I will probably be the guy who implements this.....

But don't let that slow or discourage you. If you want it, this is open source, so 
you can do it. Only a small matter of programming is required!
```

@Chris--A  I pasted the exact working solution to the folding issue from my codebase (very different from yours) and even explained what should not need any explanations. Clearly one of you can write the handfull of lines of code to adapt it to your codebase

@matthijskooijman  it is only in the software industry that people are denied an opinion unless they are willing to redo the work. Personally when someone tells me I am clearly not considering the full impact of what i am doing, I have a critical look at my work, not the messenger. This patch closes the doors it closes regardless of who makes the evaluation, and I live comfortably with the knowledge that for the issues I saw, other more experienced developpers will undoubtedly see more. Issues live in source code, not in the *eyes of the beholder*

vicnevicne:
@matthijskooijman The changes you propose seem great, but their number looks massive to me... What would be the best way to go forward ? My contribution might not be of very high value as I'm new to the IDE code, but I could review them all one-by-one, learning along the way. Would it help ? If not, how can I (how can we) help ?

lmihalkovic:
Some perspective might be useful. 

In june 2015 #3441 was added to this repo regarding the editor loosing the state of foldings when switching around between tabs. An assessment was made by a member of the core team that:
```
That will take a while to get solved. The root cause is that we use the same editor for every tab, replacing its content every time your switch one. This resets its previous settings, like folded code blocks.
This is a legacy behaviour, which forced us event to code a customized version of the new editor (RSyntaxTextArea)
Solution is to stop recycling the editor and create new ones for each tab. Once done, this issue should auto solve
```
For the better part of 6 months this remained the only voiced opinion on the topic and nothing further happened. 

Reading the comment I thought it was clearly the wrong assessment, which was also missing an obvious opportunity. To pre-empt another of @PaulStoffregen's *rethoric only* comments, I pasted my code as-is, even adding a redundant explanation to further reduce the efforts for anyone adapting my solution to this codebase.

#3441 is a small example of what I regularly refer to as *low hanging fruits*. Regardless of personal opinions, it might prompt at least one person to start thinking that maybe even just out of sheer luck, what I say about the existence of many such opportunities to drastically improve this application at low cost might have some degree of truth, that some of them might actually be as trivial as #3441 is, and that I might even have a reasonable enough sense of software development to actually be meaningfully inviting you to consider some of the further consequences of this patch... regardless of when I publish my own fork.

vicnevicne:
@lmihalkovic I read you, Laurent, and I can understand your frustration that it took a long time to fix a bug you proposed a solution for. But the fact is that we now have a proposed version that fixes that bug (agreed, not using your code but as part of a more massive refactoring), so I think the only constructive move is to help having it merged into the tree. If there is a good reason for not merging it (regression), let's find it and fix it. What best way to start a new year ? :-). KR. Vicne

matthijskooijman:
@vicnevicne, review of these commits would certainly be helpful. Even if you're not familiar with the code, just flagging things that seem strange or wrong to you is fine (easiest is to just comment on the commits themselves). Some things might look wrong, but will be fixed later, but don't let that stop you from flagging things, it's probably counter-productive to read through *all* commits before commenting :-)

@lmihalkovic as for the fix you proposed, I'll add a coment to #3441 about that, to keep the discussion about that in that issue.

lmihalkovic:
@vicnevicne i appreciate your comment and the effort it reflects to try to touch on these issues.
The code I shared in #3441 was never directly usable because the editor it applies to in my *currently private* fork has diverged from the official arduino editor (cleaning the original editor code was required to support ASM, markdown, images, libraries.properties, keyword.txt, single-compressed-file sketch, and asciidoc). I only pasted my working code to avoid a long debate and show how trivial the solution to the code folding problem is and always was, when compared to the long explanation shared by the core team. My second goal in referencing #3441 was to highlight that there are more cases like that in the codebase where people view complications that are in IMHO not there. 
As for my position on this proposed series of changes, IMHO it currently contains more harm than good as it eliminates, seemingly without acknowledging their strength or IMHO offering any improvement, some of what I saw as strong points in the original codebase and which offer flexible points of enhancements. I shared this view as an invitation to consider any deep refactoring in the broader context of where this application is going, possibly with the involvement of @mbanzi, @damellis @cmaglie and the rest of the core team's vision for both Arduino products, as opposed to conducting it as an isolated musical chairs exercise where code will randomly drop when the music stops.

cmaglie:
> As for my position on this proposed series of changes, IMHO it currently contains more harm than good as it eliminates, seemingly without acknowledging their strength or IMHO offering any improvement, some of what I saw as strong points in the original codebase and which offer flexible points of enhancements.

@lmihalkovic 
can you be more specific on that? what are the srtong points on the original codebase that are being removed and which enhancements they offer?

lmihalkovic:
IMHO, everything starts with the vision Arduino LLC has for its software products. You have an existing multi-year investment in IDE and a growing stake in Create. If the immediate future is to let IDE go into sunset mode and let the feature gap grow, then nothing can be wrong about any changes and this is as good a patch as any. If on the other hand there are some features that Arduino LLC would like to schedule as complements or stepping stones or even vendor locking into Create, then any big refactoring should be evaluated in terms of how much closer it gets you to these goals, rather than how easy or trend compliant it is. 
Without presuming anything about where Arduino LLC wants to go, you might consider that without any drastic alterations, the existing split of responsibilities between SketchCode, SCodeDoc, Sketch, STArea, EditorHeader, EditorConsole and Editor (and some cleanup of the existing boundaries) allowed support for adding/editing local libs, viewing the source code for any system lib, using single file sketches, and viewing/editing most file types related to the platform, in under 2 months. The list is longer but my path is more akin to a 2.0 than a dot increment. IMHO some of the proposed redifined complings/decouplings are bound to stand in your way, in a similar fashion to how the excessive decoupling of Compiler from Editor is currently holding several issues.
That's the best summary I can offer you at this time.

matthijskooijman:
I just added three more commits that fix up some things in previous commits. I'll make sure to merge these into the commits they fix before merging this PR, but I'll keep them separately for now for easier review.

matthijskooijman:
I rebased this PR on top of current master, now that #4358 is merged. This also ran into a small conflict with #4444 which was easy to fix. No meaningful code changes were made to any of the commits.

facchinm:
Rebased and merged

PR #4377, opened 2016-01-02
Half duplex support

peteruithoven:
Adds a parameter to the Software Serial library to enable a half-duplex mode. This allows using 1 line for RX and TX. 
It does this by setting the transmit pin by default to input, with the pull-up set. When transmitting, the pin temporarily switches to an output until the byte is sent, then flips back to input. When a module 
is receiving it should not be able to transmit, and vice-versa. 
Based on the following fork: https://github.com/nickstedman/SoftwareSerialWithHalfDuplex

The circuit described in the following post makes sure the writes aren't read back in: http://nerdralph.blogspot.ca/2014/01/avr-half-duplex-software-uart.html


peteruithoven:
It's been 4 years, I hope someone else can pick this up.

Issue #1344, opened 2013-04-02
IDE Bug: "Verify Code After Upload" preference does not work


cmaglie:
Relevant topic:

http://arduino.cc/forum/index.php/topic,154141.msg1156478.html

Original text:

```
I assume we post bugs here... if not, someone please point me to the right place!

Regardless of the state of the "Verify Code After Upload" checkbox
in Preferences, the code sent to the Due is always verified.  A little bit 
of troubleshooting on my part revealed that the culprit is the '-v' flag 
that is always passed to 'bossac':

bossac.exe --port=COM30 -U false -e -w -v -b [temp_filename] -R

If this command is run manually without the -v flag, verification does 
not occur.

This bug occurs on Windows in both the beta and the nightly build,
I can't vouch for Linux or MacOS.  The verification takes several 
seconds to run even on a small sketch so it's nice to be able to turn it off.
```

mathewjury:
This issue is still current in 1.6.4. The above recommendation by ergodic00 didn;t help. Compiling for a Nano board if thats of any assistance.

Mat.

hallard:
Hi there, I confirm this bug is still present, but I don't thing it's related to -v flag (which is verbose), in fact in the IDE when this option is unchecked the switch -V (upper case) should be added to command line, I just tested by hand and it works.
Pay attention, I 1st put this -V at the end of the command line and it did not work, then I tried putting it just after avrdude it worked, see below

```
$ avrdude -V -CS:/blablabla/avrdude.conf -patmega328p -cusbtiny -Uflash:w:C:/Users/CHARLES/AppData/Local/Temp/build8905307267913391739.tmp/bootloader_test.cpp.hex:i
avrdude.exe: AVR device initialized and ready to accept instructions
Reading | ################################################## | 100% -0.00s
avrdude.exe: Device signature = 0x1e950f
avrdude.exe: NOTE: "flash" memory has been specified, an erase cycle will be performed
             To disable this feature, specify the -D option.
avrdude.exe: erasing chip
avrdude.exe: reading input file "C:/Users/CHARLES/AppData/Local/Temp/build8905307267913391739.tmp/bootloader_test.cpp.hex"
avrdude.exe: writing flash (13596 bytes):
Writing | ################################################## | 100% 3.34s
avrdude.exe: 13596 bytes of flash written
avrdude.exe done.  Thank you.
```

gh-megabit:
[...]

PR #4385, opened 2016-01-04
Upload Verify option added

gh-megabit:
This change picks up the upload.verify preference from the preferences dialog/preferences.txt and allows you to use this with your USB/Serial upload tool of choice. Configuration is already added for avrdude. 

It uses PreferencesMap.get to avoid an exception if the config is not found in platforms.txt - this should avoid problems with other cores.

### DIFF
--- a/arduino-core/src/cc/arduino/packages/Uploader.java
+++ b/arduino-core/src/cc/arduino/packages/Uploader.java
@@ -69,6 +69,7 @@ public abstract class Uploader implements MessageConsumer {
   }
 
   protected final boolean verbose;
+  protected final boolean verifyUpload;
 
   private String error;
   protected boolean notFoundError;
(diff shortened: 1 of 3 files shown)


gh-megabit:
This resolves #1344

matthijskooijman:
This PR looks good to me, it implements this option in the same way as the verbose option. AFAICS, this would also be backward compatible with older and third-party cores, which is also a good property.

Did I remember correctly that there was already a checkbox in the preferences window for this? If so, the commit message should probably mention this.

I believe this modifies the AVR core to support this option, but the SAM (and SAMD) cores would still need to be updated, right?

gh-megabit:
Yes this implements the "Verify code after upload" option - IIRC it was working before 1.5.x but wasn't implemented in the major update. The option is there on the preferences dialog but before this commit it isn't used anywhere. 

The SAM and SAMD cores can be updated if this commit is going to be merged. They won't benefit much from the change as the upload speed is so much better.

gh-megabit:
Will resubmit with better commit details

steronydh:
Still broken in 1.6.9!

PR #4395, opened 2016-01-06
Use "$@" instead of bare $@ to protect arguments

fumiyas:
Bash expands bare $@, and then do word-splitting and file-globing. It breaks command-line arguments for arduino if command-line arguments for the wrapper script contain $IFS, '*', '?' and etc. characters.

Use "$@" in this case.


matthijskooijman:
This change looks good to me.

facchinm:
Great, thank you for spotting this!

PR #4442, opened 2016-01-17
Use of logo in Readme.md

ChathunKurera:


matthijskooijman:
Thanks for your contribution, but I'm not sure what you're trying to achieve exactly? It seems you replaced the "Arduino" name in the README with a logo? Why do you think this is important? Frankly, I'm not sure if this is much of an improvement, also because the logo itself is quite tiny and a bit fuzzy...

NicoHood:
12 commits, not bad.

PR #4447, opened 2016-01-18
Due SAM Core -Remove compiler Warning and improve readability plus simple documentation fix

techpaul:
Discovered a core library warning of signed/unsigned comparison, which at first wa confusing due to font as one variable being 'l' lower case L, actually looked like a number one ('1'). Discovered this temporary variable was actually unsigned being compared to unsigned buffer variables.

This can be tested and found compiling using ALL warnings on Arduino 1.6.7 and Due Board Manager V1.6.6,. Noting that repository has more up to date Stream.cpp to avoid another compiler error that can be seen. Simple empty sketch (empty setup and loop) will show this

This is simple fix in UARTClass.cpp to

1/ Correct temporary variable to signed to match other variable and other similar comparisons in module.

2/ Change variable name to nextWrite for better readability

Other change is to remove questionable wording comment for RingBuffer.h, that implies no one knows what the definition is for.


sandeepmistry:
@techpaul can you please remove the ```hardware/arduino/sam/cores/arduino/UARTClass.cpp~``` file. It looks like a temporary file from your editor.

sandeepmistry:
Also, same comment for ```hardware/arduino/sam/cores/arduino/RingBuffer.h~```

techpaul:
Will do thought it was on exclude

techpaul:
Should be visible now

sandeepmistry:
@techpaul thanks!

I've cherry picked https://github.com/arduino/Arduino/commit/fa0d5808462a435c7178121585f90668223e017d.

PR #4451, opened 2016-01-18
Small fix in Wire.cpp

BrentWilkins:
Removed implicit conversions from unsigned, to signed, back to unsigned.


matthijskooijman:
This change looks good to me, thanks!

sandeepmistry:
@BrentWilkins thanks!

Issue #4460, opened 2016-01-21
Unexpected behaviour of print(char, HEX) and print(int, HEX)


Lyadis:
```
char c = 255;
Serial.print(c, HEX);
```
produces
```
FFFFFFFF
```
This is due to the fact that `print(char c, int base)` convert char c (1 byte) into an unsigned long (4 bytes). As char is signed 255 is in fact -1(Two's complement), once written in 4 bytes it becomes FFFFFFFF. But as a user I would rather have FF than FFFFFFFF.

The problem exist with int (2 bytes) too

Here is a more complete example:

```
void setup() {
  Serial.begin(9600);
  int i = 1;
  Serial.print("i: ");
  Serial.print(i, 10);
  Serial.print(" | ");
  Serial.println(i, HEX);
  i = -1;
  Serial.print("i: ");
  Serial.print(i, 10);
  Serial.print(" | ");
  Serial.println(i, HEX);
  
  char c = 127;
  Serial.print("c: ");
  Serial.print(c, 10);
  Serial.print(" | ");
  Serial.println(c, HEX);
  c = 255;
  Serial.print("c: ");
  Serial.print(c, 10);
  Serial.print(" | ");
  Serial.println(c, HEX);
}

void loop() {}
```

output
```
i: 1 | 1
i: -1 | FFFFFFFF
c: 127 | 7F
c: -1 | FFFFFFFF
```

where I owuld rather have
```
i: 1 | 1
i: -1 | FFFF
c: 127 | 7F
c: -1 | FF
```

matthijskooijman:
One can argue that doing a hexadecimal print of a negative number isn't well defined at all, IIRC only the decimal (base 10) printing supports negative numbers currently. Still, I guess this is something that can be fixed - if the current behaviour is not well defined, might as well improve that. I'll have a look at your PR next.

Chris--A:
Your code is broken from the start, is it not?

`char c = 255;` ??

```C++
char a = 255;
unsigned char b = 255;

void setup() {
  Serial.begin( 9600 );
  Serial.println( a, HEX );
  Serial.println( b, HEX );
}

void loop() {}
```

Output:
> FFFFFFFF
FF

@matthijskooijman 
This is well defined behavior also.

As the signed bit pattern of `a` is -1 in a char, then the unsigned long gets the bit pattern which represents the same number as if it were signed (All 1's).

> If the destination type is unsigned, the resulting value is the least **unsigned integer** congruent to the source **integer**.

[...]

PR #4461, opened 2016-01-21
fixes arduino/Arduino#4460

nharrand:
Fixes arduino/Arduino#4460
(Problem with print(char c, int base) and print(int i, int base)

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -74,7 +74,7 @@ size_t Print::print(char b, int base)
   if(base == 10)
     return print((long) b, base);
   else
-    return print((((unsigned long) b) & 0x000000FF), base);
+    return print((unsigned char) b, base);
 }
 
 size_t Print::print(unsigned char b, int base)
@@ -87,7 +87,7 @@ size_t Print::print(int n, int base)
   if(base == 10)
     return print((long) n, base);
   else
-    return print((((unsigned long) n) & 0x0000FFFF), base);
+    return print((unsigned int) n, base);
 }
 
 size_t Print::print(unsigned int n, int base)
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -69,22 +69,30 @@ size_t Print::print(char c)
   return write(c);
 }
 
+size_t Print::print(char b, int base)
+{
+  if(base == 10)
+    return print((long) b, base);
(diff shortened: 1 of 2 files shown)


Lyadis:
Ok I fixed the indentation problem.
You were right about the unsigned char, and unsigned int. The problem came in fact from the fact that if you call `print(char c, int base)`, int is in fact `print(int i, int base)` which is called with a conversion char to int. 
I propose to fix it by adding a function `print(char c, int base)` without the implicit base=DEC in order not mess up with print(char c).
About the `& masking`, it seems to be required. I think when you do a conversion int to unsigned long, your value is first extended to 4 bytes then read as an unsigned long.

Lyadis:
About the cost, I think if both existing and new functions are used, we are looking at 82 bytes more. (If I am not mistaken)

matthijskooijman:
>  About the & masking, it seems to be required. I think when you do a conversion int to unsigned long, your value is first extended to 4 bytes then read as an unsigned long.

It's probably also solvable by first casting to unsigned int, and then to unsigned long. Could you try that?

> About the cost, I think if both previously existing and new functions are used, we are looking at 82 bytes more in the program space. (If I am not mistaken). (I calculated it by compiling the same program with both version of the library)

Hm, that sounds like much to me, but that's probably because there is an extra function now. How much is the difference for a program only printing a char, and for a program only printing an int?

Also, shouldn't the `println(int, int)` function also be adapted?

Lyadis:
After running the test on 
```
void setup() {
  Serial.begin(9600);
  char c = -1;
  Serial.print(c, HEX);
  Serial.print((int) c, HEX);
}
void loop() {}
```
It actually cost 84 more bytes to use the version propsed in this PR.
I think it is due to gcc optimization. When we have:
```
size_t Print::print(int n, int base)
{
  return print((long) n, base);
}
```
print(int, int) is in fact just a renaming of print(long, int) (with a conversion in the parameter).
I suspect that gcc just replace calls for print(int, int) and do not define a new symbol for the function. But when the function contains an if statment, I guess that gcc don't have enough information at compile time to optimize in the same way... This is just speculation.

About the masking compare to double conversion: Surprisingly double conversion actually cost 2 bytes per conversion.

About `println(int, int)`, as it relies on `print(int, int)`, it doesn't need any fixes.

Chris--A:
I posted a response  to the issue here.

I personally do not think this is needed at all. Signed to unsigned conversions are well defined and the behavior of sign bit extension is expected.

https://github.com/arduino/Arduino/issues/4460#issuecomment-174243539

PR #4463, opened 2016-01-21
Providing a more fair servicing of multiple clients by not picking lowest numbered socket in available()

sauerburger:
I modified `EthernetServer::available()` such that it will not always pick the
lowest numbered socket by introducing an internal index `int _sock` which is incremented by one every time `available()` is called. (I suppose the previous comment *XXX: don't always pick the lowest numbered socket.* was to be understood as a TODO)

### DIFF
--- a/libraries/Ethernet/src/EthernetServer.cpp
+++ b/libraries/Ethernet/src/EthernetServer.cpp
@@ -52,13 +52,17 @@ EthernetClient EthernetServer::available()
 {
   accept();
 
-  for (int sock = 0; sock < MAX_SOCK_NUM; sock++) {
-    EthernetClient client(sock);
-    if (EthernetClass::_server_port[sock] == _port) {
+  for (int i = 0; i < MAX_SOCK_NUM; i++) {
+    // increment _lastReturnedSocket to avoid returning the same socket again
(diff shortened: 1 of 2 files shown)


matthijskooijman:
This change looks useful to me. AFAIU, it allows for a more fair servicing of multiple connections right? Perhaps the commit message could expand on this a bit more.

matthijskooijman:
Your new commit looks good to me, I think the code is fine as it is. Your new commit message is also very good (except for a few spelling errors :-p)

Could you perhaps merge both commits into a single commit? That would lead to clearer history if this is merged.

sauerburger:
Ok. Well I have a few git related problems with that:
I merged the commits with git rebase, but as I expected I could not push them, so I pushed them to a new branch. Do you know if I can change source branch of this merge request? If thats impossible should I create a new merge request?

matthijskooijman:
You can use `git push -f` to force push and overwrite the existing branch. Github will then update the commits in this PR automatically.

sauerburger:
ok thanks.

agdl:
This issue was moved to arduino-libraries/Ethernet#33

agdl:
Sorry closed by mistake

Issue #4471, opened 2016-01-23
SerialUploader NullPointerException: Serial upload broken in 1.6.8


probonopd:
Using 1.6.8 Hourly Build 2016/01/19 08:26 and previous, uploading via the serial port seems to be entirely broken and leads to 

```
processing.app.debug.RunnerException
	at cc.arduino.packages.uploaders.SerialUploader.uploadUsingPreferences(SerialUploader.java:179)
	at cc.arduino.UploaderUtils.upload(UploaderUtils.java:78)
	at processing.app.Sketch.upload(Sketch.java:1187)
	at processing.app.Sketch.exportApplet(Sketch.java:1161)
	at processing.app.Sketch.exportApplet(Sketch.java:1133)
	at processing.app.Editor$DefaultExportHandler.run(Editor.java:2417)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at java.lang.String.replace(String.java:2240)
	at processing.app.helpers.StringReplacer.replaceFromMapping(StringReplacer.java:97)
	at processing.app.helpers.StringReplacer.replaceFromMapping(StringReplacer.java:89)
	at processing.app.helpers.StringReplacer.formatAndSplit(StringReplacer.java:37)
	at cc.arduino.packages.uploaders.SerialUploader.uploadUsingPreferences(SerialUploader.java:174)
	... 6 more
```

I can reproduce this on Linux and OS X.

probonopd:
I can reproduce this with __ESP8266__ and also with an __Arduino Pro Mini__, so I assume it is an issue with the Arduino IDE.

probonopd:
@me-no-dev points toward https://github.com/me-no-dev/Arduino-1/commit/b3c91ab85b6aa4dddc7e124733e2fa1084b13694#diff-db26f03d5207a3a0d217244f3de0950fR156 and suspects that `boardPort.getPrefs().get("iserial")` returns NULL, and should be changed to `boardPort.getPrefs().getOrExcept("iserial")` to trigger an exception and not fill the item.

me-no-dev:
here is a better link: https://github.com/arduino/Arduino/blob/master/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java#L156-L161

probonopd:
I can confirm that the change proposed @me-no-dev works for me.

cmaglie:
Fixed by #4472

dethoter:
[...]

PR #4472, opened 2016-01-23
Fix adding NULL value to the preferences when iserial is not defined

me-no-dev:
connected to: https://github.com/arduino/Arduino/issues/4471


probonopd:
I can confirm that the change proposed @me-no-dev works for me and would close #4471.

matthijskooijman:
This fix looks good to me, just tested it with a Pinocc.io board (which had the same problem).

@me-no-dev could you rebase your branch on top of the latest master to remove the merge commits? If you then force push to the branch used to create this PR, the PR should be updated to only contain the relevant commit, without needlessly making the commit history more complex.

cmaglie:

PR #4482, opened 2016-01-25
[WIP] Rework serial ports handling

facchinm:
Following the work with the native jni library for serial discovery and preparing for the new bootloaders release (#2474, we are stressing them really hard :smile: ) , this commit introduces a new concept (stateful ports) and fixes a number of existing bugs.
The most important patch (91b94c8) needs to be reviewed by a Java expert to check if there is any huuuuge mistake I probably made somewhere (scopes, static or things like that).

Comments appreciated :wink: 
@cmaglie  @matthijskooijman @NicoHood @PaulStoffregen


NicoHood:
Will the new bootloader just fix what we suggested or be totally different? I mean I see no difference in the sense of the bootloader. Just that its time to [do it](https://www.youtube.com/watch?v=ZXsQAXx_ao0) ;)

facchinm:
Hi Nico,
the sources for the WIP bootloaders/firmwares are [here] (https://github.com/facchinm/Arduino_avrusb_firmware).
They address all the issues found in the legacy ones and are based on LUFA15.
A more experimental version is coming soon with exciting new features :wink:

NicoHood:
Uhm I want to mention a few things:

##### 16u2 usb-serial
* If you update it, dont just update the LUFA version, please also add 2M support.
* Why not also add 16u2 reprogramming like HoodLoader2 does? Its already inside the core. Also you base the 32u4 on HoodLoader2, so why not make the 16u2 firmware a new bootloader?

##### 32u4 Bootloader
* I like the idea that you base the BootLoader on HoodLoader2 but there a a few points you have to know:
* The USB-Serial is still integrated. This will touch the Serial Pins of the Micro/Leo. People might not want to do this.
* It uses some very specific makefile optimizations which are not bad at all, but could be used different.
* With a removed Serial even more options are possible to enable as EEPROM support and all such things

So there are some things that I'd improve. If you want to chat with me about that outside github, for a faster communication, feel free to add me on [tox](www.nicohood.de). It would be nice to know what you are planning, maybe I can help you to give ideas/suggestions.

facchinm:
Hi Nico,
- 16u2: 2mbps full duplex support is not mandatory and we don't have any plan to integrate sketch loading to the 16u2 itself. 
- 32u4: I really enjoy the compiler optimizations added to Hoodloader :wink: The non-bootloader usb-serial part has already been removed but the led pulse added enough overhead so don't expect a lot of improvements (except improved upload speed and dual magic key location)

BTW, I couldn't find you on tox, but I'd really enjoy some chatting about these things!

NicoHood:
You might want to try the tox ID, not the short name. It worked for a lot of people before.

NicoHood:
Are you planning to add the double tab reset?
I am currently trying to fix the normal LUFA branch, so I need to know what you prefer. I personally like it more since you can stay inside the bootloader as long as you want and also reset the board faster and not wait 7s.

facchinm:
@ArduinoBot build this please

FalconFour:
No difference here. I was directed this way from http://forum.arduino.cc/index.php?topic=388498.msg2681245#new - that hopefully this pull is related to the 1.6.8 bug that causes all serial ports to be constantly open/closed. That issue is still present, now confirmed on multiple PCs with different OSes (first was W7 x64, this time is W10 x64). Because Arduino is opening the port (at all, in any way/shape/form), it toggles DTR pin and resets the Arduino board once/second. This is unique to Windows, as it's a "feature" of the FTDI driver that avrdude relies on for auto-resetting boards.

Opening the ports constantly also ends up with a 50/50 chance that when I try uploading a sketch or open Serial Monitor, it'll run into the port scanning in process and will give "access denied" against the already-open port used by Arduino.

I'm working with the Arduino Pro Mini (5V/16MHz) and FTDI Basic Breakout board from TinySine.

Can we completely avoid opening the serial ports?

facchinm:
@FalconFour , thanks a lot for reporting this.
The current code is not trying to open the port (of course) but, as reported by  @ozligia on #4700, it's accessing a registry entry once per second.
In theory, this operation shouldn't "open" the port but maybe there is some side effect we didn't consider. Also, the build you tested pauses the board scanning during upload (commit 	02a4200) so this shouldn't affect normal upload operations.

PR #4515, opened 2016-02-01
Increase wait for upload port timeout to 5s on all platforms

sandeepmistry:
OS X 10.11 seems to be slower (with Arduino Zero native port), increasing timeout to 5s on all platforms to keep things simple.

This helps address https://github.com/arduino/ArduinoCore-samd/issues/62.

### DIFF
--- a/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
+++ b/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
@@ -247,11 +247,10 @@ public class SerialUploader extends Uploader {
       Thread.sleep(250);
       elapsed += 250;
 
-      // On Windows, it can take a long time for the port to disappear and
-      // come back, so use a longer time out before assuming that the
-      // selected
-      // port is the bootloader (not the sketch).
-      if (((!OSUtils.isWindows() && elapsed >= 500) || elapsed >= 5000) && now.contains(uploadPort)) {
+      // On Windows and OS X, it can take a few seconds for the port to disappear and
+      // come back, so use a time out before assuming that the selected port is the
+      // bootloader (not the sketch).
+      if (elapsed >= 5000 && now.contains(uploadPort)) {
         if (verbose)
           System.out.println("Uploading using selected port: " + uploadPort);
         return uploadPort;
--- a/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
+++ b/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
@@ -136,12 +136,11 @@ public class SerialUploader extends Uploader {
           // it happened within 250 ms. So we wait until the reset should
           // have already occured before we start scanning.
           actualUploadPort = waitForUploadPort(userSelectedUploadPort, before);
-          
-          if (OSUtils.isMacOS()) {
-        	  // on OS X, if the port is opened too quickly after it is detected,
-        	  // a "Resource busy" error occurs, add a delay to workaround this
-        	  Thread.sleep(250);
-          }
+
+    	  // on OS X, if the port is opened too quickly after it is detected,
+    	  // a "Resource busy" error occurs, add a delay to workaround this,
+          // apply to other platforms as well.
+    	  Thread.sleep(250);
         }
       } catch (SerialException e) {
         throw new RunnerException(e);
--- a/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
+++ b/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
@@ -136,6 +136,12 @@ public class SerialUploader extends Uploader {
           // it happened within 250 ms. So we wait until the reset should
           // have already occured before we start scanning.
           actualUploadPort = waitForUploadPort(userSelectedUploadPort, before);
+          
+          if (OSUtils.isMacOS()) {
+        	  // on OS X, if the port is opened too quickly after it is detected,
+        	  // a "Resource busy" error occurs, add a delay to workaround this
+        	  Thread.sleep(250);
+          }
         }
       } catch (SerialException e) {
         throw new RunnerException(e);


Breidenbach:
Tested on Mac OS 10.11.3 (El Capitan)

Cannot upload to genuine Arduino Uno which does work with windows XL and which did work before El Capitan.
Bluetooth keyboard and mouse disabled.
Arduino resets, then continues running prior sketch.
No diagnostics given by IDE.
Tried again, with no visible response by Arduino.

Hal Breidenbach
(I’ll be happy to get diagnostics if you tell me how)

sandeepmistry:
@Breidenbach thanks for taking the time to try out this patch and provide feedback.

I would not expect any difference in behaviour compared to the latest [hourly build](https://www.arduino.cc/en/Main/Software) with an Arduino Uno board.

The ```waitForUploadPort``` method which was changed is only called for boards that have ```<board>.upload.wait_for_upload_port=true``` in [boards.txt](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/boards.txt) (typically only boards with USB CDC). This is not the case for the Uno board.

I tried it out with an Uno on my OS X 10.11.3 laptop and no issues:
```
Communication Device:

  Product ID:	0x0043
  Vendor ID:	0x2341
  Version:	0.01
  Serial Number:	74132343430351E03152
  Speed:	Up to 12 Mb/sec
  Manufacturer:	Arduino (www.arduino.cc)
  Location ID:	0x14200000 / 17
  Current Available (mA):	1000
  Current Required (mA):	100
  Extra Operating Current (mA):	0
```

sandeepmistry:
@tigoe if you have some time, could you please try out the latest build with a SAMD based board (on native USB) with and without a USB hub on your Mac.

I noticed OS X would report "Resource Busy" if a ``/dev/cu.usbmodem*`` device was opened quickly after the IDE discovered it. The latest commit adds an OS X specific delay.

tigoe:
Just the latest nightly? Will try next time I hit a network with my laptop

Tom Igoe
--
sent on the go. please forgive any terseness or typos

Breidenbach:
Tried 2 times with blink to Leonardo, 1 time with another sketch using serial output, none ran, but none disabled bluetooth keyboard or mouse.  Did get attached report but about excessive wake-ups.

Hal

sandeepmistry:
@tigoe sorry, this build specifically: http://downloads.arduino.cc/javaide/pull_requests/arduino-PR-4515-BUILD-515-macosx.zip (from https://github.com/arduino/Arduino/pull/4515#issuecomment-190322450)

The changes are not in master yet.

sandeepmistry:
@Breidenbach thanks for trying. I have no issues loading the blink example onto my Leonardo.

How does the behaviour you are seeing compare to the [latest hourly build](https://www.arduino.cc/en/Main/Software#hourly)?

Breidenbach:
The same.  I haven’t tried the hourly builds for a while, since they wiped out bluetooth to the keyboard and mouse and I had to keep rebooting to get them back.  I notice that the device is listed at the bottom of the screen, and the compile messages look good.

As I’m writing this, I see that the device is listed as bluetooth.  (screen shot 1)  I ran it again with the leonardo device (screen shot 2) and lost the keyboard and mouse.

On the second try the L light pulsed 14 times, the the RX light pulsed twice, then nothing.

Running Yosemite 

Hal

tigoe:
@sandeepmistry, I tried with a MKR1000 and it worked great, uploaded perfectly every time. Even when the serial monitor was open. I also tried unplugging the board with the serial monitor open then plugging it back in and uploading, and it worked perfectly. 

Tried the same thing with a 101 board, and the upload worked fine there as well. If I had the serial monitor was open and I unplugged the board, I got "board is not available" after a few seconds, but upload continues to work fine.

I also have a Feather M0 on me, and wanted to test with that, since I had a terrible time with the Feathers yesterday in a workshop. Sometimes I had to hit upload ten times, using 1.6.7. I don't know what's different about their bootloader from the standard M0 bootloader, but there is definitely a difference. But when I opened the PR4515 build 515 version, the Feather board definitions disappeared, and I couldn't install them via the boards manager. As a matter of fact, all my third party boards had disappeared. I didn't test too far on that, though. It seems like it's only looking at the arduino-beta folder in Arduino15/packages, and none of the other folders in packages/. Is that the case?

I have no hub with me right now, sorry, and no boards other than the three I mentioned.

sandeepmistry:
@Breidenbach thanks for the info. I suggest you create a forum post about your issues. As they are outside the scope of this pull request. (Btw, when you reply via email Github does not show any of your attachments).

sandeepmistry:
@tigoe great, thanks for trying! We'll hold off merging this until we get feedback on behaviour with a USB hub. Both with and without a hub, are reliable for me. 

I'll look into why the other board manager boards aren't visible with this IDE build. It's based on an hourly build from yesterday. I'm curious to see if it improves sketch uploads to the Feather M0 on a Mac.

sandeepmistry:
@tigoe I had no issues with seeing the Feather M0 board when adding via the board manager and this IDE build:

screen shot 2016-03-01 at 10 00 47 am

Could you please try the latest [hourly build](https://www.arduino.cc/en/Main/Software#hourly), to see if it has the same issue.

tigoe:
Once i eliminated the arduino-beta folder I was able to see it fine as well.

With the Feather M0, I’m still getting 

No device found on cu.usbmodem1421

about every other upload. FWIW, I’m also getting this:

Warning: platform.txt from core 'Adafruit SAMD Boards' contains deprecated recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{build.path}/{archive_file}" "{object_file}", automatically converted to recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{archive_file_path}" "{object_file}". Consider upgrading this core.

Could that be related to the upload issue? Maybe there’s something in their version of the M0 board definition that needs updating?

sandeepmistry:
@tigoe 

> Could that be related to the upload issue?

The warning is unrelated to the uploading in this care.

> Maybe there’s something in their version of the M0 board definition that needs updating?

I suspect there is something up with the bootloader, the Zero/MKR1000 bootloader seems slower than the Leonardo's for USB CDC to initialize. The Adafruit bootloader appears to be even more slower at this.

In my latest change, I added a 250ms delay after the ```/dev/cu.usbmodem1421``` device is detected by the IDE, so that we don't launch the ```bossac``` upload tool immediately.

The "No device found on cu.usbmodem1421" error message from ```bossac``` is a bit misleading, as it means either ```/dev/cu.usbmodem1421``` does not exist, or there was a error opening the device. In this case, "Resource Busy" because the USB device isn't quite setup, even though OS X has  ```/dev/cu.usbmodem1421``` available on the file system.

I'm not too keen on bumping the delay now, having the IDE try to open the port until it is non-busy might be a better option.

sandeepmistry:
@facchinm @cmaglie please review.

This resolves sketch upload issues on the MKR1000 and OS X for myself and @tigoe, both with and without a USB hub.

facchinm:
Hi @sandeepmistry ,
I'd remove the ```if (OSUtils.isMacOS())``` part when reopening the port since a small delay is useful on all platform (included Linux).
Or, alternatively, if you could test the behaviour using https://github.com/arduino/Arduino/pull/4482 it would be great, since in [this](https://github.com/facchinm/Arduino/commit/91b94c8e634af96f73980c7fdb53cf01c661d7ac) commit there is a similar delay added for all platform

facchinm:
The second part, instead, looks really good :smile:

sandeepmistry:
@facchinm thanks for reviewing!

I've applied your suggestion from https://github.com/arduino/Arduino/pull/4515#issuecomment-194736891 in https://github.com/sandeepmistry/Arduino/commit/7cb1399381219952f752f3ea43ad1d83f2139fee.

Unfortunately https://github.com/arduino/Arduino/pull/4482 still fails every second sketch upload with the MKR1000 + USB, with the following ```bossac``` error:

```
An error occurred while uploading the sketch
No device found on cu.usbmodem14221
```

don:
This fixes timeout problems for me on OS X 10.10.5 with MKR-1000.

PR #4518, opened 2016-02-03
Support extra parameter on `attachInterrupt()`

paulo-raca:
When writing an arduino library, it is difficult to connect interrupt handlers with the component instance that should be notified.

In C, the common pattern to fix this is to specifying a `void*` parameter with the callback, where the listener can store any context necessary.

This patch adds a new function, `attachInterruptParam()`, which does exactly.


paulo-raca:
Wrong branch. Duh!

PR #4519, opened 2016-02-03
Support extra parameter on `attachInterrupt()`

paulo-raca:
When writing an arduino library, it is difficult to connect interrupt handlers with the component instance that should be notified.

In C, the common pattern to fix this is to specifying a `void*` parameter with the callback, where the listener can store any context necessary.

This patch adds the new function `attachInterruptParam()` to implement this pattern.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Arduino.h
+++ b/hardware/arduino/avr/cores/arduino/Arduino.h
@@ -148,6 +148,7 @@ void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val);
 uint8_t shiftIn(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder);
 
 void attachInterrupt(uint8_t, void (*)(void), int mode);
+void attachInterruptParam(uint8_t, void (*)(void*), int mode, void* param);
 void detachInterrupt(uint8_t);
 
 void setup(void);
--- a/hardware/arduino/avr/cores/arduino/WInterrupts.c
+++ b/hardware/arduino/avr/cores/arduino/WInterrupts.c
@@ -35,7 +35,7 @@
 static void nothing(void) {
 }
 
-static volatile voidFuncPtr intFunc[EXTERNAL_NUM_INTERRUPTS] = {
+static volatile voidFuncPtrParam intFunc[EXTERNAL_NUM_INTERRUPTS] = {
 #if EXTERNAL_NUM_INTERRUPTS > 8
     #warning There are more than 8 external interrupts. Some callbacks may not be initialized.
     nothing,
@@ -65,11 +65,23 @@ static volatile voidFuncPtr intFunc[EXTERNAL_NUM_INTERRUPTS] = {
     nothing,
 #endif
 };
+static volatile void* intFuncParam[EXTERNAL_NUM_INTERRUPTS];
 // volatile static voidFuncPtr twiIntFunc;
 
-void attachInterrupt(uint8_t interruptNum, void (*userFunc)(void), int mode) {
+void attachInterrupt(uint8_t interruptNum, voidFuncPtr userFunc, int mode) {
+  // THIS IMPLEMENTATION IS NOT PORTABLE!
+  // 
+  // On AVR's calling convention, calling a function with more arguments than it declares 
+  // is OK - The extra parameter is ignored and causes no harm
+  //
+  // This implementation takes advantage of it to support callbacks with and without parameters with minimum overhead.
+  attachInterruptParam(interruptNum, (voidFuncPtrParam)userFunc, mode, NULL);
+}
+
+void attachInterruptParam(uint8_t interruptNum, voidFuncPtrParam userFunc, int mode, void* param) {
   if(interruptNum < EXTERNAL_NUM_INTERRUPTS) {
     intFunc[interruptNum] = userFunc;
+    intFuncParam[interruptNum] = param;
     
     // Configure the interrupt mode (trigger on low input, any change, rising
     // edge, or falling edge).  The mode constants were chosen to correspond
@@ -271,6 +283,7 @@ void detachInterrupt(uint8_t interruptNum) {
     }
       
     intFunc[interruptNum] = nothing;
+    intFuncParam[interruptNum] = NULL;
   }
 }
 
@@ -282,7 +295,7 @@ void attachInterruptTwi(void (*userFunc)(void) ) {
 
 #define IMPLEMENT_ISR(vect, interrupt) \
   ISR(vect) { \
-    intFunc[interrupt](); \
+    intFunc[interrupt](intFuncParam [interrupt]); \
   }
 
 #if defined(__AVR_ATmega32U4__)
--- a/hardware/arduino/avr/cores/arduino/wiring_private.h
+++ b/hardware/arduino/avr/cores/arduino/wiring_private.h
@@ -64,6 +64,7 @@ uint32_t countPulseASM(volatile uint8_t *port, uint8_t bit, uint8_t stateMask, u
 #endif
 
 typedef void (*voidFuncPtr)(void);
+typedef void (*voidFuncPtrParam)(void*);
 
 #ifdef __cplusplus
 } // extern "C"


paulo-raca:
A very simple example of how this can be used to writing components:

```
class PinListener {
  int pin;

  static void isr(void* arg) {
    PinListener* listener = (PinListener*)arg;
    listener->onInterrupt();
  }
  
public:

  PinListener(int pin) {
    this->pin = pin;
    pinMode(pin, INPUT_PULLUP);
    attachInterruptParam(digitalPinToInterrupt(pin), isr, CHANGE, this);   
  }

  void onInterrupt() {
    Serial.print("Pin ");
    Serial.print(pin);
    Serial.print(" changed to ");
    Serial.println(digitalRead(pin));
  }
};



PinListener listeners[] = {
  PinListener(0),
  PinListener(1),
  PinListener(2),
  PinListener(7)
};



void setup() {
  Serial.begin(9600);
}

void loop() {
}
```

NicoHood:
Doesnt this break the current API with no extra param? I am not sure if the code works in this case. Have you tested it?

Also this adds a lot of overhead, even though it could be very useable. Not sure about that.

However I like the compact definitions in `WInterrupts.c`.


If you ask me I'd rethink the whole API in general. I consumes quite some flash, even though only a singe PinInterrupt is used.

And I'd also change those:
```
 #define CHANGE 1
 #define FALLING 2
 #define RISING 3
```

to
```
 #define FALLING 1
 #define RISING 2
 #define CHANGE 3
````

So you have bit0 indicate falling, bit1 rising and both together a change.

**Comment by matthijskooijman** [comment]
The code looks good to me. The compatibility wrapper is a clever way to still support the existing parameterless version (@NicoHood, I think that still works), though it might be better for performance to not have the extra function call. One could instead store `void*` for the function pointers and cast them to the right type when calling, which I presume would be more efficient (perhaps smaller as well). The challenge then is how to decide whether to pass a parameter. Using NULL as a marker would work, but mean that you can never pass NULL, which isn't so great. Another option would be to use the LSB of the function pointer (since that is always 0 due to alignment), but that might be fragile as well. Perhaps storing a bitmask (which would only take 1 byte with the current boards) is ok, though.

The introduction of `IMPLEMENT_ISR` looks good to me, but ideally that would be in a separate commit, since it is totally unrelated to the main change in this PR.

**Comment by PaulStoffregen** [comment]
Has anyone tested the performance impact?

**Comment by paulo-raca** [comment]
Thanks for the feedback.

I'll split the `IMPLEMENT_ISR` macro into a separate PR.

@PaulStoffregen: How can I measure the performance impact?


@matthijskooijman, what do you think of using a poison value on voidFuncPtrParam (Maybe `nothing`) to mean "parameterless funcion". 

```
#define IMPLEMENT_ISR(vect, interrupt) \
	ISR(vect) { \
		if (intFuncParam [interrupt] == nothing) {
			((voidFuncPtr)intFunc[interrupt])();
		} else {
			((voidFuncPtrParam)intFunc[interrupt])(intFuncParam[interrupt]);
		}
	}
```

The tradeoff is that ISR functions will take a little more flash to handle the condition inline


@NicoHood, Although I agree the current implementation wastes some flash with unused interrupts, I really don't know how to rewrite it in a more efficient way. I'm taking suggestions :D

`compatFunc` maintains compatibility with current parameterless API. I've tested it and it works.

About the constant values, I agree with you, that makes more sense, but looks like the values are defined by the hardware

**Comment by matthijskooijman** [comment]
> @matthijskooijman, what do you think of using a poison value on voidFuncPtrParam (Maybe nothing) to mean "parameterless funcion". 

That sounds reasonable, checking a pointer is probably a lot faster than calling a function. There is still overhead, of course, from loading the parameter and checking it.

As for measuring the overhead - disassembling and counting cycles would work, or doing some kind of runtime timekeeping (though that's more tricky to get accurate).

**Comment by matthijskooijman** [comment]
> @matthijskooijman, what do you think of using a poison value on voidFuncPtrParam (Maybe nothing) to mean "parameterless funcion". 

Wait, that's what I proposed with letting "NULL" to mean "no parameter". However, this will remove one value from the available parameter values, which I think is not a good plan (there is a fair chance that a user will pass that value anyway, especially if the void* argument is used to pass integers instead of pointers, and because function and data pointers can overlap).

What I thought you were saying, was to use a poisoned as the function pointer value, which means that the parameter contains the actual (parameterless) function pointer to call. This is more reliable because valid values for the function pointer array are only pointers into the flash, and never pointers into RAM or plain integers.

**Comment by paulo-raca** [comment]
You are right.

Something like this?
```
        if (intFunc [interrupt] == compatFunc) {
                ((voidFuncPtr)intFuncParam[interrupt])();
        } else {
                intFunc[interrupt])(intFuncParam[interrupt]);
        }
```

**Comment by matthijskooijman** [comment]
Yup, that's what I had in mind indeed. You should probably get `intFunc[interrupt]` once and put it in a variable, though, since `intFunc` is volatile so the compiler likely won't optimize it otherwise. As for the `compatFunc` (which probably needs some better name), ideally it would be a function containing of just a "ret" instruction, so it only takes up 2 bytes. There might be some gcc attributes to achieve that (to prevent any pre- and postamble from being generated, possibly by making it a naked ISR).

**Comment by paulo-raca** [comment]
I've made a separate PR for the `IMPLEMENT_ISR` macro on #4521

**Comment by paulo-raca** [comment]
I haven't compared the performance yet, but the memory use on a pro micro (atmega32u4) is
- using `compatFunc`: +66 bytes on flash, +10 bytes of ram
- using @matthijskooijman's suggestion: +132 bytes on flash, +10 bytes of ram

**Comment by paulo-raca** [comment]
Current code:
==========

```
#define IMPLEMENT_ISR(vect, interrupt) \
  ISR(vect) { \
    intFunc[interrupt](); \
  }
```

ISR: Executes 5 instructions from code + 40 instructions of boilerplate

```
..... Here goes prologue boilerplace: 21 instructions (Mostly `push`)
743 0330 8091 0000 		lds r24,intFunc+2
744 0334 9091 0000 		lds r25,intFunc+2+1
745 0338 E82F      		mov r30,r24
746 033a F92F      		mov r31,r25
747 033c 0995      		icall
..... Here goes epilogue boilerplace: 19 instructions (Mostly `pop`)
```



My code, using `compatFunc`:
======================

```
static void compatFunc(void* realCallback) {
  voidFuncPtr f = realCallback;
  f();
}

#define IMPLEMENT_ISR(vect, interrupt) \
  ISR(vect) { \
    intFunc[interrupt](intFuncParam [interrupt]); \
  }
```

ISR: Executes 7 instructions from code + 40 instructions of boilerplate

```
..... Here goes prologue boilerplace: 21 instructions (Mostly `push`)
772 0376 2091 0000 		lds r18,intFunc
773 037a 3091 0000 		lds r19,intFunc+1
774 037e 8091 0000 		lds r24,intFuncParam
775 0382 9091 0000 		lds r25,intFuncParam+1
776 0386 E22F      		mov r30,r18
777 0388 F32F      		mov r31,r19
778 038a 0995      		icall
..... Here goes epilogue boilerplace: 19 instructions (Mostly `pop`)
```

compatFunc: Executes 24 instructions

```
    119 000e CF93      		push r28
    123 0010 DF93      		push r29
    127 0012 00D0      		rcall .
    128 0014 00D0      		rcall .
    131 0016 CDB7      		in r28,__SP_L__
    132 0018 DEB7      		in r29,__SP_H__
    139 001a 9C83      		std Y+4,r25
    140 001c 8B83      		std Y+3,r24
    142 001e 8B81      		ldd r24,Y+3
    143 0020 9C81      		ldd r25,Y+4
    144 0022 9A83      		std Y+2,r25
    145 0024 8983      		std Y+1,r24
    147 0026 8981      		ldd r24,Y+1
    148 0028 9A81      		ldd r25,Y+2
    149 002a E82F      		mov r30,r24
    150 002c F92F      		mov r31,r25
    151 002e 0995      		icall
    154 0030 0F90      		pop __tmp_reg__
    155 0032 0F90      		pop __tmp_reg__
    156 0034 0F90      		pop __tmp_reg__
    157 0036 0F90      		pop __tmp_reg__
    158 0038 DF91      		pop r29
    159 003a CF91      		pop r28
    160 003c 0895      		ret
```

Summary:
- Flash: +66 bytes
- RAM: +10 bytes
- ISR execution: +26 instructions


Suggested code, using `if` inside the ISR:
==========================================

```
#define IMPLEMENT_ISR(vect, interrupt) \
  ISR(vect) { \
    voidFuncPtrParam cb = intFunc[interrupt]; \
    void* param = intFuncParam [interrupt]; \
    if (cb == NOARG_CALLBACK) { \
      ((voidFuncPtr)param)(); \
    } else { \
      cb(param); \
    } \
  }
```

ISR: Executes 20-21 out of 27 instructions from code + 40 instructions of boilerplate

```
..... Here goes prologue boilerplace: 21 instructions (Mostly `push`)
726 034a 8091 0000 		lds r24,intFunc
727 034e 9091 0000 		lds r25,intFunc+1
728 0352 9A83      		std Y+2,r25
729 0354 8983      		std Y+1,r24
730 0356 8091 0000 		lds r24,intFuncParam
731 035a 9091 0000 		lds r25,intFuncParam+1
732 035e 9C83      		std Y+4,r25
733 0360 8B83      		std Y+3,r24
734 0362 8981      		ldd r24,Y+1
735 0364 9A81      		ldd r25,Y+2
736 0366 8F3F      		cpi r24,-1
737 0368 2FEF      		ldi r18,-1
738 036a 9207      		cpc r25,r18
739 036c 01F4      		brne .L21
741 036e 8B81      		ldd r24,Y+3
742 0370 9C81      		ldd r25,Y+4
743 0372 E82F      		mov r30,r24
744 0374 F92F      		mov r31,r25
745 0376 0995      		icall
746 0378 00C0      		rjmp .L20
747               	.L21:
749 037a 8B81      		ldd r24,Y+3
750 037c 9C81      		ldd r25,Y+4
751 037e 2981      		ldd r18,Y+1
752 0380 3A81      		ldd r19,Y+2
753 0382 E22F      		mov r30,r18
754 0384 F32F      		mov r31,r19
755 0386 0995      		icall
756               	.L20:
..... Here goes epilogue boilerplace: 19 instructions (Mostly `pop`)
```


Summary:
- Flash: +132 bytes
- RAM: +10 bytes
- ISR execution: +15 instructions

**Comment by paulo-raca** [comment]
Comparing @matthijskooijman's suggestion with my code:
- 11 instructions faster without params
- 14 instructions slower with params
- Uses 20 extra bytes in flash for each ISR

To put things in perspective, the ISR boilerplate alone takes 40 instructions, so +14/-11 doesn't seem like a big deal

**Comment by matthijskooijman** [comment]
@paulo-raca, thanks for the thorough comparison :-D

I'm actually quite surprised about the mess the compiler makes when compiling my suggestion. There's a lot of overhead (12 instructions from my count) from storing and reloading variables on the stack, when they could just as well be kept around in registers (AFAICS all of the call-clobbered registers are pushed by the boilerplate, so they should be available). I wonder if it would be worth implementing this in hand-coded assembly perhaps...

Also interesting is that compatfunction has something similar going on, which is even more silly since it stores r24 and r25 on the stack and reloads them immediately to the same registers *twice*. It's almost as if you compiled without optimizations enabled?

**Comment by paulo-raca** [comment]
You are right, I completely forgot to enable optimizations :/

I'll update it later

(I hope most of the ISR boilerplate goes away with it)

**Comment by matthijskooijman** [comment]
@paulo-raca, that suggests you compiled with avr-gcc manually? For a fair comparison, it would seem best to actually use the Arduino IDE for compilation, so you get the "official" compiler flags and compilation process. To extract the .elf file from the IDE for disassembly, just look through the verbose compilation output to find where it is stored.

**Comment by paulo-raca** [comment]
@matthijskooijman, I updated the previous comment with optimized disassembly.

Turns out that using `compatFunction` is extremely efficient and gives only +4 instructions overhead.

**Comment by matthijskooijman** [comment]
I looked  up the number of cycles in addition to the number of instructions, and I took the liberty of adding them in your comment. I believe the last code snippet could be optimized (by hand-coding in assembly, or perhaps by making the compiler smarter) to remove one `icall` and `rjmp` instruction, saving four bytes of flash and two cycles, but it' still slower than the `compatFunc` approach.

To me the `compatFunc` approach seems decent, the compiler seems smart enough to make it really small without any boilerplate. I tend to think about calls, and especially indirect ones, as expensive, but that seems to mostly true on x86 (which does heavy pipelining and speculative execution). On the AVR, an indirect calll is just three cycles (even a cycle faster than a direct call, interestingly enough).

So, your original approach seems the most efficient. Any users that do not need this new parameter feature will pay an additional 7 cycles. The current latency is 17*2+7 (assuming all 17 preamble boilerplate instructions are push at 2 cycles per instruction) 41 cycles. This new feature brings a measureable, but not enormous, cost.


On additional option we could consider is to simply pass the parameter, whether it is valid or not. Knowing AVR's calling conventions, we know that passing a parameter will simply occupy two registers. If the called function does not expect these parameters, it will just ignore them, without any additional problems. I don't see any way to actually *verify* this during compilation (e.g. by writing things in such a way that the compiler will see this optimization opportunity and exploit it, and/or an error be thrown if the compiler at some point handles calling things differently), but that does not need to be a problem necessarily (a decent comment should be written, though, to ensure this does not break when ported to other platforms).

If you would just pass the parameter, the additional overhead is two `lds` instructions, or 4 cycles for both cases, which seems reasonable to me.

**Comment by paulo-raca** [comment]
I like the idea!

**Comment by paulo-raca** [comment]
I've implemented it, and everything is looking good.

Thanks you for help, @matthijskooijman

**Comment by paulo-raca** [comment]
Is there anything else I can do about this?
Can we have it merged?

As soon as it gets settled, I plan to do the same on Due

**Comment by facchinm** [comment]
Hi @paulo-raca ,
I merged #4521, would you mind rebasing this one against current `master` ? :smile:

**Comment by paulo-raca** [comment]
Rebased. Thanks!

**Comment by paulo-raca** [comment]
It's been a while... Should we merge or discard?

**Comment by igrr** [comment]
I was about to implement something similar in the esp8266 core. Would be awesome if this pull request was merged so that we don't introduce any more non-portable stuff than we have to.

**Comment by Chris--A** [comment]
Yes, this is a great idea. +1

**Comment by Makuna** [comment]
While trying to implement a Infrared library that supports more than one instance (4 directional IR where the directions are important) I ran into this issue.

Please, merge this in as the hack to make a reusable class attaching to ISR is painful.

**Comment by paulo-raca** [comment]
@Makuna - I'm not familiar with C++11 lambdas, but it looks nice :)
Do you have any idea the memory/cpu impact needed to support lambdas?

**Comment by Makuna** [comment]
@paulo-raca in general lambdas generated code is equivalent too code you would write but the compiler can do better optimizations in some cases.  The only difference can be if your hand coded solution pre-allocates the memory for captured variables and reuses it in the case where the lambda is temporary (not likely with interrupts unless you are attaching and detaching a lot); while the compiler will allocate and free it possibly causing memory fragmentation (but if you call new and delete a lot already you will cause the same issue).

**Comment by paulo-raca** [comment]
@Makuna: I briefly tried to use `std::funciona` today, but:
- the AVR compiler was unable to recognize it
- The Due compiler it crashed with some linking errors. 

It did work for the ESP8266 compiler, and I measured the overhead (give or take):
- The minimal code and adds ~250 bytes of compiled code (flash) (`std::function<void()> x = [](){}; x();`)
- A call adds 49 bytes of flash
- Each instance takes 16 bytes of RAM (Not too bad, but a function pointer + user pointer would only take use 8 bytes on the ESP)

Anyway, since the compiler support is not there yet, I'll leave it as-is for now, and we can bring it up again later.

**Comment by Chris--A** [comment]
`std::funciton` while useful has hidden issues especially for AVR boards (which is likely why the C++ standard library is missing). Internally it uses dnamic memory.

**Comment by dplasa** [comment]
Will this be merged eventually in arduino-core upstream? It is quite handy code :)

**Comment by paulo-raca** [comment]
This has been open for a while... The overhead is pretty small and many people seem to like the idea.
@matthijskooijman @NicoHood, can we get it merged?
Otherwise, lets close this PR.

**Comment by tni** [comment]
On [ARM Cortex M4](https://gcc.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,source:'struct+callback_t+%7B%0A++++void+(*ptr)(void*)%3B%0A++++void*+context%3B%0A%7D%3B%0A%0Astatic+callback_t+fn_table1%5B64%5D%3B%0Astatic+void+(*fn_table2%5B64%5D)(void)%3B%0A%0Avoid+attach1(int+i,+void+(*ptr)(void*),+void*+context)+%7B%0A++++fn_table1%5Bi%5D+%3D+%7B+ptr,+context+%7D%3B%0A%7D%0A%0Avoid+attach2(int+i,+void+(*ptr)(void))+%7B%0A++++fn_table2%5Bi%5D+%3D+ptr%3B%0A%7D%0A%0Avoid+invokeCallback1(int+i)+%7B%0A++++fn_table1%5Bi%5D.ptr(fn_table1%5Bi%5D.context)%3B%0A%7D%0A%0Avoid+invokeCallback2(int+i)+%7B%0A++++fn_table2%5Bi%5D()%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:arm541,filters:(b:'0',commentOnly:'0',directives:'0',intel:'0'),options:'-O2+-std%3Dc%2B%2B11+-mthumb+-mcpu%3Dcortex-m4',source:1),l:'5',n:'0',o:'ARM+gcc+5.4.1+(Editor+%231,+Compiler+%232)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), invoking that callback is 5 vs. 4 instructions.

The void* parameter would be passed in a register, so a function(void) could be called as function(void*).

\\\\

@Makuna:

The void* parameter works fine with lamdas, e.g.:



```
class S {
public:
    void callback();
};

void registerCallback(void (*callback_fn)(void*), void* context) {
    // ...
}

void test(S* s) {
    // ...
    registerCallback([](void* ctx){ ((S*) ctx)->callback();  }, s);
} 
```

**Comment by paulo-raca** [comment]
I planned to follow up this patch with an ARM versions, but seems like it's not going anywhere :(

Anyway, I find that 5 instructions is still a pretty low overhead for what (I think) is a pretty useful feature.

Not sure about the lamdas, but I think it probably won't support lamdas with a capture list -- Just guessing, can you try it?

**Comment by tni** [comment]
No, you can't use lambda captures. They need storage that a function pointer doesn't have (you would need to use std::function). Lambdas without captures decay to function pointers.

The Cortex M4 overhead is 1 extra instruction. :-)

**Comment by aaronw2** [comment]
This would be extremely useful for me as well. It works well with C++ since I can just pass the this pointer as a parameter.

I didn't know about this thread. In my implementation for the ESP32 I also pass the interrupt (pin) number to the handler, though with a bit more overhead I could live without that. It just means that I need to extract the pin number from my class instead.

**Comment by paulo-raca** [comment]
@matthijskooijman, the current approach adds 2 instructions / 4 cycles on AVR:

Before:
```
     d78:       e0 91 14 01     lds     r30, 0x0114     ; 0x800114 <intFunc+0x8>
     d7c:       f0 91 15 01     lds     r31, 0x0115     ; 0x800115 <intFunc+0x9>
     d80:       09 95           icall
```

After:
```
     d78:       e0 91 14 01     lds     r30, 0x0114     ; 0x800114 <intFunc+0x8>
     d7c:       f0 91 15 01     lds     r31, 0x0115     ; 0x800115 <intFunc+0x9>
     d80:       80 91 4a 01     lds     r24, 0x014A     ; 0x80014a <intFuncParam+0x8>
     d84:       90 91 4b 01     lds     r25, 0x014B     ; 0x80014b <intFuncParam+0x9>
     d88:       09 95           icall
```

**Comment by paulo-raca** [comment]
@aaronw2, thanks for bringing it up again :)

When it gets merged I'll also make patches for ESP8266, ESP32 and Due

**Comment by cmaglie** [comment]
@paulo-raca 
Is the "non-portable" trick applicable also to ARM?
Do you have a reference implementation to compare the performance also there?

**Comment by paulo-raca** [comment]
I haven't tried, but I believe it does.

According to @tni, the overhead on arm is only 1 instruction.

**Comment by matthijskooijman** [comment]
Also, even if ARM needs a bit more instructions, that's probably less of a problem given the usually faster MCUs. And since this is just an implementation detail, that does not affect the API, this should not hinder the changes on AVR.

**Comment by cmaglie** [comment]
> I haven't tried, but I believe it does.

Ok, may you explain why calling a function with extra parameters is safe in the AVR case?
I'd like to have some clarification I'm not so easy to merge something that really looks like something you should not do :-)

> Also, even if ARM needs a bit more instructions, that's probably less of a problem given the usually faster MCUs.

The point is how much is this "bit more" on the ARM side, so I'm going to give it a try to check it out.

**Comment by paulo-raca** [comment]
> Ok, may you explain why calling a function with extra parameters is safe in the AVR case?

On AVR, the extra arguments are sent on registers `r24` and `r25`. If the function doesn't expect these arguments, these values on these registers are simply be ignored.

The big red warning mostly applies to ABIs where arguments are stored on the stack.
E.g., If the ISR pushes 2 arguments on the stack but the callback only pops one when returning, everything breaks.

**Comment by paulo-raca** [comment]
I've just written an equivalent PR for Arduino Due.

The same trick also worked on ARM, and the ISR overhead is only 2 instructions.

**Comment by dplasa** [comment]
> The big red warning mostly applies to ABIs where arguments are stored on the stack.
> E.g., If the ISR pushes 2 arguments on the stack but the callback only pops one when returning, 
> everything breaks.

Is this really true? I believe the caller has to clean up the stack or else one could not realise functions with variable count of parameters like 
int printf(char* format, ...);
where the calles function cannot know the number of passed arguments.

At least that's how it's done in C on x86.

**Comment by paulo-raca** [comment]
@dplasa, this is true for variadic functions, but the compiler might use a different calling convention on them.

E.g., on AVR:
```
void foo(char a, char b) {}
void bar(char a, ...) {}
void callFoo() {
  foo(1,2);
}
void callBar() {
  bar(1,2);
}
```
compiles to
```
[...]
void callFoo() {
[...]
  foo(1,2);
  2e:   62 e0           ldi     r22, 0x02       ; 2
  30:   81 e0           ldi     r24, 0x01       ; 1
  32:   e6 df           rcall   .-52            ; 0x0 <foo>
}
[...]
void callBar() {
[...]
  bar(1,2);
  42:   1f 92           push    r1
  44:   82 e0           ldi     r24, 0x02       ; 2
  46:   8f 93           push    r24
  48:   81 e0           ldi     r24, 0x01       ; 1
  4a:   8f 93           push    r24
  4c:   e5 df           rcall   .-54            ; 0x18 <bar>
  4e:   0f 90           pop     r0
  50:   0f 90           pop     r0
  52:   0f 90           pop     r0
}
[...]
```

On x86 there is a number of calling conventions where the callee should clean up the task: https://en.wikipedia.org/wiki/X86_calling_conventions#Callee_clean-up

**Comment by dplasa** [comment]
Still my argument holds :)
- If the caller passes arguments on the stack - it cleans them up after the call (void callBar()...)
- If the caller passes arguments in registers (void callFoo()) nothing needs to be cleaned

If an void ISR() without argument was called the compiler would probably just rcall the ISR code when the interrupt occurs. 

If we change this to ISR(void* arg) then arg is either passed by register which is no deal to an ISR(void) function which just ignores the register...
If the compiler would generate code such that arg is passed over the stack, the caller will also clean up the stack -  and again the ISR(void) just ignores the bit on the stack.

**Comment by paulo-raca** [comment]
Hello,
Since the AVR core has been moved into a separate sepository, I created a new PR with the same changes: https://github.com/arduino/ArduinoCore-avr/pull/58

There is also a version for SAM: https://github.com/arduino/ArduinoCore-avr/pull/58

If these are merged in the official API, I'll also port this feature for ESP8266 / ESP32

**Comment by aaronw2** [comment]
I'm not aware of any Arduino compatible core that would have a problem with this. In every ABI I see supported, the first argument is always passed in a register.

**Comment by paulo-raca** [comment]
I'm not aware either, but it's better to make it very clear, since it is an abuse of the ABI

**Comment by paulo-raca** [comment]
Ok, so what if I replace the comment with this?
```
// To support callbacks with and without parameters with minimum overhead,
// this relies on that fact that in C calling conventions extra argument on a
// function call are safely ignored without side-effects.
```

aaronw2:
That sounds good to me.

PR #4521, opened 2016-02-04
Boilerplate reduction on ISR declaration

paulo-raca:
The current code is very verbose and a painful to maintain

(Change ISR implementation in 20 different places? No Thanks!).

(This was originally part of #4519, but we all agreeded it deserved it's own PR)


matthijskooijman:
This commit looks excellent to me, I'm in favor of merging it.

matthijskooijman:
(On good way to test its correctness would be to compiler a program using attachInterrupt and verify that the resulting .hex is identical)

paulo-raca:
I've checked that the .hex files are identical :)

PR #4547, opened 2016-02-09
Added to #define for 1284/1284p Timer 3 inclusion.

feilipu:
The ATmega1284/ATmega1284p has a 16 bit Timer 3 (not present in the ATmega644 and below).
Add to the #define to use Timer 3 or Timer 1 optionally.


facchinm:
Looks good to me :+1:

PR #4566, opened 2016-02-12
Remove dependency on Sd2PinMap.h

feilipu:
Sd2PinMap.h contains a legacy implementation of the information provided in pins_arduino.h, and is only used by Sd2Card.h.

This commit removes the #include "Sd2PinMap.h", and redefines the SD SPI interface to pin names provided by pins_arduino.h.

The file Sd2PinMap.h should also be removed as it is no longer relevant.


feilipu:
Will revise to limit effects to hardware SPI interface only.

PR #4576, opened 2016-02-15
Update Sd2Card.h to use pins_arduino.h definitions

feilipu:
Update Sd2Card.h to use pins_arduino.h or variant.h definitions, depending on architecture, when Hardware SPI is in use.

The definitions for Software SPI pins (for ATmega1280 and ATmega2560 etc) remain unchanged as defined in Sd2PinMap.h.


feilipu:
Bump. Comments?

sandeepmistry:
@feilipu thanks, looks good!

I've tested with:
 * Uno
 * Mega: hardware and software SPI (sketch size was 11,536 bytes before and after for software SPI)
 * Due
 * Zero

PR #4588, opened 2016-02-17
Add Arduino/Genuino 101 support to TFT library

bigdinotech:


q2dg:
I think this library is deprecated.
Sorry if I am wrong.

Issue #4525, opened 2016-02-04
Documentation should use more specific types


matthijskooijman:
Currently, the documention uses types like "unsigned long" (for example here: https://www.arduino.cc/en/Reference/millis). However, this is only correct for the AVR architecture, on e.g. the Due a long is (AFAIK) 64 bit, but the code explicity uses `uint32_t`. It would make sense to adopt the same explicit approach in the documentation (and, while we're at it, probably apply this to the code of the AVR port as well).

agdl:
@matthijskooijman +1 I think that actually all the documentation and software should use standard data types like uint8_t, int32_t  etc. In this way it is immediately clear the "dimension" of a variable and code can be optimized and standardized

agdl:
BTW once upon a time someone from the high levels told me that for beginners standard data types are not too friendly...

matthijskooijman:
> BTW once upon a time someone from the high levels told me that for beginners standard data types are not too friendly...

They might be a bit scary-looking, that's true, but the using platform-dependent types of unknown size looks nicer but doesn't work as well. We could consider introducing more friendly alternative names, but then Arduino code becomes even less portable, and things people learn will also be less reusable.

Perhaps a decent alternative is to use the standard names in the docs, but make them a link to a page explaining what these types mean? Anyone that is confused about a typename can click it and get an explanation?

agdl:
I think we need @tigoe

tigoe:
I would rather we adopt the Arduino types to new platforms as we go. As you said, Arturo, the point in making them in the first place was to make the language more readable for a non-technical audience. I'd like to maintain that approach as we move to new platforms.

matthijskooijman:
@tigoe, what "Arduino types" do you mean exactly? There is the custom "byte" type, but AFAIK no other custom integer types are defined. The AVR core and documentation use int and long for the bigger integers, but those can't be portably used on other platforms, since their sizes will be different.

Or is your suggestion to introduce new types (e.g. word and dword in addition to byte or something)?

tigoe:
I mean the ones listed here, as used in Arduino sketches over the last ten years: 
https://www.arduino.cc/en/Reference/HomePage

I'm suggesting that if they don't exist for other platforms, we introduce them, to make those platforms compatible with Arduino as it currently exists.

matthijskooijman:
[...]

PR #4625, opened 2016-02-28
save some bytes by replacing int with uint8_t in Prints, digitalRead(…

tico-tico:
…) and analogWrite()


sandeepmistry:
@tico-tico I've cherry-picked https://github.com/tico-tico/Arduino/commit/a7454b6b5c59187b95c4224aad87bb01faa06e85 and will also port to the SAM and SAMD cores.

The changes in https://github.com/tico-tico/Arduino/commit/275a47da89df98468f60545ba5864f621511fd53 break the Arduino API signatures, this is not a good trade off for a reduction in stack space. For this reason I'm closing this.

NicoHood:
what API signature?
I think the patch is great and should be merged fully. There is no good reason to not do it like that.

tico-tico:
JFYI it's not stack size, it's code size

matthijskooijman:
@sandeepmistry, the changes that break API only break binary compatibility, AFAICS source compatibility should still be fine (I didn't look really close, though). That is, all source programs that were valid and meaningful before should still work in exactly the same way. Not sure if it's good to change the API though, also related to #4525 perhaps.

sandeepmistry:
@tico-tico 

> JFYI it's not stack space, it's code size

I missed this (was thinking about other instruction sets), I do agree it saves code size as well.

@NicoHood @matthijskooijman 

The change will fail or cause incompatibilities with the following (although unlikely in the real world):

 * code that is using a function pointer to ```digitalRead``` or ```analogWrite```
 * classes overriding ```Print::print``` methods changed

In general as discussed in #4525, in particular https://github.com/arduino/Arduino/issues/4525#issuecomment-186172224 almost all Arduino API's using the ["Data Types" defined in the Language Reference page](https://www.arduino.cc/en/Reference/HomePage).

Also from the [Arduino style guide](https://www.arduino.cc/en/Reference/StyleGuide):

> Use the wiring/Processing-style variable types, e.g. boolean,char,byte,int,unsigned int,long,unsigned long,float,double,string,array,void when possible, rather than uint8_t, etc. The former are explained in the documentation, and less terse names.

NicoHood:
We changed the whole usb core, so someone can change the digitalWrite pointers (who uses this? i know not a single useful example. If you got this far, you rather use direct port manipulation...)

The only virtual method in print is write. And write stays uint8_t so I see no problem.
Also it would make even more sense to use enums there and to variadic templates to make print a lot more flexible (i've opened a PR for this.)
https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Print.h

With time comes progress. And we should do it, since it breaks almost nothing.

matthijskooijman:
@sandeepmistry I see your point, but I agree with @NicoHood that the breakage should be minimal. Perhaps it would be good to keep track of this change for later (such as when releasing 1.7/1.8), so we can combine a couple of not-fully-compatible changes in a single release and document them properly?

PaulStoffregen:
On 32 bit chips, int8_t is actually slower than int or int32_t in many cases, because it forces the compiler to insert extra logical AND or sometimes sign extension instructions.

tico-tico:
> On 32 bit chips, int8_t is actually slower than int or int32_t in many case

Aren't they use different "cores" for different targets? At least I always thought so. But if you are right about 32 bits future, then i think **u**/**int_fast8_t** should be used, not some shady **int**s.

Speaking of **Print**s, using **int**s in the "API" (does this thing exist?), dragging them through all the functions and then casting them to **uint8_t** at the the very end doesn't do good job for tiny AVRs where every byte matters. Anyway, it's up to you guys. I think people who faced code size issues already made all possible changes to the "core" to shrink the code.

PR #4626, opened 2016-02-29
Use "$@" in shell script

cousteaulecommandant:
Use "$@" in this shell script instead of "$1" "$2" "$3" "$4" "$5" "$6" "$7" "$8" "$9".
(Is this file still used?)

### DIFF
--- a/build/cmd/dist/processing
+++ b/build/cmd/dist/processing
@@ -1,4 +1,4 @@
- #!/bin/sh
+#!/bin/sh
  
 APPDIR="$(dirname -- "${0}")"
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Hm, this PR fixes build/cmd/dist/processing. The fix itself looks good to me, but I wonder if that script is actually used at all?

q2dg:
Well...if it's not, let's remove it.

matthijskooijman:
Indeed seems it is no longer used. I created #5625 to remove it, and I'm closing this PR.

PR #4627, opened 2016-02-29
Update from USB 2.0 to USB 2.1

beaufortfrancois:
Let's update USB Version from 2.0 to 2.1 so that the host could tell if the USB device exposes a [Binary Object Store descriptor](http://blogs.msdn.com/b/usbcoreblog/archive/2013/04/11/usb-2-1-2-0-1-1-device-enumeration-changes-in-windows-8.aspx) or not. 
For info, this is where the [WebUSB](https://github.com/WICG/webusb) descriptor (and Microsoft OS Descriptor 2.0) reading sequence starts.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.h
+++ b/hardware/arduino/avr/cores/arduino/USBCore.h
@@ -130,7 +130,7 @@
 typedef struct {
 	u8 len;				// 18
 	u8 dtype;			// 1 USB_DEVICE_DESCRIPTOR_TYPE
-	u16 usbVersion;		// 0x200
+	u16 usbVersion;		// 0x210
 	u8	deviceClass;
 	u8	deviceSubClass;
 	u8	deviceProtocol;
@@ -260,7 +260,7 @@ typedef struct
 
 
 #define D_DEVICE(_class,_subClass,_proto,_packetSize0,_vid,_pid,_version,_im,_ip,_is,_configs) \
-	{ 18, 1, 0x200, _class,_subClass,_proto,_packetSize0,_vid,_pid,_version,_im,_ip,_is,_configs }
+	{ 18, 1, 0x210, _class,_subClass,_proto,_packetSize0,_vid,_pid,_version,_im,_ip,_is,_configs }
 
 #define D_CONFIG(_totalLength,_interfaces) \
 	{ 9, 2, _totalLength,_interfaces, 1, 0, USB_CONFIG_BUS_POWERED | USB_CONFIG_REMOTE_WAKEUP, USB_CONFIG_POWER_MA(500) }


NicoHood:
But where is the BOS coded in the usb core? As far as I understood 2.1 supports this BOS thing and enables micro$$$$ to read the BOS. But what if its not there? Is it? Never heard of that. Why would you need it?

reillyeon:
This patch doesn't add any code to the USB core to respond to a BOS request but leaves it up to libraries using PluggableUSB to provide it. Without the version change the host OS will not request the BOS descriptor. I don't believe it will be an issue for devices to advertise this version and still return nothing in response to a GET_DESCRIPTOR(BOS) request.

NicoHood:
That makes sense now. Do you have a working usecase of a pluggable usb module?

beaufortfrancois:
We do: the [WebUSB Arduino library](https://github.com/webusb/arduino/tree/25e33acb0345fffa610c07aa20c898e2a3861859/library/WebUSB) part of the [WebUSB API](https://wicg.github.io/webusb/)

beaufortfrancois:
(gentle ping)

NicoHood:
https://github.com/arduino/Arduino/pull/4793

beaufortfrancois:
Yeah!

beaufortfrancois:
@reillyeon  have a look at https://github.com/arduino/Arduino/pull/4793

facchinm:
Hi everyone,
what about setting using a define instead than an hardcoded value?
```
#ifndef USB_VERSION
#define USB_VERSION 0x200
#endif
```
and 
```
#define D_DEVICE(_class,_subClass,_proto,_packetSize0,_vid,_pid,_version,_im,_ip,_is,_configs) \
{ 18, 1, USB_VERSION, _class,_subClass,_proto,_packetSize0,_vid,_pid,_version,_im,_ip,_is,_configs }
```

In this way you could add a bunch of variant boards which should be exactly the same as Micro/Leonardo but defining `-DUSB_VERSION=0x210` in boards.txt, leaving them out of the core.

I'm preparing a PR right now, I'll ask @reillyeon to test it and eventually update the WubUSB arduino integration howto.

matthijskooijman:
@facchinm (and probably even more @beaufortfrancois), are there any downsides to just setting the USB version to 2.1? Are there additional things we need to support or do?

facchinm:
@matthijskooijman ,
windows 8.1 and 10 require a valid BOS descriptor to correctly enumerate if USB2.1 is declared. This brings a huge unwanted overhead on all other sketches.

NicoHood:
We could make the whole usb core pluggable/optional. This way you could also make the Leonardo sketched without usb smaller and faster. I've opened a PR about this some time ago. The people who need USB 2.1 could fork the usb core and reimplement the stuff they need. But they could still use the rest of the core.

If you ask me I'd make the whole core more like a set of libraries. For USB, Seriel, Millis, Interrupts etc. Then you have less code and more flexible in overwriting those.

matthijskooijman:
Alternatively, perhaps the USB version could be determined at runtime? Either by some code (the same code that adds the BOS descriptor) setting it to 2.1, or perhaps by somehow detecting that a BOS descriptor was plugged in and doing it automatically (not sure if that is feasible, though).

facchinm:
Superseded by #5128, now the board which needs to advertise as USB 2.1 can do it with a simple extra boards.txt (see https://github.com/webusb/arduino/pull/16)

facchinm:
Thanks everyone for the discussion and contribution !

PR #4658, opened 2016-03-05
Show board name on generic "Error compiling" message

PaulStoffregen:
As discussed on the developers mail list, this change adds the board name to the generic "Error compiling" message.

https://groups.google.com/a/arduino.cc/d/msg/developers/rs72zE8DuN0/B7elCSB-EQAJ

This generic error only appears when the IDE is unable to highlight a specific line in the editor.

### DIFF
--- a/arduino-core/src/cc/arduino/Compiler.java
+++ b/arduino-core/src/cc/arduino/Compiler.java
@@ -269,7 +269,7 @@ public class Compiler implements MessageConsumer {
     }
 
     if (result != 0) {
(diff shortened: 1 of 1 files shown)


cmaglie:
Thanks!
I pushed the change in a slightly different way to allow easier translation: https://github.com/arduino/Arduino/commit/8d995d73f318979f5b47fa33a186b6da30d60cb9

lmihalkovic:
Counting the amount of time spent by all those who contributed their 2¢, that must be one of the most expensive one-liner I have seen in a while.
Cheers

PR #4667, opened 2016-03-08
Speed and size improvement in Print::printFloat()

bblanchon:
Avoid using the overload of print() for signed integer since a negative value is not allowed here.
This results in a smaller (unless print(int) is used somewhere else in the program) and faster code because the overload for unsigned integer is simpler.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -257,7 +257,7 @@ size_t Print::printFloat(double number, uint8_t digits)
   while (digits-- > 0)
   {
     remainder *= 10.0;
-    int toPrint = int(remainder);
+    unsigned toPrint = unsigned(remainder);
     n += print(toPrint);
     remainder -= toPrint; 
   } 


bblanchon:
Please merge this PR :pray: 
It just change two words in one file.
It'll take 2 seconds to review, literally.

proppy:
friendly @cmaglie ping

sandeepmistry:
I just tested this with the following test sketch on an Uno:

```arduino
void setup() {
  Serial.begin(9600);
  while(!Serial);

  uint32_t start = millis();
  for (int i = 0; i < 100; i++) {
    Serial.println(PI);
  }
  uint32_t end = millis();

  Serial.println(end - start);
}
void loop() {
}
```

Here are some metrics:
 * Without patch:
   * Sketch size 3,824 bytes
   * Execution time 556 ms
 *  With patch:
   * Sketch size 3,644 bytes
   * Execution time 556 ms

sandeepmistry:
@bblanchon thanks!

I've also ported this change to the SAM (https://github.com/arduino/Arduino/commit/05be7ce395f4fabc4482ab1c964075024b511ec9) and SAMD (https://github.com/arduino/ArduinoCore-samd/commit/94d6f56cee48a63ffa3963011f3e75ca5b9cb5aa) cores.

More notes on size reduction for test sketch above:
 * Due
  * Before: 25,212 bytes
  * After: 25,068 bytes
 * MKR1000:
  * Before: 16,256 bytes
  * After: 16,088 bytes

Execution times were negligible.

PR #4669, opened 2016-03-09
Add missing cast to SoftwareSerial::peek() when returning buffered value

### LINKED ISSUE #2858: SoftwareSerial returns impossible values
I have the following code (Arduino 1.6.1):

    if(_bt.available() > 0) {
    	int data = _bt.peek();
    	bool found = true;

    	if(data == 'a') {
    		// do something
    	} else if(data == 'b') {
    		// do something else
    	} else {
    		found = false;
    	}

    	if(found) _bt.read();
    }

(It's set up like this because it's a library, and if unknown messages are found they should be passed to the code handling other commands, if that makes sense)

At random points the Arduino would stop reacting to any received messages, and after adding some debug code I found out that while `_bt.available()` would keep incrementing with each message received, `_bt.peek()` returns some a nonsense negative value but `_bt.read()` will return something completely different, these are some i've encountered so far:

-1 == 255
-76 == 180
-1 == 255
-36 == 220
-1 == 255
-5 == 251
-72 == 184
-36 == 220
-18 == 238

I'm not sure where these values originate from because I'm never sending them to the HC-05 attached to this SerialPort. Worth noting is that once this happens even if i disconnect the HC-05 the SoftwareSerial will keep returning the last values, so I assume some buffering is done internally.

sandeepmistry:
Resolves #2858.

Prior to this, ```read``` and ```peek``` would return different values if the byte value was greater than 127.

For example, if the buffered byte was ```180```, ```read``` would return ```180```, but ```peek``` would return ```-76```.

### DIFF
--- a/hardware/arduino/avr/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/hardware/arduino/avr/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -482,5 +482,5 @@ int SoftwareSerial::peek()
     return -1;
 
   // Read from "head"
-  return _receive_buffer[_receive_buffer_head];
+  return (uint8_t)_receive_buffer[_receive_buffer_head];
 }
--- a/hardware/arduino/avr/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/hardware/arduino/avr/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -48,7 +48,7 @@ http://arduiniana.org.
 // Statics
 //
 SoftwareSerial *SoftwareSerial::active_object = 0;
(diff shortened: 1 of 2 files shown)


matthijskooijman:
I wonder if it would not be better to change the buffer to an `uint8_t` instead, removing the need for these casts. I don't see any reason for the buffer to use `char`, as it does now (even more because `char` is ambiguous in signedness, *if* is must be signed it should at least use `signed char` or `int8_t`).

sandeepmistry:
Yes, I agree. It can be changed to ``unsigned char`` (this is what HardwareSerial uses) or ```uint8_t```. This will also slightly clean up the ```read``` method.

I'll work on this, and then update the PR.

sandeepmistry:
@matthijskooijman I've pushed the changes, if things look good, I will go ahead and squash the commits.

PR #4693, opened 2016-03-15
allow 3rd party boards that depend on sam core to use their own USB vid/pid and manufacturer/product strings

philmanofsky:
My company is creating a product (open source MIDI controller) that uses the same chip as the Arduino Due (SAM3X series). I created a new board package with a "boards.txt" file which uses the Due's core files. The changes that I propose in this Pull Request allow boards like my company's board (which depend on the sam core) to define their own USB vendor/product ID's and manufacturer/product strings in "boards.txt" that show up when enumerating the USB device. 

Please consider committing this request. Otherwise, my company's product will show up as "Arduino Due" in Digitial Audio Workstations like Ableton after people edit/upload the firmware.


sandeepmistry:
@philmanofsky thanks!

philmanofsky:
@sandeepmistry you are welcome and thank you! Question. When will this code be integrated as part of the "Arduino ARM" package that you install from boards manager? Should I have also changed the platform.txt version to 1.6.8? Again, thank you for the merge. That was super quick and is going to help my company a lot :)

philmanofsky:
@sandeepmistry also, is there an hourly build of the SAM core?

sandeepmistry:
> When will this code be integrated as part of the "Arduino ARM" package that you install from boards manager?

It will be available in the next release, however we do not have any timelines on this.

> is there an hourly build of the SAM core?

Unfortunately no. There are a few issues preventing us from splitting the SAM core into a new repo like the SAMD core, this is blocking having hourly builds.

philmanofsky:
@sandeepmistry if i want to speed up the process of a new sam core release, do I need to change "version=1.6.7" in the platform.txt file to "version=1.6.8"?

Issue #4387, opened 2016-01-05
Git integration inside the IDE (commit only)


naevtamarkus:
I would like to ask for a small feature request: it would be awesome to have small support for Git inside the Arduino IDE.

I understand GIT is really feature-rich and making the IDE fully compliant will be a full project on its own... but having a simple "Commit changes to this file" option in the File menu (with a key shortcut) if the file belongs to a git repo would be AWESOME already. It does not need to do much: just check that git is installed, check if the opened file is part of an existing local repository, and give the user a dialog with a 1-line commit message and OK/Cancel buttons if you click on it (or push the right key combination).

I think this will cover 90% of the use cases with Git, that is already a must for any-sized development projects. Having to do 'git commit -m blabla filename' on a separate console is quite some burden that makes you(me) not use it at the end.

Thanks for a great IDE!

lmihalkovic:
that one sounds like a good idea...

cmaglie:
I disagree, for me this covers 1% of the git usage, "commit" is just the very last action you do, most of the time you do diffs, add in staging files selectively (not altogether) or even part of a file. Less often you look the log, push/pull, merge or rebase.

This issue inspired #4865, #4816 and #4704 but using them leaves the feeling that they are incomplete, in particular you don't know what you're actually committing (this lead the author of #4865 to implement also the "Diff" command even if not explicitly request but it's really a draft and needs more work). Also having just 1 line for the commit message induce a bad habit, because commits should be well commented (possibly with more than 1 line if needed).

naevtamarkus:
I disagree, but I believe it's about perspective.
  - For an mid-expert developer that works with branches to code different features with other people that's absolutely what you should be doing with git, moving back and forth and diffing.
  - For an Arduino dummy like me who works on a single file on his own you use git with commits as a safety net for backtracking and eventually push something to github

Anyway... this is the magic with open source, you never know where a new door may lead you! I honestly think it would be a mistake to put a lid to contributions like those ones because "they are not complete". There is a lot of energy there! Featuring git in an IDE will never be a complete task (or else, it will never seem complete for an expert, otherwise it's a titanic task), but please bear in mind those Arduino newbies like me :-)

But I also understand the risks: if you are going on a road-trip, even if you don't know where are you really going, better make sure the car is in good condition and full of gas.

feraldruid22:
[...]

PR #4704, opened 2016-03-16
Git integration inside the IDE (commit only)

baydindima:
This add opportunity for commit with one-line message. New item "Commit" locate in Edit menu (I do not know how it wisely). I saw this issue https://github.com/arduino/Arduino/issues/4387 and decided to implement this. 
Please review and comment my solution.


cmaglie:
@ArduinoBot build this please

app/src/cc/arduino/view/git/GitCommit.java
```diff
@@ -0,0 +1,224 @@
...
+  private javax.swing.JPanel buttonsContainer;
+  private javax.swing.JButton cancelButton;
+  private javax.swing.JTextField commitMessageField;
+  private javax.swing.JButton okButton;
+  // End of variables declaration//GEN-END:variables
+
+  private String[] getArgsForProcess(String command) {
+    String[] args = new String[3];
+    if (OSUtils.isWindows()) {
+      args[0] = "cmd.exe";
```
cmaglie: Why not call directly git?

app/src/cc/arduino/view/git/GitCommit.java
```diff
@@ -0,0 +1,224 @@
...
+    args[2] = command;
+
+    return args;
+  }
+
+  private void setSystemEnvironment(Map<String, String> environment) {
+    environment.put("HOME", System.getProperty("user.home"));
+    environment.put("LANG", "en_US.UTF-8");
+    environment.put("GDM_LANG", "en_US.UTF-8");
+    environment.put("LANGUAGE", "us");
```
cmaglie: AFAICS you don't parse the output of git, so why you set the locale to en_US?

app/src/cc/arduino/view/git/GitCommit.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private void commit() {
+    String filePath = editor.getSketch().getMainFilePath();
+    StringBuilder command = new StringBuilder(String.format("git add %s && git commit %s ", filePath, filePath));
+
+    String message = commitMessageField.getText();
+    if (!message.isEmpty()) {
+      command.append(" -m \"")
+        .append(message)
```
cmaglie: This doesn't handle correctly commit messages that contains "quotes"

cmaglie:
Ok, it seems that @ArduinoBot is on holidays, so I've tested the patch manually.

First, some comments about the patch (I've added some other minor notes inline in the commits too)
- IMHO the "Commit" command should not appear in the menu if the "git" command is not available on the system path. As it is now you get a "file not found" error if you try to use it without git installed, but it's better to prevent the error message and not show the "Commit" command at all to the user if git is not available.
- In case "git" is available the "Commit" menu should be disabled (grayed out) in all the cases when the commit cannot be performed, for example, if the sketch folder is not a git repository.
- The patch doesn't handle sketch with multiple files (for example if you create a new tab with Ctrl+Shift+N the new file doesn't get committed, or if you remove a file the deletion is not committed)

Now, besides the notes above, It seems that the usecase is very narrow, becuase the user need anyway an external git client to "git clone" the sketch from a remote repository, or "git initialize" it after saving the sketch for the first time.
Also all the other operations (branching, push/pull, merge, etc..) needs an external client and seems improbable that all this being implemented in the Arduino IDE.
So I'm wondering if this patch is really worth the effort...

NicoHood:
The bot is on holidays?? O_o

I'd like to test a build when its available. However from what I've heard it might make more sense to just add a command line to the IDE, rather than a "not so useful half complete" git implementation.

Because windows users wont use git (or only the bright and shiny github gui) and linux users just know how to use proper software the normal way.

PR #4716, opened 2016-03-19
Corrected comments of rx and tx in a SoftwareSerial example.

asukiaaa:
Comments about rx and tx were wrong in TwoPortReceive.ino which is an example file of SoftwareSerial.
Thank you.


sandeepmistry:
@asukiaaa thank you!

asukiaaa:
@sandeepmistry Thank you for merging!

Issue #4680, opened 2016-03-11
Core lib - Stream class: getTimeout property


mattb5906:
Modify the Stream class to allow reading of the protected _timeout value.

`unsigned long getTimeout(void) { return _timeout; }`

[Arduino/hardware/arduino/avr/cores/arduino/Stream.h](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Stream.h)
[ Arduino/hardware/arduino/sam/cores/arduino/Stream.h](https://github.com/arduino/Arduino/blob/master/hardware/arduino/sam/cores/arduino/Stream.h)

sandeepmistry:
Closed via #4718.

PR #4718, opened 2016-03-19
Add getTimeout accessor method.

mattb5906:
Adds the ability to read the protected _timeout member from the stream class. As per suggestion #4680 .


sandeepmistry:
@mattb5906 thank you!

sandeepmistry:
I've also ported this to the SAMD core in https://github.com/arduino/ArduinoCore-samd/commit/0a35b5535a7d4e3158a63af31b08f3ece3f8f826.

PR #4743, opened 2016-03-22
Add more of OS X's native Emacs keybindings

nopdotcom:
OS X supports many simple Emacs keybindings in native text fields. (It can do this because the system UI uses `Cmd` instead of `Ctrl` as its main modifier key.) Add bindings to functions which already exist in
`DefaultEditorKit`.

I have a `Ctrl-k` (cut to end of line) implementation, but it is an additional static `RecordableTextAction` class. I have not included it.


cmaglie:
Interesting, never know that, are those key-bindings standard in OSX?

cmaglie:
@ArduinoBot build this please

app/src/processing/app/syntax/SketchTextAreaDefaultInputMap.java
```diff
@@ -67,8 +67,16 @@ public SketchTextAreaDefaultInputMap() {
 
       put(KeyStroke.getKeyStroke(KeyEvent.VK_A, ctrl), DefaultEditorKit.beginLineAction);
       put(KeyStroke.getKeyStroke(KeyEvent.VK_E, ctrl), DefaultEditorKit.endLineAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_N, ctrl), DefaultEditorKit.downAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_P, ctrl), DefaultEditorKit.upAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_B, ctrl), DefaultEditorKit.backwardAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_F, ctrl), DefaultEditorKit.forwardAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_D, ctrl), DefaultEditorKit.deleteNextCharAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_Y, ctrl), DefaultEditorKit.pasteAction);
+      put(KeyStroke.getKeyStroke(KeyEvent.VK_W, ctrl), DefaultEditorKit.cutAction);
```
cmaglie: The 'cut' actions that here is set as Ctrl-W seems to be Ctrl-K instead.
From various sources I see that the mnemonics are:

Ctrl-K -> Kill (cut)
Ctrl-Y -> Yank (paste)

is that right?
All the other shortcuts works as expected.

cmaglie:
ok after commenting here https://github.com/arduino/Arduino/pull/4743/files#r57473913 I re-read the OP comment:

> I have a Ctrl-k (cut to end of line) implementation, but it is an additional static RecordableTextAction class. I have not included it.

and now I've all the pieces together: it seems that Ctrl-K works as "cut to the end of the line" if no text is selected, otherwise it works as a normal "cut".

@nopdotcom do you agree?

nopdotcom:
tl;dr; Your described behavior is a good improvement; for example, Ctrl-K then Ctrl-Y should put the line back the way it was. The real Ctrl-K is complicated, 

Long version: I hadn’t realized how complicated it actually is. 

The native OS X behavior for Ctrl-K mimics Emacs very well, but it is confusing. There is a *separate cut buffer* used by OS X Ctrl-K and Ctrl-Y, unrelated to the clipboard.

Every action taken with Ctrl-K appends to the *special buffer*:

- If the cursor has moved since the last Ctrl-K, start with an empty special buffer.
- If there is a selection, append the selection into the special buffer, and delete the selection.
- If not at end of line, append the remainder of the line to the special buffer, and delete.
- If at end of line, add a newline to the special buffer, and delete the newline.

Ctrl-Y pastes the special buffer.

The end effect of this is that:

- Pressing Ctrl-K any number of times followed by Ctrl-Y is a no-op.
- From the start of line, Ctrl-K Ctrl-K cuts the whole line in a buffer, which you can then paste multiple times. In particular, Ctrl-A (beginning of line) Ctrl-K (cut line) Ctrl-K (cut newline) Ctrl-Y Ctrl-Y duplicates the current line.

These are wired into my fingers. But I would find it acceptable for Ctrl-A Ctrl-K Ctrl-Y RETURN C-y to duplicate the current line. Which would work with the mapping you described.

nopdotcom:
Would you like me to clean up my Ctrl-K code and add it to this branch?

cmaglie:
@nopdotcom 
Yes, thank you!
If you can keep the implementation of CTRL-K / CTRL-Y on a separate commit I can start to cherry-pick the other shortcuts.

cmaglie:
@nopdotcom I've merged your commit without the Ctrl+W and Ctrl+Y part, so we can move forward and close this PR.

https://github.com/arduino/Arduino/commit/02914d79b9d67c82ece6c9e0fe933dcfe141ccae
https://github.com/arduino/Arduino/commit/f74afc4ad9c3cd399e986a986e7abbc4383edb43

Feel free to open another PR when you have a correct implementation of the Kill/Yank feature!

Thank you!

PR #4750, opened 2016-03-24
modify the margin on the left 

CarlosMiguelCuevas:
I increase the margin of the scrollPane because of the issue #4562

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -293,7 +293,9 @@ public class Editor extends JFrame implements RunnerListener {
 
     // RTextScrollPane
     scrollPane = new RTextScrollPane(textarea, true);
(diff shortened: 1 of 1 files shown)


ivanebernal:
LGTM

cmaglie:
This patch leaves a lot of blank space on the left side I don't think that it's what the OP of #4562 intented when he asked for some blank space on the left.
Another ugly side effect is that the text seems to "disappear" when scrolling horizontally:

weird

Moreover this doesn't solve the problem reported by @per1234 here: https://github.com/arduino/Arduino/issues/4562#issuecomment-183281701
it's still difficult to select text from the first character of a line.

CarlosMiguelCuevas:
I don't think I will be working in this issue any further, @cmaglie could yo close it?

Issue #4789, opened 2016-03-31
Cut/Copy menuItems are always enabled


avargas-nearsoft:
In "Edit" menu, "Cut" and "Copy" options are always enabled regardless if there is or not selected text in textarea.

In fact there's a TODO for that issue: 
 ```
// TODO "cut" and "copy" should really only be enabled
    // if some text is currently selected
    JMenuItem cutItem = newJMenuItem(tr("Cut"), 'X');
    cutItem.addActionListener(new ActionListener() {
        public void actionPerformed(ActionEvent e) {
          handleCut();
        }
      });
    menu.add(cutItem);

    JMenuItem copyItem = newJMenuItem(tr("Copy"), 'C');
    copyItem.addActionListener(new ActionListener() {
        public void actionPerformed(ActionEvent e) {
          textarea.copy();
        }
      });
    menu.add(copyItem);
```

at processing.app.Editor (1372)

cmaglie:
Fixed by #4790

PR #4790, opened 2016-03-31
Disable cut and copy if there is not selected text

avargas-nearsoft:
Fix for https://github.com/arduino/Arduino/issues/4789

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1369,8 +1369,6 @@ public class Editor extends JFrame implements RunnerListener {
 
     menu.addSeparator();
 
-    // TODO "cut" and "copy" should really only be enabled
-    // if some text is currently selected
     JMenuItem cutItem = newJMenuItem(tr("Cut"), 'X');
     cutItem.addActionListener(new ActionListener() {
         public void actionPerformed(ActionEvent e) {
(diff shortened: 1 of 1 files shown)


app/src/processing/app/Editor.java
```diff
@@ -1515,6 +1513,21 @@ public void actionPerformed(ActionEvent e) {
...
+    menu.addMenuListener(new MenuListener() {
+      @Override
+      public void menuSelected(MenuEvent e) {
+        boolean enabled = textarea.getSelectedText() != null;
+        cutItem.setEnabled(enabled);
+        copyItem.setEnabled(enabled);
+      }
+
+      @Override
+      public void menuDeselected(MenuEvent e) {}
```
CarlosMiguelCuevas: why are you Overriding this?

CarlosMiguelCuevas: is it because MenuListener implements an interface?

CarlosMiguelCuevas:
LGTM

ivanebernal:
LGTM!

app/src/processing/app/Editor.java
avargas-nearsoft: Yes. Because I'm instancing an anonymous class from MenuListener, it's mandatory to implement every method. But I choose use only menuSelected for checking if there's selected text.

cmaglie:
Thank you!

PR #4793, opened 2016-04-01
[DO NOT MERGE] Fix bootloader and USB improvements on 32u4

facchinm:
Various fixes for USB-enabled AVR devices (32u4).
Commit 	197fe07 needs some community testing (to make sure that "advertising" as USB2.1 has no side affect on any platform)

### DIFF
--- a/hardware/arduino/avr/cores/arduino/CDC.cpp
+++ b/hardware/arduino/avr/cores/arduino/CDC.cpp
@@ -34,6 +34,8 @@ typedef struct
 static volatile LineInfo _usbLineInfo = { 57600, 0x00, 0x00, 0x00, 0x00 };
 static volatile int32_t breakValue = -1;
 
+bool _updatedLUFAbootloader = false;
+
 #define WEAK __attribute__ ((weak))
 
 extern const CDCDescriptor _cdcInterface PROGMEM;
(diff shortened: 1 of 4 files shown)


beaufortfrancois:
You may want to update comment as well to `0x210` at https://github.com/facchinm/Arduino/blob/cc773bf60cb498c5afe96e3e4652e2ca484b277b/hardware/arduino/avr/cores/arduino/USBCore.h#L133

hardware/arduino/avr/cores/arduino/CDC.cpp
```diff
@@ -129,10 +139,14 @@ bool CDC_Setup(USBSetup& setup)
...
-#else
-				*(uint16_t *)MAGIC_KEY_POS = 0x0000;
+				// Restore backed up (old bootloader) magic key data
+				if (magic_key_pos != (RAMEND-1)) {
+					*(uint16_t *)magic_key_pos = *(uint16_t *)(RAMEND-1);
+				} else {
 #endif
+				// Clean up RAMEND key
+					*(uint16_t *)magic_key_pos = 0x0000;
+				}
```
NicoHood: Here is an error. You need to add an #if around this too.

facchinm:
@NicoHood , I fixed the bug you spotted and squashed it, thanks for taking a look!

NicoHood:
Great! It compiles now fine. It adds some overhead to the older bootloader boards, but removes it for newer boards.

beaufortfrancois:
@facchinm @NicoHood You may want to update `usbVersion` comment to `0x210` to indicate 2.1 instead of 2.0 at https://github.com/facchinm/Arduino/blob/cc773bf60cb498c5afe96e3e4652e2ca484b277b/hardware/arduino/avr/cores/arduino/USBCore.h#L133

How about https://github.com/arduino/Arduino/blob/7350d5a10c6ee6db90b34bad1dbc4b6321523577/hardware/arduino/sam/cores/arduino/USB/USBCore.h also?

facchinm:
@beaufortfrancois  
Declaring as USB 2.1 and not providing BOS descriptor will hang enumeration on Win 8.1 and newer, so I'd let the patchset here until we have a viable solution.

beaufortfrancois:
@reillyeon Does that make sens?

NicoHood:
I thought those fixes were applied already?

@facchinm I made the official lufa bootloader Arduino compatible with a few changes. I think this is a better idea than modifying HoodLoader2, as I've changes a lot to non standard to get the maximum size reduction.

lufa has also additional fetures as eeprom, lock bit etc support (as it has more room without the serial bridge). The leds blink a bit different, but to be honest this is totally fine to me, as it will use the usb tx/rx leds which I think is quite a good idea anyways.

I have not added the old magic key support, as people who use the new bootloader will use the new ide. I wanted to keep lufa at least simple, feel free to add this little patch yourself.

facchinm:
@NicoHood , the latest patch (runtime discovery of bootloader version) was cherrypicked, 197fe07 instead brings some headache on win10 (if we are declaring as USB 2.1 we must provide a BOS descriptor, otherwise the drivers won't load at all). I still had no time to tackle this and find a good workaround.

facchinm:
BTW, the LUFA patches are very clever and we could really think about integrating them to be as mainline as possible :+1:

reillyeon:
Couldn't we just fix this on Windows 10 by responding with a basic BOS descriptor with a container ID?

facchinm:
The ram/flash usage overhead was quite high when I coded a basic BOS response, and in a resource constrained device like 32u4 it's a huge problem losing 500bytes for a WIN10 specific issue.
But if you have a patch for adding a light (and overridable by a PluggableUSB library) BOS descriptor I'll be really happy to merge it

reillyeon:
I'll see what I can do to shrink it. That seems like a lot of bytes.

NicoHood:
If "runtime discovery of bootloader version" was cherry picked, why is the commit still shown in this PR?

PR #4795, opened 2016-04-01
Add specifier to generated tool property keys

facchinm:
By doing so, runtime variable are resolved by using the board specific tools

real life example:
- Intel i585 defines sketchUploader (version 1.6.2)
- Intel arc32 defines sketchUploader (version 1.6.4)
- `runtime.tools.sketchUploader.path` gets the value of the last one processed

with this PR

`runtime.tools.sketchUploader.path.Intel.arc32` and `runtime.tools.sketchUploader.path.Intel.i586` get created

when resolving `{runtime.tools.sketchUploader}`, the routine searches for a key `runtime.tools.sketchUploader.Vendor.Architecture`

If found, the value is obtained by `{runtime.tools.sketchUploader.Vendor.Architecture}.getKey()`, which always contains the required value.
If no value is found, the old method is applied


facchinm:
Superseded by https://github.com/arduino/Arduino/issues/5199

PR #4800, opened 2016-04-02
Translated "Loading configuration..." splashscreen text

### LINKED ISSUE #4440: "Loading configuration..." splashscreen text never appear as translated
The first text on splashscreen - "Loading configuration..." never appears as translated, because the preferences.txt file which contains an info about UI language preference isn't readed at this point yet.

ivanebernal:
I had to move the splash text after the preferences file is created. Fixes #4440.


cmaglie:
Rebased and merged, thank you!

Issue #4778, opened 2016-03-29
Unexpected Ctrl-Delete behavior in Arduino IDE


dtwaddell:
When I press Ctrl-Delete, I expect that the word in front of the cursor would be deleted, but instead, the whole rest of the line is deleted.

avargas-nearsoft:
It seems it's a problem from http://bobbylight.github.io/RSyntaxTextArea/ , the JTextArea implementation that Arduino IDE used in his Editor. I don't know if change it would be feasible, but I opine the same that you: it's a little confusing this feature. 
I will work in that and I will send a pull request.
What do you think about @ivanebernal @carlosMiguelCuevas

CarlosMiguelCuevas:
@dtwaddell , Why do you expect such behaviour? Where is that feature described? is it in the doc? I tried it and it seems that the behaviour of the Ctrl+Delete is to delete everything that is ahead of the cursor.

ivanebernal:
@CarlosMiguelCuevas I think It's because this behavior is present almost everywhere, except some text editors like notepad.

avargas-nearsoft:
Yes. It's a common feature for every text editor but I don't know why (neither where) RSyntaxTextArea is using this implementation.

dtwaddell:
@avargas-nearsoft  Right, notepad is the only text editor I know of that doesn't just delete the single word in front of the cursor.  But, Ctrl+Backspace doesn't work in notepad either.  Aside from that, this is the general behavior for Windows and other OSs.  Ctrl+Backspace deletes the word behind the cursor, Ctrl+Delete deletes the word in front of the cursor, and Ctrl+arrows moves left or right one word at a time.  

I checked a few programs to check, and the single word Ctrl+Delete behavior occurs in Microsoft Word, Google Chrome, Adobe Illustrator, Windows Explorer, Notepad++, and Processing (although they also don't do it perfectly either -- the extra space is still left behind, so pressing Ctrl+delete twice will still only get rid of one word + space).

avargas-nearsoft:
I found this class https://github.com/bobbylight/RSyntaxTextArea/blob/d84a498dd8c2e042ced9ccde4c96ea1d54770f68/src/main/java/org/fife/ui/rtextarea/RTADefaultInputMap.java where, effectively, they are using this KeyStroke for delete the rest of line 
```

public RTADefaultInputMap() {

		super();

		int defaultModifier = getDefaultModifier();
[...]

PR #4815, opened 2016-04-04
Change the CTRL+DELETE behaviour to delete the next word

avargas-nearsoft:
According to https://github.com/arduino/Arduino/issues/4778 , this restores the "normal" behaviour for the KeyStroke (CTRL | Command) + DELETE to delete only one word in front of cursor instead of the rest of the line.

### DIFF
--- a/app/src/processing/app/syntax/SketchTextArea.java
+++ b/app/src/processing/app/syntax/SketchTextArea.java
@@ -30,6 +30,8 @@
 
 package processing.app.syntax;
 
+import java.awt.event.KeyEvent;
+import javax.swing.KeyStroke;
 import org.apache.commons.compress.utils.IOUtils;
 import org.fife.ui.rsyntaxtextarea.*;
 import org.fife.ui.rsyntaxtextarea.Token;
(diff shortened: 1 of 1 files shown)


ivanebernal:
The keyStroke does delete the word in front of the caret. LGTM.

avargas-nearsoft:
@cmaglie Could you take a look at this? I had some free time at work but now I think I could not take more time for the project. 
Thanks!

cmaglie:
Thanks for the patch!

I tested and it works for Windows and Linux, but it should be handled differently for MacOSX: right now the short-cut is set as Cmd+Delete (since MacOSX uses Cmd instead of Ctrl as `MenuShortcutKeyMask`) but it really should be Alt+Delete.

Could you fix this? you can use the method `OSUtils.isMacOS()` to check if the IDE is running on MacOSX.

avargas-nearsoft:
It now works on OSX.

cmaglie:
Thank you!

dtwaddell:
Thanks @avargas-nearsoft :D

matthijskooijman:
Hm, I wonder if it would be cleaner to change this in https://github.com/arduino/Arduino/blob/master/app/src/processing/app/syntax/SketchTextAreaDefaultInputMap.java instead? AFAICS that's where the other Arduino-specific input map changes are made, and having a single one of them in some random method somewhere else just causes confusion...

PR #4817, opened 2016-04-05
Emoro2560

inovatic-ict:
[EMoRo 2560.zip](https://github.com/arduino/Arduino/files/204042/EMoRo.2560.zip)


NicoHood:
Whats this? Has anyone checked if this file contains a virus? It looks like this is a spam bot?

cmaglie:
The file doesn't contains virus, it looks like a zipped source folder. I've blocked the user for spamming (at least I think... never tried this option...).

PR #4828, opened 2016-04-06
Filter examples based on contributed libraries by architecture

facchinm:
Adds "COMPATIBLE" filter for Contributed libraries, this is not enough to ensure that the correct examples get displayed because the `addOrReplace` routine overrides the previous results.
Adding an architecture-aware routine to handle this (overrides old library only if the the architecture field is compatible)

Solves #4762, to be tested with nightly (`arduino-builder` >= 1.3.15)


sandeepmistry:
@facchinm this looks good on OS X.

When I select a SAMD core board, the following examples are visible for the USBHost library:
* ADKTerminalTest
* KeyboardController
* MouseController
* USB_desc

and for SAM:
* ADKTerminalTest
* KeyboardController
* MouseController

Which lines up with:
 * SAMD: https://github.com/arduino-libraries/USBHost/tree/master/examples
 * SAM: https://github.com/arduino-libraries/USBHost/tree/master/examples

Issue #4002, opened 2015-10-20
Conflicting type for first argument of utoa with stdlib.h declaration


jlbirccyn:
Hello

In `hardware/arduino/sam/cores/arduino/itoa.h` utoa is declared as follow:

```
extern char* utoa( unsigned long value, char *string, int radix ) ;
```

in a gcc 4.9.3 for arm distribution, it conflicts with the declaration of the built-in utoa:

```
char *	_EXFUN(__utoa,(unsigned, char *, int));
```

(_EXFUN is defined as follow `#define	_EXFUN(name, proto)		__cdecl name proto
`)

cmaglie:
Is this specific for arm-gcc 4.9.3 or for other versions as well?
Can you provide a test sketch?

jlbirccyn:
Seems to be with 4.9.3. I do not have the error with 4.9.2. No need to provide a sketch. Go in `hardware/arduino/sam/cores/arduino` and type `arm-none-eabi-gcc -c -o test.o -I. WString.cpp` (or whatever name your compiler have, provided it is 4.9.3). You can get a 4.9.3 from https://launchpad.net/gcc-arm-embedded

jsanchezgarcia:
any updates on this?

neojou:
I have the same problem when trying to upgrade toolchain from 4.8.3 to 5.2.1.

patrickelectric:
Same error here :disappointed:

jlbirccyn:
Just sent a pull request to fix the problem

cmaglie:
Should be fixed with #5480.

Please reopen, or furhter comment, if not.

clovisf:
I have tested (**and confirm it works**) this fix with Arduino IDE Nightly Build 1.6.13 on windows 10 using a third-party board with ATSAMD21G18A, after burning the latest Arduino Zero bootloader to it using ATMEL Studio 7.0

PR #4830, opened 2016-04-06
Fix conflict with declaration in std lib and gcc version > 4.9.2. See issue #4002

jlbirccyn:
See issue #4002

### DIFF
--- a/hardware/arduino/sam/cores/arduino/itoa.h
+++ b/hardware/arduino/sam/cores/arduino/itoa.h
@@ -31,7 +31,14 @@ extern void itoa( int n, char s[] ) ;
 
 extern char* itoa( int value, char *string, int radix ) ;
 extern char* ltoa( long value, char *string, int radix ) ;
+#if __GNUC__ > 4 || \
+    (__GNUC__ == 4 && (__GNUC_MINOR__ > 9 || \
+                       (__GNUC_MINOR__ == 9 && \
+                        __GNUC_PATCHLEVEL__ > 2)))
+extern char* utoa( unsigned value, char *string, int radix ) ;
+#else
 extern char* utoa( unsigned long value, char *string, int radix ) ;
+#endif
 extern char* ultoa( unsigned long value, char *string, int radix ) ;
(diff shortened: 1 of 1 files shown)


jlbirccyn:
See issue #4002

sandeepmistry:
@jlbirccyn thank you for submitting this!

I just tried it with ```gcc-arm-none-eabi-5_3-2016q1```, but get the following build error:

```
/Users/smistry/Documents/Arduino/hardware/arduino/sam/cores/arduino/itoa.c:124:14: error: conflicting types for 'utoa'
 extern char* utoa( unsigned long value, char *string, int radix )
              ^
In file included from /Users/smistry/Documents/Arduino/hardware/arduino/sam/cores/arduino/itoa.c:19:0:
/Users/smistry/Documents/Arduino/hardware/arduino/sam/cores/arduino/itoa.h:38:14: note: previous declaration of 'utoa' was here
 extern char* utoa( unsigned value, char *string, int radix ) ;
              ^
exit status 1
Error compiling for board Arduino Due (Programming Port).
```

Is there any reason why the equivalent change from ```itoa.h``` is not in ```itoa.c```?

joshingyou:
Agree with @sandeepmistry just tried your change and got the same error about conflicting declaration of utoa

cmaglie:
The latest newlib provides `itoa` too, also this function should be put under guards. Weird that the compiler doesn't throw errors on `itoa`, maybe because in this case the stub has the exact same signature as the lib function.

```
arm-none-eabi/include$ grep [iu]toa -r *
stdlib.h:char *	_EXFUN(__itoa,(int, char *, int));
stdlib.h:char *	_EXFUN(__utoa,(unsigned, char *, int));
stdlib.h:char *	_EXFUN(itoa,(int, char *, int));
stdlib.h:char *	_EXFUN(utoa,(unsigned, char *, int));
```

cmaglie:
Fixed with #5480.
The implementation of `itoa` and `utoa` inside the Arduino core will be removed when the bundled arm-gcc will be upgraded.

PR #4864, opened 2016-04-14
Subtract one from USB_EP_SIZE in USB_SendSpace

### LINKED ISSUE #3946: Leonardo fails to send exactly 64 bytes.

sandeepmistry:
This avoids dealing with ZLP’s in ```USB_Send```, because the max packet size will be EP size - 1.

Alternative to #4138 to resolve #3946.

cc/ @NicoHood @embmicro @BlackBrix.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -255,7 +255,9 @@ u8 USB_SendSpace(u8 ep)
 	LockEP lock(ep);
 	if (!ReadWriteAllowed())
 		return 0;
-	return USB_EP_SIZE - FifoByteCount();
+	// subtract 1 from the EP size to never send a full packet,
+	// this avoids dealing with ZLP's in USB_Send
+	return USB_EP_SIZE - 1 - FifoByteCount();
 }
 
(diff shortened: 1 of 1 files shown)


NicoHood:
Was this internally discussed and tested?

agdl:
@NicoHood Yes, tested by me on Ubuntu and Windows. If you can also try it out is good :)

NicoHood:
A user reported that this PR breaks the raw hid code:
https://github.com/NicoHood/HID/issues/91

sandeepmistry:
@NicoHood thanks for letting us know.

Do you have any suggestions on how to move forward? Please note, the SAMD core has the same behaviour as this PR: https://github.com/arduino/ArduinoCore-samd/pull/159/commits/40d9554ed333c363fee2f384295b8bcf800a6a97.

NicoHood:
No I dont know how to solve this problem, but I remember that there were multiple solutions provided. I'd revert the PR for now, as this is defenitely implemented wrong and needs to move to the serial library, not the usb core.

PR #4871, opened 2016-04-15
fast serialUSB read function readBlock

sined23:
The main idea is that bytes are written and read into/from CDC buffer not one by one but block of bytes.

Serial_::accept write into CDC buffer as maximum as possible (only to not overflow the buffer)
Serial_::readb (you can change just read, not use readb) get from CDC buffer as maximum as possible or nessasary length

and one more important thing: in USBD_Recv "while (n--) *dst++ = UDD_Recv8(ep & 0xF);" should be replaced into UDD_Recv(ep & 0xF, dst, len);

after that read speed is incredible increases !!!  Up to 12Mbit/s

### DIFF
--- a/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
+++ b/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
@@ -191,7 +191,6 @@ void Serial_::accept(void)
 	// current location of the tail), we're about to overflow the buffer
 	// and so we don't write the character or advance the head.
 	while (i != buffer->tail) {
-		uint32_t c;
 		if (!USBD_Available(CDC_RX)) {
 			udd_ack_fifocon(CDC_RX);
 			break;
--- a/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
+++ b/hardware/arduino/sam/cores/arduino/USB/CDC.cpp
@@ -196,12 +196,13 @@ void Serial_::accept(void)
 			udd_ack_fifocon(CDC_RX);
 			break;
 		}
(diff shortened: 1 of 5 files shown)


hardware/arduino/sam/cores/arduino/USB/CDC.cpp
```diff
@@ -196,12 +196,13 @@ void Serial_::accept(void)
 			udd_ack_fifocon(CDC_RX);
 			break;
 		}
-		c = USBD_Recv(CDC_RX);
-		// c = UDD_Recv8(CDC_RX & 0xF);
-		buffer->buffer[buffer->head] = c;
-		buffer->head = i;
-
-		i = (i + 1) % CDC_SERIAL_BUFFER_SIZE;
+		uint8_t c[CDC_SERIAL_BUFFER_SIZE];
```
NicoHood: You are overwriting the c variable above. Remove it or rename the array.

sined23:
No, array is used instead of variable now. Pls check carefully and pls read my decription above

hardware/arduino/sam/cores/arduino/USB/CDC.cpp
sined23: No, array is used instead of variable now. Pls check carefully and pls read my decription above

sined23: deleted c variable. thanks

facchinm:
Moved to https://github.com/arduino/ArduinoCore-sam/pull/3

PR #4873, opened 2016-04-15
Implement Do-Not-Verify-After-Upload preference for Serial Uploads

gh-megabit:
This change picks up the existing upload.verify preference from the preferences dialog/preferences.txt and allows you to use this with your USB/Serial upload tool of choice. Configuration is already added for avrdude. This option stopped working somewhere around the update to 1.5.

It uses PreferencesMap.get to avoid an exception if the config is not found in platforms.txt - this should avoid problems with other cores and allow backwards compatibility.

The SAM and SAMD cores can be updated later if required. They won't benefit much from the change as the upload speed is so much better.


gh-megabit:
This resolves #1344

cmaglie:
Thanks!

PR #4877, opened 2016-04-16
SD: File open with O_WRITE would always append, even when O_APPEND is…

SodaqMoja:
… off

Without this fix it is not possible to open an existing file and truncate
the contents.


sandeepmistry:
Hi @SodaqMoja,

Thank you for submitting this pull request. Could you please resubmit to the following repository: https://github.com/arduino-libraries/SD

The `libraries/SD` folder is synced from this repo.

Issue #4870, opened 2016-04-15
New functionality in String class (WString.cpp/.h)


TobiasKnauss:
I need some new functions in the the String class.
I had created 2 posts about that in the arduino developers google group some hours ago, but the posts never appeared. Maybe the moderator is currently N/A...?

While waiting for the posts to appear, I have created the functionality and tested it.
It's ready to use and I'd be happy to see it implemented in the official version of the library:

```
WString.h:
  void trim(void);
  void trim(char remove);
  void trimStart(void);
  void trimStart(char remove);
  void trimEnd(void);
  void trimEnd(char remove);
  void padLeft(unsigned int newlength);
  void padLeft(unsigned int newlength, char add);
  void padRight(unsigned int newlength);
  void padRight(unsigned int newlength, char add);
  void cropLeft(unsigned int count);
  void cropRight(unsigned int count);
  void keepLeft(unsigned int count);
  void keepRight(unsigned int count);

WString.cpp:
void String::trim(void)
{
  trim (0x20);
}

void String::trim(char remove)
{
  trimStart(remove);
  trimEnd  (remove);
}

void String::trimStart(void)
{
  trimStart(0x20);
}

void String::trimStart(char remove)
{
  if (!buffer || len == 0 || remove == 0) return;
  char *begin = buffer;
  while ((*begin) == remove) begin++;
  len = buffer + len - begin;
  if (begin > buffer) memmove(buffer, begin, len);
  buffer[len] = 0;
}

void String::trimEnd(void)
{
  trimEnd(0x20);
}

void String::trimEnd(char remove)
{
  if (!buffer || len == 0 || remove == 0) return;
  char *end = buffer + len - 1;
  while ((*end) == remove && end >= buffer) end--;
  len = end + 1 - buffer;
  buffer[len] = 0;
}

void String::padLeft(unsigned int newlength)
{
  padLeft(newlength, 0x20);
}

void String::padLeft(unsigned int newlength, char add)
{
  if (newlength < len) newlength = len;
  if (!reserve(newlength)) return;
  unsigned int lenAdd = newlength - len;
  char *begin = buffer + lenAdd;
  memmove (begin, buffer, len);
  memset (buffer, add, lenAdd);
  len = newlength;
  buffer[len] = 0;
}

void String::padRight(unsigned int newlength)
{
  padRight(newlength, 0x20);
}

void String::padRight(unsigned int newlength, char add)
{
  if (newlength < len) newlength = len;
[...]

PR #4890, opened 2016-04-23
String class (WString.cpp/.h) new functions

TobiasKnauss:
See
https://groups.google.com/a/arduino.cc/forum/#!topic/developers/ryrMAJiqwf0

summary:
- constructors enhanced with "unsigned int bufferSize = 0" to allow a preset size for the internal char array malloc.
- void String::setReservePercentage (unsigned char percentage)
  adds a reserve on the realloc so that further small String concats don't have to realloc every time.
- void String::setDefaultReservePercentage (unsigned char percentage)
  sets a default value for reservePercentage, that is used when creating new Strings.

These 3 functions operate in init() and reserve().

Added String manipulation functions:
  // modification
  // notice:  void    function  (...) operates on           the object and returns void
  //          String  functionC (...) operates on a copy of the object and returns the copy (C=Copy)
  //          String\* functionS (...) operates on           the object and returns a pointer to the object (S=Self)
  #### existing before ####  void    replace (char find, char replace);
  String  replaceC(char find, char replace);
  String\* replaceS(char find, char replace);
  #### existing before ####  void    replace (const String& find, const String& replace);
  String  replaceC(const String& find, const String& replace);
  String\* replaceS(const String& find, const String& replace);
  #### existing before ####  void    remove (unsigned int index);
  String  removeC(unsigned int index);
  String\* removeS(unsigned int index);
  #### existing before ####  void    remove (unsigned int index, unsigned int count);
  String  removeC(unsigned int index, unsigned int count);
  String\* removeS(unsigned int index, unsigned int count);
  #### existing before ####  void    toLowerCase (void);
  String  toLowerCaseC(void);
  String\* toLowerCaseS(void);
  #### existing before ####  void    toUpperCase (void);
  String  toUpperCaseC(void);
  String\* toUpperCaseS(void);
  #### existing before ####  void    trim (void);
  String  trimC(void);
  String\* trimS(void);
  void    trim (char remove);
  String  trimC(char remove);
  String\* trimS(char remove);
  void    trimStart (void);
  String  trimStartC(void);
  String\* trimStartS(void);
  void    trimStart (char remove);
  String  trimStartC(char remove);
  String\* trimStartS(char remove);
  void    trimEnd (void);
  String  trimEndC(void);
  String\* trimEndS(void);
  void    trimEnd (char remove);
  String  trimEndC(char remove);
  String\* trimEndS(char remove);
  void    padLeft (unsigned int newlength);
  String  padLeftC(unsigned int newlength);
  String\* padLeftS(unsigned int newlength);
  void    padLeft (unsigned int newlength, char add);
  String  padLeftC(unsigned int newlength, char add);
  String\* padLeftS(unsigned int newlength, char add);
  void    padRight (unsigned int newlength);
  String  padRightC(unsigned int newlength);
  String\* padRightS(unsigned int newlength);
  void    padRight (unsigned int newlength, char add);
  String  padRightC(unsigned int newlength, char add);
  String\* padRightS(unsigned int newlength, char add);
  void    cropLeft (unsigned int count);
  String  cropLeftC(unsigned int count);
  String\* cropLeftS(unsigned int count);
  void    cropRight (unsigned int count);
  String  cropRightC(unsigned int count);
  String\* cropRightS(unsigned int count);
  void    keepLeft (unsigned int count);
  String  keepLeftC(unsigned int count);
  String\* keepLeftS(unsigned int count);
  void    keepRight (unsigned int count);
  String  keepRightC(unsigned int count);
  String\* keepRightS(unsigned int count);

detailed descriptions at
https://github.com/arduino/Arduino/issues/4870


cousteaulecommandant:
Instead of adding the -S versions, wouldn't it be better to just *replace* the functions returning `void` with functions returning `String&`?  That way you would be able to do things like `myString.trim().replace('\\','/').toLowerCase()`.
Also I think it would be simpler to just add a `copy()` method for when you want to explicitly operate on a copy of the String rather than adding a lot of new -C functions potentially increasing the library size (plus I think the -C suffix is a bit cryptic).

BillyDonahue:
@cousteaulecommandant, can't the proposed `s.dup()` already be expressed as `String(s)`?

cousteaulecommandant:
> @cousteaulecommandant, can't the proposed `s.dup()` already be expressed as `String(s)`?

Yes, but I think `string2 = string1.dup().trim().replace('\\','/').toLowerCase()` might look more consistent than `string2 = String(string1).trim().replace('\\','/').toLowerCase()`, if this "chain of methods" approach is to be followed.
Then again, maybe an explicit string creation is a better idea than a "method"-based copy, so you might be right.  I was just proposing an alternative to the functions proposed here.

facchinm:
I agree with https://github.com/arduino/Arduino/pull/4890#issuecomment-217368512 , and this discussion should be migrated to https://github.com/arduino/ArduinoCore-api since it allows very interesting stuff. We are not going to add `C` and `S` versions so I'm closing this PR

PR #4928, opened 2016-05-05
Issue 879

rei-vilo:
A9 on pin 26


facchinm:
I believe it was not intended to become a PR :smile: 
Closing as invalid

PR #4962, opened 2016-05-17
Add .eep preference support to the Serial Uploader

fulda1:
Hello,

I'm not good in java programming and using github, but I would like to present small update for uploadre to recognize new eep file and upload them to eeprom of arduino. See bug #1318 https://github.com/arduino/Arduino/issues/1318

It is necessary to extract also platform.txt from my patch-2


sandeepmistry:
Based on my notes in https://github.com/arduino/Arduino/pull/4963#issuecomment-233742690 I don't think this change is needed if we do decide to support .eep file writing. A .eep file will always be generated during AVR builds.

facchinm:
Closing as related to https://github.com/arduino/Arduino/pull/4963

PR #4963, opened 2016-05-17
Add support to use AVR dude to upload .eep files on sketch upload

fulda1:
Hello,

I'm not good in java programming and using github, but I would like to present small update for uploadre to recognize new eep file and upload them to eeprom of arduino. See bug #1318 https://github.com/arduino/Arduino/issues/1318

It is necessary to extract also SerialUploader.java from my patch-1

### DIFF
--- a/hardware/arduino/avr/platform.txt
+++ b/hardware/arduino/avr/platform.txt
@@ -98,12 +98,15 @@ tools.avrdude.config.path={path}/etc/avrdude.conf
 tools.avrdude.upload.params.verbose=-v
 tools.avrdude.upload.params.quiet=-q -q
 tools.avrdude.upload.params.noverify=-V
-tools.avrdude.upload.pattern="{cmd.path}" "-C{config.path}" {upload.verbose} {upload.verify} -p{build.mcu} -c{upload.protocol} -P{serial.port} -b{upload.speed} -D "-Uflash:w:{build.path}/{build.project_name}.hex:i"
+tools.avrdude.upload.params.uploadeep="-Uflash:w:{build.path}/{build.project_name}.eep:i"
+tools.avrdude.upload.pattern="{cmd.path}" "-C{config.path}" {upload.verbose} {upload.verify} -p{build.mcu} -c{upload.protocol} -P{serial.port} -b{upload.speed} -D "-Uflash:w:{build.path}/{build.project_name}.hex:i" {upload.uploadeep}
+
 
 tools.avrdude.program.params.verbose=-v
 tools.avrdude.program.params.quiet=-q -q
 tools.avrdude.program.params.noverify=-V
-tools.avrdude.program.pattern="{cmd.path}" "-C{config.path}" {program.verbose} {program.verify} -p{build.mcu} -c{protocol} {program.extra_params} "-Uflash:w:{build.path}/{build.project_name}.hex:i"
+tools.avrdude.program.params.uploadeep="-Uflash:w:{build.path}/{build.project_name}.eep:i"
+tools.avrdude.program.pattern="{cmd.path}" "-C{config.path}" {program.verbose} {program.verify} -p{build.mcu} -c{protocol} {program.extra_params} "-Uflash:w:{build.path}/{build.project_name}.hex:i" {program.uploadeep}
 
 tools.avrdude.erase.params.verbose=-v
 tools.avrdude.erase.params.quiet=-q -q
--- a/hardware/arduino/avr/platform.txt
+++ b/hardware/arduino/avr/platform.txt
@@ -98,14 +98,14 @@ tools.avrdude.config.path={path}/etc/avrdude.conf
 tools.avrdude.upload.params.verbose=-v
 tools.avrdude.upload.params.quiet=-q -q
 tools.avrdude.upload.params.noverify=-V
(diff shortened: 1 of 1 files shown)


Chris--A:
This (IMO) will need to be implemented with a switch in the settings. It is fairly well accepted that while using the IDE, what you place in the EEPROM while a sketch is running will stay there when you upload the next version of your code.

The keyword `EEMEM` is not widely known (if at all in the Arduino community) and this feature would cause absolute havoc with many peoples sketches and tutorials if enabled by default. 

It definitely would be a good feature to have. For one my [`EEWrap`](https://github.com/Chris--A/EEWrap#eewrap-library-for-avr-and-arduino-boards) library would have its productivity value increased significantly if you were able to initialize EEMEM variables.

As it is an AVR specific switch, careful consideration of where to put it in the IDE may be required.

fulda1:
Well, this is good "chicken / egg" question. Searching forums I found lot of questions why arduino not uploading those eep files. Once you would like to be conservative, then here can be configuration in prefferences.txt with default "false". Then you will see reaction of users.
I'm not sure I'm good programmer for that complex patch.

Chris--A:
You wont find questions arguing the former, due to it being accepted, and widely known. Arduino is a large platform, so a number of questions still only represents a small number of the total user base.

fulda1:
OK, you are right. But anyway, once you are interested in, you can try it can try it:
[HERE](http://www.fucik.name/Arduino/eemem.php) is description how to change the IDE.

But until [EEPROM](https://www.arduino.cc/en/Reference/EEPROM) library documentation will not contain info about using EEMEM directive, it will not be widely used.

sandeepmistry:
Hi @fulda1,

I am trying this change out with my Uno and the [eeprom_read.ino example sketch](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/libraries/EEPROM/examples/eeprom_read/eeprom_read.ino). However, when I add the following to the sketch:

```arduino
char ee_data[] EEMEM = { 'h', 'e', 'l', 'l', 'o' };
```

The uploading with verbose logging outputs:
```
avrdude: writing eeprom (5 bytes):

Writing | ################################################## | 100% 0.03s

avrdude: 5 bytes of eeprom written
avrdude: verifying eeprom memory against /var/folders/m9/y22yzb0d49x_kws_6jtwdphw0000gn/T/build754a3b352d03b45e4b62df0151980356.tmp/eeprom_read.ino.eep:
avrdude: load data eeprom data from input file /var/folders/m9/y22yzb0d49x_kws_6jtwdphw0000gn/T/build754a3b352d03b45e4b62df0151980356.tmp/eeprom_read.ino.eep:
avrdude: input file /var/folders/m9/y22yzb0d49x_kws_6jtwdphw0000gn/T/build754a3b352d03b45e4b62df0151980356.tmp/eeprom_read.ino.eep contains 5 bytes
avrdude: reading on-chip eeprom data:

Reading | ################################################## | 100% 0.02s

avrdude: verifying ...
avrdude: verification error, first mismatch at byte 0x0000
         0xff != 0x68
avrdude: verification error; content mismatch
```

Based on the following forum post, the optiboot boot loader for the Uno does not contain support for write EEPROM data with AVR dude: http://forum.arduino.cc/index.php?topic=335117.msg2310790#msg2310790

Which board have you been testing with?

Note: If the sketch contains no `EEMEM` variables an empty .eep file is created:
```
:00000001FF
```

fulda1:
Hello sandeepmistry,

I'm using default boot loader I received in my boards from china. I'm using Nano and Pro Mini boards (variety of, from various suppliers). All of them (minimum 3 types tested) support upload of eeprom.

Regarding empty eep file - yes, that is why I was added condition for minimum file size 13 octets, because empty intel hex file not mean 0 file size.

sandeepmistry:
Hi @fulda1,

I was able to do some more tests and have this patch working with my Leonardo and Mega 2560 boards.

I now understand why you check the minimum file size, AVR dude hangs if it tries to write a zero length .eep data:

fulda1:
Or it is possible it is specific for bootloader version or for MCU version. I have only 328 types and no 168.
Do you know this page: http://www.fucik.name/Arduino/eemem.php ??

matthijskooijman:
IIRC optiboot does not support writing to EEPROM when compiled with the default options. I guess you could be seeing this, since I think the Uno uses optiboot, but the Mega2560 does not?

Chris--A:
This should also be an option in the menus. Like where you can choose the native/programming usb for Due, this could also be a board specific option.

If it is enforced, there isn't an easy way to update an AVR program that has data already in the EEPROM. For building/debugging purposes this is really working against the user. I mean, lets face it, people do not use the EEPROM for static data.

Maybe for boards that support it, show an additional upload button with some nice modification to the graphic. This can signify upload EEPROM data as well. It would be better suited as a board specific menu item though as there are now probably more boards supported by the IDE, than the AVR's with EEPROM (or at least will be in the future).

facchinm:
I'm closing the issue here; if anyone wants to adopt this functionality (after carefully testing it with latest avrdude and a range of official boards) please reopen it using this link https://github.com/arduino/ArduinoCore-avr/compare/pr_4963

PR #5014, opened 2016-06-05
Fix buffer being overwritten by multiple twi_transmit calls

kellerkindt:
Fixes that more complex methods (like Stream::print(float)) do not work properly.

Without this fix, Wire.print(1.01f); results in '1' because Print::printFloat(double, uint8_t) performs multiple print() and therefore multiple twi_transmit calls. Also Wire.println("Heyho"); results only in a newline character.


sandeepmistry:
@kellerkindt thank you!

Issue #1981, opened 2014-04-01
Command line requires X11


Franky47:
I'd like to use the CLI in Arduino 1.5 on a headless Raspberry Pi (accessed through SSH only, no X11 environment), to make a continuous integration platform.

When calling arduino in the command line, Java complains that the X11 environment is not set and is required (which should not be necessary for CLI programs).

CLI output:

```
$ arduino --verify foo.ino
Exception in thread "main" java.lang.ExceptionInInitializerError
    at processing.app.Base.main(Base.java:113)
Caused by: java.awt.HeadlessException: 
No X11 DISPLAY variable was set, but this program performed an operation which requires it.
    at sun.awt.HeadlessToolkit.getMenuShortcutKeyMask(HeadlessToolkit.java:231)
    at processing.core.PApplet.<clinit>(Unknown Source)
    ... 1 more
zsh: exit 1     arduino --verify foo.ino
```

ea48gad:
Same issue with 1.5.8 on Debian Wheezy (Raspbian) running in a Raspberry Pi.

raquel-ucl:
Same issue - trying to do Continuous Integration on a Jenkins OS X node to verify Arduino files, but it needs to be headless. I've tried [ino](http://inotool.org) but it seems quite outdated. Is there any other workaround in the meanwhile? I've heard of [bii hive](https://asciinema.org/a/8324), I'll give it a try too, but any other recommendations welcome.

Cheers,
Raquel

ffissore:
We use jenkins all the time. To workaround the needs of the IDE, we have installed https://wiki.jenkins-ci.org/display/JENKINS/Xvnc+Plugin

raquel-ucl:
Thanks so much @ffissore, I'll give it a try.

raquel-ucl:
Sorry if this is getting a bit off-topic. @ffissore, In which platform are your running that plugin? I'm trying to test Arduino verification on a Jenkins Mac node and after much configuring and working around, I've got stuck.

ffissore:
On mac we don't use that plugin because the node is launched via java web start. This allows the java process on mac node to access the display. Indeed, every once in a while, the IDE pops up and executes its functional tests

raquel-ucl:
I'm not sure I understand. Does that mean it's impossible to test this on a Mac node? :(

ffissore:
[...]

PR #5132, opened 2016-07-14
Do not show splash screen when options are present

### LINKED ISSUE #5131: Arduino requires X11 DISPLAY where none should be needed
If you use Arduino from the command-line, it complains about requiring an X11 DISPLAY:

```
$ arduino --pref boardsmanager.additional.urls=http://arduino.esp8266.com/stable/package_esp8266com_index.json --save-prefs
Picked up JAVA_TOOL_OPTIONS: 
Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: java.awt.HeadlessException: 
No X11 DISPLAY variable was set, but this program performed an operation which requires it.
	at java.awt.SplashScreen.getSplashScreen(SplashScreen.java:117)
	at processing.app.Base.<clinit>(Base.java:91)
```

This forces people to implement ugly work-arounds using Xvfb in order to accomodate Arduino IDE: https://github.com/timelab/ADEM/blob/master/.travis.yml#L5
```yaml
  # Ensure we have a DISPLAY for the Arduino IDE
  - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_1.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :1 -ac -screen 0 1280x1024x16"
  - sleep 3
  - export DISPLAY=:1.0
```

dagwieers:
This ensures now splash-screen is shown when preferences are changed or CLI-mode is used.

More information in #5131.

However this does NOT fix #5131, since the Arduino IDE in CLI-mode still requires X11 (according to #1981).

### DIFF
--- a/build/linux/dist/arduino
+++ b/build/linux/dist/arduino
@@ -27,7 +27,8 @@ fi
 # Collect options to java in an array, to properly handle whitespace in options
 JAVA_OPTIONS=("-DAPP_DIR=$APPDIR" "-Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel")
 
-if [[ "$@" != *"--upload"* && "$@" != *"--verify"* && "$@" != *"--get-pref"* && "$@" != *"--install-board"* && "$@" != *"--install-library"* ]] ; then
+# Only show the splash screen when no options are present
+if [[ "$@" != *"--"* ]] ; then
 	JAVA_OPTIONS+=("-splash:$APPDIR/lib/splash.png")
 fi
 


build/linux/dist/arduino
```diff
@@ -27,7 +27,7 @@ fi
 # Collect options to java in an array, to properly handle whitespace in options
 JAVA_OPTIONS=("-DAPP_DIR=$APPDIR" "-Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel")
 
-if [[ "$@" != *"--upload"* && "$@" != *"--verify"* && "$@" != *"--get-pref"* && "$@" != *"--pref"* && "$@" != *"--install-board"* && "$@" != *"--install-library"* ]] ; then
+if [[ "$@" != *"--upload"* && "$@" != *"--verify"* && "$@" != *"--get-pref"* && "$@" != *"--save-prefs"* "$@" != *"--install-boards"* && "$@" != *"--install-library"* ]] ; then
```
matthijskooijman: The commit message mentions --board, but I don't see that here? Also, the commit removes --pref again, isn't that needed after all? The commit message doesn't mention it.

dagwieers: Right, I adapted my commit but not the commit-message. I chose to add `--save-prefs` instead of adding both `--board` and `--pref` since they go together when used CLI-only. In fact, if you would be starting the IDE with `--board` (which I guess is possible) the splash will still be there.

Maybe this does not cover all the use-cases, but it covers my use-cases cfr. the man-page:
https://github.com/arduino/Arduino/blob/master/build/shared/manpage.adoc

dagwieers: The original issue title was also wrong. Somehow I was hoping that this would fix the X11 DISPLAY requirement, but that's rooted much deeper into the code I learned from another issue :-/

matthijskooijman: > Maybe this does not cover all the use-cases, 

I wonder if it would be better to just suppress the splash screen whenever *any* options are passed, since keeping this list of options up-to-date is probably going to fail (again) in the future. The most-used invocation is without options, which should show a test screen. The invocations with options that do show the GUI and should (ideally) show a splash screen too are probably very seldomly used, so it might be acceptable to not show a splash screen in those cases.
> Somehow I was hoping that this would fix the X11 DISPLAY requirement (as the original error message was related to the Splash-screen), but that's rooted much deeper into the code I learned by experience :-/ So this is a minor improvement only.

Ah, I already wondered about that indeed :-)

dagwieers: That makes a lot of sense too. The Splash screen is not an integral/functional part of Arduino, so only showing it when options (not just arguments) are present makes sense. I will make a new proposal.

dagwieers:
@matthijskooijman I simplified this PR to a single small change as you proposed. I hope you like it ;-)

PR #5135, opened 2016-07-14
Send an USB remote wakeup if data need to be written

facchinm:
On Linux, setting autosuspend_delay_ms to N and control to auto allows the host pc to suspend the peripheral. Some Linux distro (Ubuntu, Mint) apply this behaviour by default.
If the sketch's prints where less frequent than N milliseconds the sketch prints would never arrive.
This patch allows sending a remote wakeup event to unsuspend the peripheral and allow the serial prints to be received.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -1,6 +1,7 @@
 
 
-/* Copyright (c) 2010, Peter Barrett  
+/* Copyright (c) 2010, Peter Barrett
+** Sleep/Wakeup support added by Michael Dreher
 **  
 ** Permission to use, copy, modify, and/or distribute this software for  
 ** any purpose with or without fee is hereby granted, provided that the  
@@ -265,6 +266,11 @@ int USB_Send(u8 ep, const void* d, int len)
 {
 	if (!_usbConfiguration)
 		return -1;
+
+	if (_usbSuspendState & (1<<SUSPI)) {
+		//send a remote wakeup
+		UDCON |= (1 << RMWKUP);
+	}
 
 	int r = len;
 	const u8* data = (const u8*)d;


matthijskooijman:
This sounds like it has been a fun debugging session, but I guess it's a good thing to fix (can't really comment on the implementation, though).

Is a similar fix needed for the 16u2 on the Uno and similar?

facchinm:
Quite funny debug session indeed :smile: On 16u2 it shouldn't be necessary, the OS **should** never suspend a peripheral if it doesn't expose the `REMOTE_WAKEUP` capability. 

The kernel never applies the autosuspend policy automatically (from [https://www.kernel.org/doc/Documentation/usb/power-management.txt](https://www.kernel.org/doc/Documentation/usb/power-management.txt) _This means that non-hub devices won't be autosuspended unless the user or a program explicitly enables it_ ). 

In any case, we know how to handle it now :wink:

NicoHood:
Have you tried compiling it for the u2 series? If this function is useless, you could add a #define around it, so we don't use an additional overhead for the small flash we have.

facchinm:
Hi Nico, I didn't test on 16u2 using the AVR core but I believe it should behave like the 32u4 (so the fix is needed). My answer was about stock 16u2 firmware on UNO (which doesn't expose as `REMOTE WAKEUP`). The overhead is 16bytes in flash, 0 bytes in RAM so I'd keep it without ifdefs

NicoHood:
Looks like the 16u2 should support this from the datasheet.

matthijskooijman:
> Looks like the 16u2 should support this from the datasheet.

Yes, but AFAIU the regular USB-to-serial 16u2 firmware does not advertise support, so there is no problem. The Arduino core intended for the 16u4 *does* advertise it, so it should also actually support it.

NicoHood:
There is not a single board available with a 16u4, so I do not understand what you are trying to point out.

matthijskooijman:
Uh, that should have read 32u4. I mean the Arduino core that runs on e.g. the Leonardo.

PR #5138, opened 2016-07-16
Contributed tools may not be platform-compatible -- this situation cu…

kevans91:
…rrently fails terribly. This may not be the ideal situation, but it would be nice if it didn't crash and burn the IDE if it happens. =)

With this change and manually replacing libastylej/liblistSerialsj, the IDE comes up relatively cleanly and I can move onto some other platform-related issues before going back and addressing the build issues.


kevans91:
Closing this, because there's an underlying problem here that I need to address.

PR #5178, opened 2016-07-29
Improve Examples menu

PaulStoffregen:
This patch changes the File > Examples menu to show libraries grouped by their type, either built in to the IDE, supplied by a platform, or custom installed.

Currently, libraries are grouped into "Examples from Libraries" and "Examples from Custom Libraries", with little rhyme or reason.  This patch attempts to give users a visual cue about the source of their many libraries, so they can understand which are from the IDE, which depend on their board's platform, and which libs are the ones they've chosen to install.

A few corner cases are automatically handled.  When the platform references another platform, both sets of their libraries are automatically shown.  In the custom libraries, any which are known to be incompatible are shown at the end using a sub-menu, similar to how the retired libs are shown.

examplesmenu

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1261,8 +1261,8 @@ public class Base {
     }
 
     // Add examples from libraries
-    ideLibs.sort();
     if (!ideLibs.isEmpty()) {
+      ideLibs.sort();
       label = new JMenuItem(tr("Examples from Built-in Libraries"));
       label.setEnabled(false);
       menu.add(label);
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -80,10 +80,6 @@ import static processing.app.I18n.tr;
  */
 public class Base {
 
-  public static final Predicate<UserLibrary> CONTRIBUTED = library -> library.getTypes() == null || library.getTypes().isEmpty() || library.getTypes().contains("Contributed");
-  public static final Predicate<UserLibrary> RETIRED = library -> library.getTypes() != null && library.getTypes().contains("Retired");
-  public static final Predicate<UserLibrary> COMPATIBLE = library -> library.getArchitectures() != null && (library.getArchitectures().contains("*") || library.getArchitectures().contains(BaseNoGui.getTargetPlatform().getId()));
-
   private static final int RECENT_SKETCHES_MAX_SIZE = 10;
 
   private static boolean commandLine;
@@ -1100,30 +1096,6 @@ public class Base {
     }
   }
 
-  public LibraryList getIDELibs() {
-    LibraryList installedLibraries = new LibraryList(BaseNoGui.librariesIndexer.getInstalledLibraries());
-    List<UserLibrary> libs = installedLibraries.stream()
-      .filter(CONTRIBUTED.negate())
-      .filter(RETIRED.negate())
-      .filter(COMPATIBLE)
-      .collect(Collectors.toList());
-    return new LibraryList(libs);
-  }
-
-  public LibraryList getIDERetiredLibs() {
-    LibraryList installedLibraries = new LibraryList(BaseNoGui.librariesIndexer.getInstalledLibraries());
-    List<UserLibrary> libs = installedLibraries.stream()
-      .filter(RETIRED)
-      .collect(Collectors.toList());
-    return new LibraryList(libs);
-  }
-
-  public LibraryList getUserLibs() {
-    LibraryList installedLibraries = new LibraryList(BaseNoGui.librariesIndexer.getInstalledLibraries());
-    List<UserLibrary> libs = installedLibraries.stream().filter(CONTRIBUTED).collect(Collectors.toList());
-    return new LibraryList(libs);
-  }
-
   private List<ContributedLibrary> getSortedLibraries() {
     List<ContributedLibrary> installedLibraries = new LinkedList<ContributedLibrary>(BaseNoGui.librariesIndexer.getInstalledLibraries());
     Collections.sort(installedLibraries, new LibraryByTypeComparator());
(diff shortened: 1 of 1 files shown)


pfeerick:
Looks good Paul. I tried this on my system which has a rather odd multi-sketchbook layout now to keep the core-specific library matching gremlins at bay :) The end result was rather good still... I especially like how the examples specific to the AVR core are grouped together. The PR also seems to fix an issue with some libraries not being put in the custom group where they belong (ie. the Adafruit Nokia library should have been in the examples from custom libraries grouping, which it now is).

**Stock IDE (Arduino sketchbook)**
stock-1 6 10

**PR5178 IDE (Arduino sketchbook)**
pr5178

app/src/processing/app/Base.java
```diff
@@ -1206,36 +1178,157 @@ public void rebuildExamplesMenu(JMenu menu) {
...
+    String platformName = null;
+    String referencedPlatformName = null;
+    String myArch = null;
+    TargetPlatform targetPlatform = BaseNoGui.getTargetPlatform();
+    if (targetPlatform != null) {
+      myArch = targetPlatform.getId();
+      platformName = targetPlatform.getPreferences().get("name");
+      platformLibraryPath = new File(targetPlatform.getFolder(), "libraries");
+      String core = BaseNoGui.getBoardPreferences().get("build.core", "arduino");
+      if (core.contains(":")) {
```
matthijskooijman: Isn't there existing code to handle finding out the referenced platform? Or is that removed and moved into arduino-builder now? This would seem like the wrong place to have this logic, perhaps moving it into a Platform method might be more appropriate?

app/src/processing/app/Base.java
```diff
@@ -1206,36 +1178,157 @@ public void rebuildExamplesMenu(JMenu menu) {
...
+        }
+      // Other libraries of unknown type (should never occur)
+      } else {
+        otherLibs.add(lib);
+      }
+    }
+
     // Add examples from libraries
-    LibraryList ideLibs = getIDELibs();
     ideLibs.sort();
```
matthijskooijman: For consistency, this sort should be moved inside the next if, just like happens below.

app/src/processing/app/Base.java
```diff
@@ -1206,36 +1178,157 @@ public void rebuildExamplesMenu(JMenu menu) {
...
       menu.add(label);
-      for (UserLibrary lib : userLibs) {
+      for (UserLibrary lib : sketchbookLibs) {
+        addSketchesSubmenu(menu, lib);
+      }
+    }
+
+    if (!sketchbookIncompatibleLibs.isEmpty()) {
+      sketchbookIncompatibleLibs.sort();
+      JMenu incompatible = new JMenu(tr("INCOMPATIBLE"));
```
matthijskooijman: Should this list the boardname explicitly, e.g. "Incompatible with <board name>" to be more clear?

matthijskooijman:
Looks good to me. I added some nitpick comments to the code, but overall the code looks great.

app/src/processing/app/Base.java
PaulStoffregen: If TargetPlatform has a function to get the referenced platform within its API, I agree it should be used rather than parsing the build.core pref.  Likewise for finding the human-readable name of the platform.  But I didn't see anything like this inside TargetPlatform.  I'd be happy to change this to use an official API... I just need someone to tell me which API does this?

app/src/processing/app/Base.java
PaulStoffregen: Good question.  I tried to copy the style used above for retired libraries.

If the string becomes "Incompatible with {0}", what string should be shown?  If the platform human readable name is used, which one do you show if there's referenced platform?  Always the main platform?  What if it's null?  If the architecture ID is used, will it be human readable?  Will end users understand "samd"?

app/src/processing/app/Base.java
PaulStoffregen: Done https://github.com/arduino/Arduino/pull/5178/commits/9b324bc9cc5c101f403a994d5e53160ee89aa7c6

PaulStoffregen:
I added another commit to fix 1 of the 3 nitpicks.  :)

On the other 2, happy to edit, but I need some guidance.  If there isn't an API for finding referenced platforms or getting the human readable names (I didn't see any such thing), I believe adding such an API ought to be a separate pull request, which would update this and other places within the IDE.

app/src/processing/app/Base.java
matthijskooijman: I was thinking it should just list the board selected, not the architecture, since the board name will have more meaning than the architecture name. However, it does hide some information, since the real incompatibility is with the architecture used by the board of course.

matthijskooijman:
> If there isn't an API for finding referenced platforms or getting the human readable names (I didn't see any such thing), I believe adding such an API ought to be a separate pull request, which would update this and other places within the IDE.

I haven't looked to see if there is any code for this, I can imagine that this is no longer needed in the IDE with arduino-builder. If adding such an API, it might as well be a separate commit within this, pullrequest instead of a separate PR, IMHO. However, I don't think this should block this pullrequest, I guess it's perfectly acceptable is it is, it can be moved later if needed.

q2dg:
<offtopic> Sorry if I'm a little burensome, but I don't know why TFT, GSM and (maybe) Ethernet libraries are still displayed as not "retired". Let's clean up!</offtopic>

PaulStoffregen:
@q2dg - Did you try running the test build?  Do you have any opinion about this change to the Examples menu?  Or is your only input here which libs are retired?

q2dg:
@PaulStoffregen Sorry, I'm waiting to a nightly build to test it

I didn't want to polute this thread. My only input was about asking for clarify which official libraries have future and which no. I've written several issues here about this topic because for me is very important that information showed in web and in boards and libraries menu showed in IDE is updated. If not, the impression for a newbie is chaotic.

Thanks a lot for your work.

PaulStoffregen:
@q2dg - No need get a nightly build.  There's a test build right here.  Just scroll up.  It's the 2nd message, from ArduinoBot.  However, this pull request is 5 days old, so the nightly build should certainly have it by now.  Really, there shouldn't be any reason you can't actually try this, if you wanted to contribute in a productive way.

I know you wish to talk about changing the properties files within some libraries, but that's outside the scope of this work.  This pull request is about improving the Examples menu code.  As a simple matter of correctness, the code is supposed to show the libraries marked retired in the RETIRED sub-menu, and the others normally, according to the actual libraries properties data.  This code is *not* supposed to guess which libraries ought to become retired or otherwise disregard the actual library properties data.

q2dg:
Last nightly build is from 26th July.

I'm not developer, I'm just an user which takes care of the clarity and userfulness of documentation for users like me.

You're right this was not the place to ask for deprecate libraries/examples related to retired products. Sorry.

cmaglie:
> However, this pull request is 5 days old, so the nightly build should certainly have

No, the nightly is a build of the master branch, so it will include this PR only after merging.
At the moment the only build available is the ArduinoBot's build in the comment above.

damellis:
I haven't been following this closely, but I did have a question. Does this show examples from library versions that won't be used for compilation (because they're being overridden by a library of the same name in a higher-priority location)? If so, that seems confusing. To me, it seems like it would be better to only show examples from library versions that will be used.

PaulStoffregen:
> Does this show examples from library versions that won't be used for compilation (because they're being overridden by a library of the same name in a higher-priority location)?

No, it does not show those unusable libs, because it gets the list of all libraries from librariesIndexer.getInstalledLibraries().

Of course, this depends on librariesIndexer matching what arduino-builder actually does.  Whether the Java & Go code perfectly implement the same algorithm is a good question.  But if they don't perfectly agree, that's an issue with librariesIndexer.

matthijskooijman:
>  Does this show examples from library versions that won't be used for compilation (because they're being overridden by a library of the same name in a higher-priority location)? If so, that seems confusing. To me, it seems like it would be better to only show examples from library versions that will be used.

That would be a useful distinction to make (though I wouldn't hide unused version but mark them as such), but also tricky to implement. In part due to the java vs go distinction, but that might be solved by calling arduino-builder to sort these things out. But more importantly, the question of what library will or will not be used, depends highly on circumstances. If two libraries share one header file but also have another, different header file, which library is used depends on which of these header files are actually included.

PaulStoffregen:
> the question of what library will or will not be used, depends highly on circumstances

Perhaps these are 2 different questions?  The builder *will* use specific libraries depending on the include lines it discovers, which depends on the compiler's preprocessing which takes conditional compilation into account.  The librariesIndexer in Java is concerned with the simpler question of all the libraries the builder *could* possibly use.

The examples menu has always offered all the libraries which *could* be used.  It's always hidden the redundant copies of libraries which can't actually be used.  This patch preserves that long-standing design, but reorganizes the list to give a better visual cue, so the end user can see whether the Servo library they're using will be the one that's built into Arduino, or a copy provided by the platform, or one they've installed which is overriding both of those.

PaulStoffregen:
@cmaglie - How do you feel about merging this for 1.6.11?  Too risky?

cmaglie:
@PaulStoffregen 
this is a very good implementation, thank you!

cmaglie:
@PaulStoffregen 
I need to revert this one, I didn't considered that this PR changes the ordering of the libraries, and I must clear this UX change internally (even to sync with Create) before merging.

I want to release 1.6.11 ASAP and this would block it again.

My apologies for the back and forth. I'll follow this up closely after 1.6.11 and make sure it go into 1.6.12.

PaulStoffregen:
Ok, that's fine.  There really no hurry.

facchinm:
@ArduinoBot build this please

00alis:
Sorry for the late feedback! I think from the UX point of view this works and it is pretty compatible to the Web IDE. Online we have 2 main tabs: BUILT-IN and FROM LIBRARIES, please check it out yourself on create.arduino.cc/editor (you just need a Arduino ID to login).

Here a screenshot:
screenshot 2016-09-05 14 54 37

Built-in tab is exactly the same as you show it in the Java IDE; in the From Library tab we just have a Tag for the architecture and we offer the opportunity to filter out the libraries based on the architecture of the board currently linked to the sketch. We also have search embedded in the panel, which is the feature users seem to use the most.

We are currently working on the complete restructure of the Library and Example API, it is going to be ready in a couple of weeks. This new release will also include the RETIRED section.

Thanks!

PaulStoffregen:
@cmaglie Any chance this can be reconsidered for 1.6.12 ?

PR #5197, opened 2016-08-01
Test json-aware builder

facchinm:


PaulStoffregen:
Are you sure it's a good idea to put this into the builder, rather than have the IDE which already manages platform stuff simply tell the building which toolchain to use?

PaulStoffregen:
Already there's quite a lot of redundant stuff between the builder and the IDE which must be maintained perfectly in sync to avoid bugs.  For example, which libraries are actually used needs to track perfectly between the IDE's File > Examples & Sketch > Include Library and what the builder will actually use when it compiles the sketch.

To illustrate the fragile nature of this situation, I believe a bug was recently added by the change to also use the library.properties architecture to alter the platform vs builtin vs sketchbook library precedence.  The corresponding change wasn't made on the IDE side in how libraries are actually presented to the user by the GUI.

Duplicating this sort of subtle platform dependency stuff might require much more long-term maintenance to keep the IDE and builder using an identical set of criteria.  I fear this might be making the whole system much more brittle in the long run...

facchinm:
Hi Paul,
you are super right, duplicating the effort and keeping things in sync is a PITA. Also, subtle differences could result in totally unexpected behaviours. Also, this PR doesn't solve the problem that the IDE still choose the wrong upload tool.

IMO, the oversimplified logic behind a lot of decisions that the IDE takes is not ready for a world with continuous updates and deployments. Small specialised tools would make the difference here, leaving the gui simple to code and to extend (which is the great advantage of Java)

In the long term, a collection of standalone tools (the builder, the uploader, the library crawler) which cooperate could fit better the great variety of situations the IDE needs to solve correctly.

For example, the library indexer could use specific OS notify methods (I'm thinking about https://fsnotify.org/) to cache the library resolution and provide snappy results. The IDE and the builder would benefit from this offload without any code duplication. 

This long term strategy needs a lot of effort which we can't spend right now because the toolchain bug must be solved really soon. @cmaglie is also coding a Java only solution to address it so we can choose which solution fits better

PaulStoffregen:
I agree, a fix is needed urgently.

PaulStoffregen:
The last thing I want to do is act act an obstacle.  Hopefully later there'll be an opportunity to consider the overall design....

facchinm:
Your contribution is invaluable so you will never be an obstacle, quite the opposite :smile: . Once the "crisis" is over we'll surely reconsider the design to make everyone happy and less bug prone

facchinm:
Superseded by #5199 (better, faster and non hacky solution). Closing this while leaving https://github.com/arduino/arduino-builder/issues/173 open but freezed

PR #5250, opened 2016-08-14
Update SoftwareSerialExample.ino

paulvivier:
Also includes Micros per https://www.arduino.cc/en/Reference/SoftwareSerial


sandeepmistry:
@paulvivier thank you!

PR #5269, opened 2016-08-20
Key "MimeType" is a list and needs a semicolon as trailing character

probonopd:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#mime-types

### DIFF
--- a/build/linux/dist/desktop.template
+++ b/build/linux/dist/desktop.template
@@ -7,6 +7,6 @@ Exec=<BINARY_LOCATION>
 Icon=<ICON_NAME>
 Terminal=false
 Categories=Development;IDE;Electronics;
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Commit looks good to me. However, it seems that the trailing ; is not strictly required. The spec you linked says (emphasis by me):

> Some keys can have multiple values. In such a case, the value of the key is specified as a plural: for example, string(s). The multiple values should be separated by a semicolon and **the value of the key may be optionally terminated by a semicolon**. Trailing empty strings must always be terminated with a semicolon. Semicolons in these values need to be escaped using \;.

probonopd:
At least desktop-file-validate complains when the trailing semicolon is missing,

PR #5297, opened 2016-08-26
Add '--useprogrammer' option to the man page

aykevl:
It took me quite a while to hunt down how to upload using a programmer from the command line, and then discovering there is a simple switch for it. I hope documenting it prevents other people from having to dig in the source code.


facchinm:
Thanks a lot!

PR #5300, opened 2016-08-27
Add all analog references supported by the ATtinyX5 series

### LINKED ISSUE #1739: wiring_analog.c doesn't properly set REFS2 bit on ATTiny platform(s).
If you search, you find a lot of folks suggesting the 2.56 volt analog reference with no external bypass cap by using an analogReference value of 6.

That ostensibly works if you look at the ATTiny data sheet... until you look up a bit and notice that the three bits of that value are not next to each other.

The original wiring_analog.c says:

#if defined(ADMUX)
        ADMUX = (analog_reference << 6) | (pin & 0x07);
#endif

But that shifts the top bit of the analogReference() value out of the byte. The correct thing to do - at least on the ATTiny platform - is

#if defined(ADMUX)
        ADMUX = (analog_reference << 6) | (pin & 0x07) | ((analog_reference & 0x4)?0x10:0);
#endif

That is, REFS0 is bit 6 in the ADMUX word, REFS1 is bit 7, and REFS2 is bit FOUR.

See 17.13.1 in the ATTiny85 data sheet.

mischnic:
Now it should possible to use the 2.56V voltage reference on the ATtinyX5's.

Should resolve #1739 , where the formula is derived from.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Arduino.h
+++ b/hardware/arduino/avr/cores/arduino/Arduino.h
@@ -61,10 +61,18 @@ void yield(void);
 #define FALLING 2
 #define RISING 3
 
-#if defined(__AVR_ATtiny24__) || defined(__AVR_ATtiny44__) || defined(__AVR_ATtiny84__) || defined(__AVR_ATtiny25__) || defined(__AVR_ATtiny45__) || defined(__AVR_ATtiny85__)
-#define DEFAULT 0
-#define EXTERNAL 1
-#define INTERNAL 2
+#if defined(__AVR_ATtiny24__) || defined(__AVR_ATtiny44__) || defined(__AVR_ATtiny84__)
+  #define DEFAULT 0
+  #define EXTERNAL 1
+  #define INTERNAL1V1 2
+  #define INTERNAL INTERNAL1V1
+#elif defined(__AVR_ATtiny25__) || defined(__AVR_ATtiny45__) || defined(__AVR_ATtiny85__)
+  #define DEFAULT 0
+  #define EXTERNAL 1
+  #define INTERNAL1V1 2
+  #define INTERNAL INTERNAL1V1
+  #define INTERNAL2V56 6
+  #define INTERNAL2V56_EXTCAP 7
 #else  
 #if defined(__AVR_ATmega1280__) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1284__) || defined(__AVR_ATmega1284P__) || defined(__AVR_ATmega644__) || defined(__AVR_ATmega644A__) || defined(__AVR_ATmega644P__) || defined(__AVR_ATmega644PA__)
 #define INTERNAL1V1 2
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@mischnic thanks for submitting. The changes look good to me testing with the following sketch (based on the AnalogInput example):

```arduino
int sensorPin = A0;    // select the input pin for the potentiometer
int ledPin = 13;      // select the pin for the LED
int sensorValue = 0;  // variable to store the value coming from the sensor

void setup() {
  // declare the ledPin as an OUTPUT:
  pinMode(ledPin, OUTPUT);

  analogReference(EXTERNAL); // added
}

void loop() {
  // read the value from the sensor:
  sensorValue = analogRead(sensorPin);
  // turn the ledPin on
  digitalWrite(ledPin, HIGH);
  // stop the program for <sensorValue> milliseconds:
  delay(sensorValue);
  // turn the ledPin off:
  digitalWrite(ledPin, LOW);
  // stop the program for for <sensorValue> milliseconds:
  delay(sensorValue);
}
```

Here are the sketch size differences:

| Board | Before | After | Difference |
| ------ | ------ | ----- | ----------- |
| Uno | 1,042 bytes | 1,054 bytes | 12 bytes |
| ATtiny25 |  748 bytes | 762 bytes | 14 bytes |
| ATtiny45 |  752 bytes | 766 bytes | 14 bytes |
| ATtiny85 |  752 bytes | 766 bytes | 14 bytes |

Without the additional call to `analogReference(EXTERNAL);` inside `setup()` there was no change in code size.

@cmaglie @facchinm @matthijskooijman any opinions on the extra code size? We could wrap the change in `wiring_analog.c` in an `#ifdef` to avoid the additional code size on non-ATtinyX5 boards like the Uno.

facchinm:
I'd personally avoid to change the ADMUX line for all targets, since the same can be accomplished with saner ` INTERNAL2V56` and ` INTERNAL2V56_EXTCAP ` defines and an #ifdef guard.

2016-09-07-092843_1366x768_scrot

Something like 
```
 +#elif defined(__AVR_ATtiny25__) || defined(__AVR_ATtiny45__) || defined(__AVR_ATtiny85__)
 +  #define DEFAULT 0
 +  #define EXTERNAL 4
 +  #define INTERNAL1V1 8
 +  #define INTERNAL INTERNAL1V1
 +  #define INTERNAL2V56 9
 +  #define INTERNAL2V56_EXTCAP 13
```
and
```
#if defined(ADMUX)
if defined(__AVR_ATtiny25__) || defined(__AVR_ATtiny45__) || defined(__AVR_ATtiny85__)
ADMUX = (analog_reference << 4) | (pin & 0x07);
#else
ADMUX = (analog_reference << 6) | (pin & 0x07);
#endif
```

sandeepmistry:
@facchinm that's a great suggestion!

@mischnic what do you think? Do you have time to update your pull request with the suggestion?

mischnic:
@sandeepmistry Ok

sandeepmistry:
@mischnic thank you!

PR #5309, opened 2016-08-29
remove extra semicolon

ChisholmKyle:
I was compiling with -Werror and this little error popped up


facchinm:
Looks good, thanks!

PR #5336, opened 2016-09-02
Improve sketches that use Serial comunication

biagiom:
Add support for boards that use Native USB port as Arduino/Genuino 101, Arduino/Genuino Zero, Arduino Due and boards based on ATMEGA32U4.

### DIFF
--- a/build/shared/examples/01.Basics/AnalogReadSerial/AnalogReadSerial.ino
+++ b/build/shared/examples/01.Basics/AnalogReadSerial/AnalogReadSerial.ino
@@ -11,6 +11,11 @@
 void setup() {
   // initialize serial communication at 9600 bits per second:
   Serial.begin(9600);
+  // wait for Serial port initialization.
+  // Needed for boards that use native USB port such as
+  // ATMEGA32U4 based boards, Arduino/Genuino 101,
+  // Arduino Due and Arduino/Genuino Zero (Native Port)
+  while(!Serial) ;
 }
 
 // the loop routine runs over and over again forever:
--- a/build/shared/examples/01.Basics/DigitalReadSerial/DigitalReadSerial.ino
+++ b/build/shared/examples/01.Basics/DigitalReadSerial/DigitalReadSerial.ino
@@ -12,6 +12,11 @@ int pushButton = 2;
 void setup() {
   // initialize serial communication at 9600 bits per second:
   Serial.begin(9600);
+  // wait for Serial port initialization.
+  // Needed for boards that use native USB port such as
+  // ATMEGA32U4 based boards, Arduino/Genuino 101,
+  // Arduino Due and Arduino/Genuino Zero (Native Port)
+  while(!Serial) ;
   // make the pushbutton's pin an input:
   pinMode(pushButton, INPUT);
 }
--- a/build/shared/examples/01.Basics/ReadAnalogVoltage/ReadAnalogVoltage.ino
+++ b/build/shared/examples/01.Basics/ReadAnalogVoltage/ReadAnalogVoltage.ino
@@ -11,6 +11,11 @@
 void setup() {
   // initialize serial communication at 9600 bits per second:
   Serial.begin(9600);
+  // wait for Serial port initialization.
+  // Needed for boards that use native USB port such as
+  // ATMEGA32U4 based boards, Arduino/Genuino 101,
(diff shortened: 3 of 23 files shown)


q2dg:
Thanks. I think https://www.arduino.cc/en/Serial/IfSerial should be updated too because it only refers to retired boards and not the newer ones.

Regarding this, I think comments in commit should be more generic to avoid the problem above: something like "boards with native usb (refer to specific documentation)" could do the trick instead of enumerating the concrete models ("Zero, 101, etc").

On the other hand...maybe a mention in examples to the use of SerialUSB object (is defined in every board with native usb port?)  would be appreciated.

Thanks

biagiom:
Hi @q2dg ,
I add some board names in order to provide more informations to newbies and people that start programming an Arduino board for the first time.
Moreover there are also boards as Arduino Due and Zero which have both the programming port and the native port but only when using the native usb port, you have to wait for the Serial port to connect and so it's useful to add the line `while(!Serial) ;`, especially when you want to send data and text to the Serial monitor in your `void setup() { }`
Since the sketches in Examples section have to be valid for all Arduino/Genuino boards, I thought to update them and to add more comments.
Regarding the SerialUSB object, reading the documentation on Arduino website, I found that this feature is available only for Arduino Due (https://www.arduino.cc/en/Reference/Serial).
Using boards that have the Native USB port, I notice that if you add the line `while(!Serial) ;` in `void setup() { }`, you have to open the Serial monitor to execute the next lines of code, but, as said above, this is necessary if you want to send some text to the Serial monitor (in fact if you don't add the line `while(!Serial) ;`, you could lose the serial data sent to the Serial monitor).
Hope to be quite clear.
Best regard.

q2dg:
Yes, specifying concrete board names is the best for newbies, I agree, but there's a lot of documentation in web which is outdated because of mentioning retired models...my suggestion was intended to prevent this in a future. But your reasoning it's perfectly fine, never mind.

The SerialUSB object is defined in Zero and MKR1000 (https://github.com/arduino/ArduinoCore-samd/search?utf8=%E2%9C%93&q=serialusb), Micro (https://github.com/arduino/Arduino/pull/4192) and 101 (https://github.com/01org/corelibs-arduino101/search?utf8=%E2%9C%93&q=serialusb). It seems to be an alias of Serial in some boards but (I think) not in all (sorry,  I don't understand very well this, my programming level is horrible). Anyway, I suppose your commit is ok, never mind again.

**However, I insist (again!) to anyone who read this in updating web documentation regarding this topic:  https://www.arduino.cc/en/Serial/IfSerial is outdated and, as @biagiom has pointed, https://www.arduino.cc/en/Reference/Serial too. **

Thanks

biagiom:
Hi @q2dg,
thanks a lot for your comments and your link to SerialUSB have been very helpful.
If you look at [USBAPI.h file (LINE 178)](https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/USB/USBAPI.h#L178) in the [Arduino core repository for Samd architecture (ArduinoCore-samd)](https://github.com/arduino/ArduinoCore-samd), you can see that SerialUSB is an object (instance) of Serial_ class and it is used to create a virtual serial communication over the USB port. Moreover if you look at [line 190 of variant.h file for MKR1000](https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkr1000/variant.h#L190) you can see that SerialUSB is an alias for Serial so every time you manage serial communication with the Serial object in your sketches, you are implicitly using the SerialUSB object (MKR1000 has one micro-usb port used for Serial communication).
You can also manage a Serial communication with other devices using the Hardware (UART) Serial port with the RX (0) and TX (1) pins.
In many new Arduino/Genuino boards you can find either a USB virtual Serial port and one or more Hardware Serial port (RX/TX pin).
For more info you can see : 
1) The Getting Started web page for Arduino Leonardo and Micro => [Differences from the Arduino Uno ](https://www.arduino.cc/en/Guide/ArduinoLeonardoMicro#toc9)
2) The great [documentation about Serial ports for Arduino/Genuino Zero](https://www.arduino.cc/en/Guide/ArduinoZero#toc9)

biagiom:
This pull request can be closed because the modified sketches don't send data to the Serial monitor in the _void setup()_ function and so the line _while(!Serial) ;_ is not really necessary.

PR #5338, opened 2016-09-03
Add a --get-build command line option

housel:
Allow build systems external to the Arduino IDE to retrieve build
definitions for a given platform and board. The --get-build option dumps
properties (whose definitions are found in the 3rd Party Hardware
Specification) that an alternative build system can interpret and use to
construct build commands and tool invocations.


sandeepmistry:
Hi @housel,

Thank you for taking the time to submit this pull request. However, at this time we prefer not to go ahead with the changes you've proposed.

Instead our focus is `arduino-cli` for command line use.

Issue #5200, opened 2016-08-03
www.arduino.cc/en/Reference/Int add: Arduino Zero


avandalen:
Change:
On the Arduino Due, an int stores a 32-bit (4-byte) value
to
On the Arduino Due and Zero, an int stores a 32-bit (4-byte) value

q2dg:
Well...I remember this https://github.com/arduino/Arduino/issues/4525 (and this https://github.com/arduino/Arduino/issues/3801)

cousteaulecommandant:
...And while we're at it, fix the signed overflow thing (#4511) (either remove that part from the documentation or make Arduino actually do what the documentation says)

cmaglie:
Actually the page says:

> On the Arduino Due and SAMD based boards (like MKR1000 and Zero),

I don't know if there the page has been edited in the meantime.

avandalen:
Dear Cristian

I will help you but I don’t know which article you refer?

Albert

Issue #3801, opened 2015-09-16
[DOC] A better data types explanation is needed


q2dg:
* In https://www.arduino.cc/en/Reference/Word should be added a note saying that only in AVR boards it is a 2-byte data type : in Due (**and Zero???**) it is a 4-byte data type (as it is already explained in https://www.arduino.cc/en/Reference/UnsignedInt) 

* Maybe should be added a table somewhere simply showing three columns for each data type (row): the bytes stored in both architectures (AVR & ARM) if it's signed or not and their resepective numerical boundaries. This would help to see, for instance, that in Due (**and Zero???**) the *word* type and the *unsigned long* type are the same and also the *int* type and the *long* type, whereas in AVR the *short* type is equal to *int* type. 

* @VogonJeltz points in #3797 that the *char* type in Due (**and Zero???**) is unsigned. I haven't a Due and I can't test it, but if it is true, it should be documented too!!

Former points show some mess in data types definitions depending on the architecture of the board (and besides, these definitions are different from ANSI C, as @VogonJeltz points again in #3797). I suppose it's very out of scope of this issue, but maybe a general review of how data types are defined in Arduino platform as a whole should be done. 

Thanks

agdl:
@q2dg As you have seen this argument is a little bit hard in the Arduino eco-system :( Since you referenced this issue in the newer one, I'm closing this (they are almost the same) and leaving the other one opened because it is newer so upper in the issue history

PR #5362, opened 2016-09-12
WString: add `toDouble`

Ivan-Perez:
`String::toFloat` internally converts the string into a double (using [`atof`](http://www.cplusplus.com/reference/cstdlib/atof/)) and then truncates into a
float, so why not to add a method to return the double?

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -740,6 +740,11 @@ long String::toInt(void) const
 
 float String::toFloat(void) const
 {
-	if (buffer) return float(atof(buffer));
-	return 0;
+	return float(toDouble());
+}
+
+double String::toDouble(void) const
+{
+    if (buffer) return atof(buffer);
+    return 0;
 }
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -745,6 +745,6 @@ float String::toFloat(void) const
 
 double String::toDouble(void) const
 {
-    if (buffer) return atof(buffer);
-    return 0;
+	if (buffer) return atof(buffer);
+	return 0;
 }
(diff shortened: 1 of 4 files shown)


sandeepmistry:
I've been trying this out today. One observation, on AVR:

```
sizeof(float) = 4
sizeof(double) = 4
```

So both floats and doubles are 32-bits. So, looks good to merge on AVR.

Will test non-AVR cores (SAM, SAMD) shortly, since we would want to port the new API to those as well.

sandeepmistry:
@Ivan-Perez were there some particular test sketches or values you used to determine that `String::toDouble()` was needed?

Ivan-Perez:
I didn't know that both float and double types were the same size. I supposed that double might be 64-bit, but I've just checked it and as you say they are the same type.

Then maybe this PR is not useful at all and can be closed without merging.

Chris--A:
@Ivan-Perez 
> Then maybe this PR is not useful at all and can be closed without merging.

Well it'll be useful for the 32-bit platforms that support a `double` that is larger than a `float`.

sandeepmistry:
@Ivan-Perez thank you!

@akash73 could you please add the new `toDouble` function to the [String Object](https://www.arduino.cc/en/Reference/StringObject) reference.

sandeepmistry:
This change has been ported to the SAMD core in https://github.com/arduino/ArduinoCore-samd/commit/c7c6f709b56f80b664196161fc9e7f203505b79d, and I've also opened a pull request for the 101 core: https://github.com/01org/corelibs-arduino101/pull/293.

Ivan-Perez:
Reviewing the code once again, I think that in case of failure both functions (`toFloat` and `toDouble`) should return **NaN** instead of **0**.

In the next days I'll create a new PR to address it.

q2dg:
Well...data types in deserve a coffe break...
https://github.com/arduino/Arduino/issues/4525
https://github.com/arduino/Arduino/issues/3801
https://github.com/arduino/Arduino/issues/5200
...

PR #5369, opened 2016-09-14
Increase buffer size for converting floats + doubles to String's

sandeepmistry:
Port of https://github.com/arduino/ArduinoCore-samd/pull/163

Troublesome sketch:
```arduino
#include <float.h> 

void setup() {
  Serial.begin(9600);
  while (!Serial);

  Serial.println(String(FLT_MAX));
  Serial.println(String(DBL_MAX));
  Serial.println(String(-FLT_MAX));
  Serial.println(String(-DBL_MAX));
}

void loop() {
}
```

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -19,6 +19,8 @@
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <float.h>
+
 #include "WString.h"
 
 /*********************************************/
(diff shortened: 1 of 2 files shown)


hardware/arduino/avr/cores/arduino/WString.cpp
```diff
@@ -108,14 +110,14 @@ String::String(unsigned long value, unsigned char base)
 String::String(float value, unsigned char decimalPlaces)
 {
 	init();
-	char buf[33];
+	char buf[FLT_MAX_10_EXP + 4 + decimalPlaces]; // +4, one for: 10, -, ., \0
```
matthijskooijman: What do you mean with "10" there? I can understand how -, . and \0 are added to the number of digits that FLT_MAX_10_EXP represents, but not what 10 means.

sandeepmistry: On AVR `DBL_MAX_10_EXP` is `38` and `Serial.println(String(-DBL_MAX));` prints out `-340282350000000000000000000000000000000.00`.

Which is `43` characters in length, then you need an extra byte for the `\0`, which brings us to `44`.

`44 - 38 - 2 = 4`. Maybe, I was thinking the extra character was for the `1's` digit ...

facchinm:
Partially moved to https://github.com/arduino/ArduinoCore-sam/pull/32

PR #5378, opened 2016-09-16
Some improvements in the library, concerning constness and increment operators.

sdicke:


libraries/LiquidCrystal/src/LiquidCrystal.h
```diff
@@ -44,22 +44,22 @@
...
-		uint8_t d0, uint8_t d1, uint8_t d2, uint8_t d3);
-  LiquidCrystal(uint8_t rs, uint8_t enable,
-		uint8_t d0, uint8_t d1, uint8_t d2, uint8_t d3);
-
-  void init(uint8_t fourbitmode, uint8_t rs, uint8_t rw, uint8_t enable,
-	    uint8_t d0, uint8_t d1, uint8_t d2, uint8_t d3,
-	    uint8_t d4, uint8_t d5, uint8_t d6, uint8_t d7);
+  LiquidCrystal(const uint8_t rs, const uint8_t enable,
+		const uint8_t d0, const uint8_t d1, const uint8_t d2, const uint8_t d3,
+		const uint8_t d4, const uint8_t d5, const uint8_t d6, const uint8_t d7);
```
Chris--A: Making code clear is great, but I think this overuse of `const` in places where it essentially does nothing for the user is hindering the readability. This will only ever benefit someone writing the implementation (if even then).

peabo7:
I agree with Chris--A:

> if((character > static_cast<uint8_t>(31)) && (character < static_cast<uint8_t>(127)))

is difficult to understand, and it does nothing that the previous 

> if((c>31)&&(c<127))

doesn't do.

> const uint8_t rs

This in an argument list doesn't do anything that uint8_t rs doesn't do.

The reason is that the argument doesn't imply anything about how it is going to be used subsequently.  But someone learning Arduino without prior knowledge of C++ might think this has something to do with the pin assignment not being immutable within the Stepper object.

The only cases where const is important in argument lists is where there is a pointer (or a reference) to an object which the function can reference which needs to be declared as an object which must not be changed.

libraries/GSM/src/GSM3CircularBuffer.cpp
```diff
@@ -8,9 +8,9 @@ This file is part of the GSM3 communications library for Arduino
 -- TCP/IP connections
 -- HTTP basic clients
 
-This library has been developed by Telef�nica Digital - PDI -
+This library has been developed by Telef?nica Digital - PDI -
```
sdicke: Obviously an encoding problem.

libraries/GSM/src/GSM3CircularBuffer.cpp
```diff
@@ -198,121 +198,121 @@ bool GSM3CircularBuffer::extractSubstring(const char* from, const char* to, char
...
 		printCharDebug(theBuffer[b]);
 	Serial.println();
 }
 
-void GSM3CircularBuffer::printCharDebug(uint8_t c)
+void GSM3CircularBuffer::printCharDebug(const uint8_t character)
 {
-	if((c>31)&&(c<127))
-		Serial.print((char)c);
+	if((character > static_cast<uint8_t>(31)) && (character < static_cast<uint8_t>(127)))
```
sdicke: Okay, both variants result obviously in the same binary result. I expected more memory usage because 31 and 127 are integer literals not unsigned integer literals.

Chris--A: This was only one example I pointed out, there are many c-style casts that have been changed.

Its an optimisation. The compiler will do conversions at compile time for literals.
For example, the line below will not have an `int` temporary created, then at run time convert it to a `long long`.

`long  long  i = 1234;`

libraries/GSM/src/GSM3CircularBuffer.h
```diff
@@ -38,14 +38,12 @@ The latest version of this library can always be found at
...
-#define byte uint8_t
-#endif
+typedef uint8_t byte;
 
 // These values have to be interrelated
 // To-Do: may we have just one? (BUFFERMASK)
-#define __BUFFERSIZE__ 128
-#define __BUFFERMASK__ 0x7F
+const byte __BUFFERSIZE__ = 128;
+const byte __BUFFERMASK__ = 0x7F;
```
Chris--A: The original defines benefit from the same optimisation that we looked at below (`static_cast<uint8_t>(31)`).
The compiler will use the correct type when used as a define. As a `byte` a temporary is needed to match the bit-width of the LHS.

PR #5391, opened 2016-09-20
Add a way to read the attached pin no for servos

rkost:
Sometimes it is nice to get this information so why not have a function for this?


matthijskooijman:
This seems like a reasonable change to me. However, should the isActive flag not be checked as well? Now, I believe you get 0 before attaching, and the previous pin value after detaching, which might not be entirely consistent. Also, if this is merged, the new function should be documented. Care to write some documentation that be pasted into the reference right away?

rkost:
You are right. The isActive flag should be checked. Documentation may be written by me in the next week. Where do I contribute the docs?

matthijskooijman:
In a comment here is fine, they can be formatted and copied to the site by the Arduino devs.

PR #5407, opened 2016-09-23
Minor change to example sketch comment

pfeerick:
Can't be the right word! ;)

### DIFF
--- a/build/shared/examples/02.Digital/Debounce/Debounce.ino
+++ b/build/shared/examples/02.Digital/Debounce/Debounce.ino
@@ -41,7 +41,7 @@ int buttonState;             // the current reading from the input pin
 int lastButtonState = LOW;   // the previous reading from the input pin
 
 // the following variables are unsigned long's because the time, measured in miliseconds,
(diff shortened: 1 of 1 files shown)


rkost:
I think 'can' is the right word here. You need to read it as conclusion. The number is to big than to store it to an integer variable. One may say '... so it is not possible to store in an int' or something like this

facchinm:
Wording is correct indeed :smile:

pfeerick:
@rkost Indeed... hence why I removed the 't... as it was originally "can't" ...  as "become a bigger number than can not be stored in an int" makes no grammatical sense! @facchinm might want to come back at apply this at some point ;)

facchinm:
Err, it turned out that in Italy we have some problems with double negatives :smile: 
After a discussion with the best linguists it turned out that you are right, so I'm merging this :wink: Sorry for closing too soon

pfeerick:
No worries... I probably didn't help with my joke about "can't"... I did the assume thing, thinking the commit would speak for itself... 😵

PR #5410, opened 2016-09-23
added support for missing LED_BUILTIN

jpralves:
Some boards don't define (yet) the LED_BUILTIN define and it will prevent blink from compiling. The small fix only sets the LED_BUILDIN to the value 13 if it doesn't exist.


cmaglie:
This should be really solved in the boards definition not in examples (many libraries now relies on the presence/absence of this value).

The absence of `LED_BUILTIN` means that there is no LED on the board (and that you should not toggle pin 13 nor any other pin, unless you know what you are doing).

Issue #5437, opened 2016-10-03
https downloads


NicoHood:
A lot of link/downloads in the source do use http for download while a secure https is available instead.
Those files can be found for example via github:
https://github.com/arduino/Arduino/search?utf8=%E2%9C%93&q=http%3A%2F%2F

Most important are the sources in build.xml:
https://github.com/arduino/Arduino/blob/master/build/build.xml

And also the avr cores:
https://github.com/arduino/Arduino/blob/master/hardware/package_index_bundled.json

Since arduino has https encryption you could possibly replace verything starting with `http://www.arduino.cc` with https. That would cover most links and would not break anything.

cmaglie:
All files downloaded from build.xml or from json indexes are verified with checksums (sha/sha256).
Changing everything to https will increase the servers load without any improvement in security.

Maybe there are some places where using to https will improve security but this is not in build.xml or inside the indexes.

NicoHood:
Its not about integrity, its about confidentiality (but still happy to see that some people understand the meaning of integirty finally =) ).

I personally do not want everyone to see what I download, no matter if its "just" arduino or not. I am not sure about sever loads, but https is the (... not future, actually the past) ... its time to use https. The server should be able to handle such downloads.

NicoHood:
Please conside the switch to https. Archlinux now switched every possible source to https and gpg signatures. And if i am right debian also has a policy about that:
https://www.archlinux.org/todo/use-gpg-signatures-and-https-sources/

Here is what google thinks about https:
https://developers.google.com/web/fundamentals/security/encrypt-in-transit/why-https

Everything is already covered in this PR:
https://github.com/arduino/Arduino/pull/5438

And the gpg signing is tracked here:
https://github.com/arduino/Arduino/issues/5619

Issue #5379, opened 2016-09-17
Packaging arduino with upstream avr-gcc or avrdude question


NicoHood:
Hey guys,
I recently got a TU for archlinux which means I can now package arduino for archlinux. The question came up if archlinux should link `avr-gcc` and/or `avrdude` with the distribution/upstream packages.

I'd like to present my opinion on this, but also would like to hear other opinions, as I might be wrong. It would be nice if one of you guys could answer, as I think you should know it best: @cmaglie @facchinm @sandeepmistry @matthijskooijman @focalintend @kriegsman

My opinion is to **not** package the upstream tools.
* Compilation size differs
* Assemler code differs, [makes it hard to debug](https://github.com/arduino/Arduino/issues/5236)
* Compiler specific code as in fastled could break
* New avrlibc definitions might give problems
* [Reverting of avrdude](https://github.com/arduino/Arduino/pull/5208) already showed such problems
* Atmel also uses 4.9.x which is quite new if you ask me

As a side note: We currently package avrdude 6.3 because of a [dependency issue](https://github.com/arduino/Arduino/issues/5094), but that is only temporary.

matthijskooijman:
@NicoHood, your arguments sound sane to me. However, does Arch have any policy on this? I know that Debian has a policy of only allowing a single piece of software to be present in one package, if other software needs it, the other software should use a dependency (I think the rationale is mostly for security, this prevents missing a copy of some vulnerable software hidden inside another package).

plantarum:
Is it even possible to package Arduino for redistribution? Debian has been stuck on version 1.05 for ages due to unresolved licensing issues: https://github.com/arduino/Arduino/pull/2703

NicoHood:
So I've still been trying to package arduino with the upstream avr-gcc. I [had to solve some issues](https://github.com/arduino/Arduino/issues/660#issuecomment-251142922) first, now the packaging can continue.

Its a good practise to download all files via makepkg and not within the build script itself. Its quite a lot to download and then i found the `light-bundle` option. I've tried it with `ant clean build -Dlight_bundle=true`, but that gave me some errors when launching the IDE in the 2nd run:

```
java.io.FileNotFoundException: /usr/share/arduino/hardware/package_index_bundled.json (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
[...]

PR #5438, opened 2016-10-03
Buildfixes: Better linux packaging and https

NicoHood:
This PR fixes several build bugs or adds missing features that are essential when packaging arduino for a linux distribution. More details can be found here https://github.com/arduino/Arduino/issues/5379#issuecomment-251198838 and here https://github.com/arduino/Arduino/issues/5437

The edits are small, please look at the separate commits.

The build can be called like this as an example with all new features:

```
ant clean build -Dlight_bundle=true -Dno_docs=true -Dlocal_sources=true
```


build/build.xml
```diff
@@ -834,8 +834,21 @@
 
   <!-- Retrieve tool -->
   <target name="untar-unzip-download" depends="untar-unzip-check" unless="${archive_file}_available">
+      <antcall target="untar-unzip-download-web" />
+      <antcall target="untar-unzip-download-local" />
+  </target>
+  <target name="untar-unzip-download-web" unless="local_sources">
```
cmaglie: Is this distinction really needed?

AFAIR if you copy the needed archives in the right place before launching the build, the file is not downloaded again (this is done in the `untar-unzip-check` that is a dependency of `untar-unzip-download`).

build/build.xml
```diff
@@ -220,7 +220,8 @@
   </target>
 
   <!-- copy library folder -->
-  <target name="assemble-libraries" unless="light_bundle">
+  <target name="assemble-libraries" depends="assemble-libraries-full, assemble-libraries-light" />
+  <target name="assemble-libraries-full" unless="light_bundle">
```
cmaglie: I guees this is to keep the empty `libraries` folder. Is it really needed?

build/build.xml
```diff
@@ -254,13 +258,17 @@
...
+  <target name="assemble-hardware" depends="assemble-hardware-full, assemble-hardware-light" />
+  <target name="assemble-hardware-full" unless="light_bundle">
     <copy todir="${target.path}/hardware">
       <fileset dir="../hardware">
         <exclude name="arduino/sam/**"/>
       </fileset>
     </copy>
   </target>
+  <target name="assemble-hardware-light" if="light_bundle">
+    <copy file="../hardware/package_index_bundled.json" todir="${target.path}/hardware" />
```
cmaglie: In this case the file `../hardware/package_index_bundled.json` is wrong, it should be a json with zero platforms defined. I don't know maybe the best thing to do is to add a check to the IDE...

build/build.xml
NicoHood: One of the biggest reasons I remember is that some paths are different for different architectures or not existant (mkdir -p). On top of that you need to always change the PKGBUILD and need to carefully check where the build script wants to have its downloaded files. This makes it harder to maintain.

This patch makes it simpler to search the download files from a local source instead.

build/build.xml
NicoHood: I had a lot of problems with this file. Currently I add a copy fo the avr core to the hardware path with a different name and patches to use the archlinux tools. However If i define this package in this json the IDE always crashs on a name not equal to "arduino". I am not sure if the file is required in this case.

If it should be empty, maybe you should add a stub with no data in it instead. However this works for now.

NicoHood:
Any problems left with the PR? I'd also like to address some more build changes, but possibly in another PR then.

NicoHood:
Another release has been tagged and still no progress in this PR. It makes it complicated to package arduino if those patches are not applied. Could you please consider to add this?

facchinm:
1.6.13 has been a bugfix release, so we tried to include the minimum number of changes. We are currently merging a lot of pending PRs and we'll surely analyze this in the next days.

NicoHood:
Thank you for the information.

If you consider to fix the suggested build improvements you might also want to think of adding GPG signatures: https://github.com/arduino/Arduino/issues/5619

I am about to put arduino into [community] of ArchLinux, but I hesitate to do that because of the missing gpg checks and also because there was no clear statement if we are allowed to do so. I see no point why we should not, but in any case, feel free to contact me for making a clear statement about yes or no. The ArchLinux team would be happy about some more clear statements and gpg checks. Thanks.

NicoHood:
I've resolved the merging conflicts. Please consider to merge this, as it is a real struggle to patch back the sources with those patches.

erlanger:
@facchinm yes, could you plese include this patch in the next release? It makes it difficult for us linux users  who have to wait longer for a new release (until the patches are merged manually). 

Tante grazie!!

NicoHood:
@facchinm any news about this? Another required fix would be to exclude arduino-builder from the build, but i first wanted this pr to be reviewed and merged. Can we see this soon please, as i need to create patches for every release?

NicoHood:
The arduino-builder download contains a file `platform.txt` and `platform.keys.rewrite.txt` that needs to be placed inside the hardware folder. This should be part of the IDE and not the arduino-builder download. I try to exclude arduino-builder from the package and those 2 files are missing.

cmaglie:
@NicoHood 
I've pushed into master the first three commits about:
- Https downloads
- No Docs bundled (`no_docs` option)
- Archives downloaded outside build folder (`local_sources` option)

can you rebase this PR on top of master so we can continue discussion on the other points?

NicoHood:
@cmaglie Thanks a lot! I rebased my PR.

And I've also removed the platform.txt again. A user pointed me to the information that it needs to be packaged with arduino-builder instead. It would be good if those 2 file would be inside the arduino-builder github repository though. Dont know where they are currently coming from in the zip download.

So left in my pr is the arduino-builder, hardware and libraries exclusion. I also want to exclude liblistserial and libastyle so that every dependency is built from source for arduino on archlinux. This makes it simpler and safer for us to link against shared libraries at build time.

cmaglie:
> I rebased my PR.

mmh not exactly, you merged the master branch in your PR, but this leaves a lot of useless history (various upstream merge, reverts, and the commits that I have already cherry picked duplicated in your branch).

I suggest to experiment with interactive rebase if you didn't already: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase

once rebased you can force push your branch and get a clean history with only the meaningful commits.

cmaglie:
> And I've also removed the platform.txt again. A user pointed me to the information that it needs to be packaged with arduino-builder instead. It would be good if those 2 file would be inside the arduino-builder github repository though. Dont know where they are currently coming from in the zip download.

Those file are here:
https://github.com/arduino/arduino-builder/tree/master/src/arduino.cc/builder/hardware

build/build.xml
cmaglie: `package_index_bundled.json` is, basically, a core-package json that describe what's bundled inside the IDE `hardware` folder. The IDE merges the content of `package_index_bundled.json` with the general `package_index.json` that is downloaded from arduino.cc. The bundled index has priority over the downloaded index.

If you want to make a real no-bundle (or light) version of the IDE, the correct thing to do is to create an empty (== with zero platforms defined) `package_index_bundled.json`.

build/build.xml
```diff
@@ -835,6 +779,24 @@
...
+  <target name="build-arduino-builder" unless="no_arduino_builder">
+    <delete dir="${staging_folder}/arduino-builder-${platform}" includeemptydirs="true"/>
+    <mkdir dir="${staging_folder}/arduino-builder-${platform}"/>
+    <antcall target="untar">
+      <param name="archive_file" value="./arduino-builder-${platform}-${ARDUINO-BUILDER-VERSION}.tar.bz2" />
+      <param name="archive_url" value="https://downloads.arduino.cc/tools/arduino-builder-${platform}-${ARDUINO-BUILDER-VERSION}.tar.bz2" />
+      <param name="final_folder" value="${staging_folder}/arduino-builder-${platform}/arduino-builder" />
+      <param name="dest_folder" value="${staging_folder}/arduino-builder-${platform}" />
+    </antcall>
+    <copy file="${staging_folder}/arduino-builder-${platform}/arduino-builder" tofile="linux/work/arduino-builder" />
```
cmaglie: This is wrong because the file is always copied into `tofile="linux/work/arduino-builder"` while it should be something like `${staging_folder}/work/arduino-builder`. BTW this is still not correct for macosx, probably the best way is to use `${staging_folder}/work/${staging_hardware_folder}/../arduino-builder` for everything.

build/build.xml
```diff
@@ -835,6 +779,24 @@
     <exec executable="./linux/work/arduino" spawn="false" failonerror="true"/>
   </target>
 
+  <target name="build-arduino-builder" unless="no_arduino_builder">
+    <delete dir="${staging_folder}/arduino-builder-${platform}" includeemptydirs="true"/>
+    <mkdir dir="${staging_folder}/arduino-builder-${platform}"/>
+    <antcall target="untar">
+      <param name="archive_file" value="./arduino-builder-${platform}-${ARDUINO-BUILDER-VERSION}.tar.bz2" />
```
cmaglie: For linuxarm the filename is `arduino-builder-linuxarm-1.3.23.tar.bz2`, but the file is actually called `arduino-builder-arm-1.3.23.tar.bz2`. I'm going to copy `arduino-builder-arm-1.3.23.tar.bz2` to `arduino-builder-linuxarm-1.3.23.tar.bz2` on the download server (and change the build script on jenkins to output the archive with `linuxarm` for the new releases).

cmaglie:
Superseded by #5778

PR #5443, opened 2016-10-04
Upgrade Serial Plotter

duff2013:
This adds a x-axis and labels, grid lines, more y-axis ticks and moves Serial Baud Rate combo box to the other side to reflect the look of the Serial Monitor. It also scales right when the window is scaled.
**Before:**
screen shot 2016-10-03 at 5 05 28 pm

**After:**
screen shot 2016-10-03 at 7 41 11 pm


duff2013:
I hope Arduino devs can merge this since it makes the Serial Plotter much more useful. If there are issues with the way I did thing let me know?

duff2013:
Even better, I like using the theme.txt for the grid colors, still trying to get handle on how everything fits together. I'll try to make this work.

cmaglie:
Thank you!

PR #5444, opened 2016-10-04
Add FreeBSD Platform Tag -- supports x86_64-freebsd, arm*-freebsd, an…

kevans91:
…d i386-freebsd with optional extension for tagging a specific release (e.g. freebsd11) if it's decided to do so later.

This should give us the widest range of flexibility/compatibility, I believe.

### DIFF
--- a/arduino-core/src/cc/arduino/contributions/packages/HostDependentDownloadableContribution.java
+++ b/arduino-core/src/cc/arduino/contributions/packages/HostDependentDownloadableContribution.java
@@ -79,6 +79,16 @@ public abstract class HostDependentDownloadableContribution extends Downloadable
       }
     }
 
+    if (osName.contains("FreeBSD")) {
+      if (osArch.contains("amd64")) {
+        return host.matches("x86_64-freebsd[0-9]*");
+      } else if (osArch.contains("arm")) {
+        return host.matches("arm.*-freebsd[0-9]*");
+      } else {
+        return host.matches("i386-freebsd[0-9]*");
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


kevans91:
Hi, ping @cmaglie, @facchinm =)

This seems like a fairly straightforward addition that shouldn't affect anyone else -- it would be nice if it could become a candidate for inclusion in 1.8.1 if you have time.

The arch matching doesn't get as in-depth as the other platforms because there's not really any need. Packages are not supplied by Arduino, so we build them ourselves on demand.

Because of this, we roll a custom version of the package_index_bundled.json that suits the machine that it's currently building on or for. Checksums and sizes go away, and we keep one "systems" entry that generally sets "host" to "$(uname -m)-freebsd$(major_version)" -- it's worth noting that the major version isn't actually worth anything here, because the package being built already reflects the FreeBSD version. I've included it, though, in case people try to do funky things.

Your consideration is appreciated, =)

facchinm:
LGTM :+1:

kevans91:
=)

Some additional thoughts: would you guys consider a PR adding some mechanism that allows either the packager or you to declare a platform as 'unsupported', bypassing these platform checks?

I ask because right now, I have to heavily patch the package_index_bundled.json to effectively do the above. This isn't necessarily a good solution -- I know my platform is unsupported, and I know that I have to put the work in to making sure that the dependencies are all there and proper.

Adding this host tag is still nice, though, because eventually I might be in a position where I can provide the dependencies bundled as you do and would rather do it through this mechanism.

PR #5480, opened 2016-10-13
Fix signature and definition of utoa to match name

kevans91:
Currently causes issues at the very least when compiled w/ newer versions of Newlib/arm-none-eabi-gcc.

### DIFF
--- a/hardware/arduino/sam/cores/arduino/itoa.c
+++ b/hardware/arduino/sam/cores/arduino/itoa.c
@@ -121,7 +121,7 @@ extern char* ltoa( long value, char *string, int radix )
   return string;
 }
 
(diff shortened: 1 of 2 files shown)


sandeepmistry:
Hi @kevans91,

Does this change still work with the old  Newlib/arm-none-eabi-gcc version that is installed by the Arduino IDE board manager?

kevans91:
@sandeepmistry: Yup, tested w/ an Arduino Due using 1.6.11 on Windows.

clovisf:
I have tested (**and confirm it works**) this fix with Arduino IDE Nightly Build 1.6.13 on windows 10 using a third-party board with ATSAMD21G18A, after burning the latest Arduino Zero bootloader to it using ATMEL Studio 7.0

PR #5549, opened 2016-11-01
Allow a default slave select value in SPI

GilesBathgate:
Some devices have an inverted slave select pin which is labeled such as C̅S̅ and has to be pulled LOW to select the slave instead of high (which is in contradiction to what the SPI spec says it should be). Assuming you have some of these devices and not a mixture of both, you might want to specify this in the begin method so as not to momentarily enable the device.

### DIFF
--- a/hardware/arduino/avr/libraries/SPI/src/SPI.cpp
+++ b/hardware/arduino/avr/libraries/SPI/src/SPI.cpp
@@ -25,6 +25,13 @@ uint8_t SPIClass::inTransactionFlag = 0;
 
 void SPIClass::begin()
 {
+  begin(HIGH);  
+}
+
+// slaveSelectDefault can be set LOW instead of the default HIGH if all
+// of your devices use inverted logic on slave select pin. 
+void SPIClass::begin(int slaveSelectDefault)
+{
   uint8_t sreg = SREG;
   noInterrupts(); // Protect from a scheduler and prevent transactionBegin
   if (!initialized) {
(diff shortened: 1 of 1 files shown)


matthijskooijman:
> Some devices have an inverted slave select pin which is labeled such as C̅S̅ and has to be pulled LOW to select the slave instead of high (which is in contradiction to what the SPI spec says it should be). 

I actually think that active-low SS is the most common one. Also, do you have a link for this SPI specification? AFAIK SPI is just an ad-hoc standard that most devices implement (more or less) the same, not an actual standard.

Furthermore, the SS pin changes in SPI.begin are not intended to control a slave, but they are to prevent the Arduino itself from accidentally switching into SPI *slave* mode (which happens if the SS pin is INPUT and LOW). When talking to SPI slaves, you are responsible for toggling the appropriate slave select pin yourself anyway.

To prevent the SS pin from becoming HIGH by SPI.begin, if you have an active-high device connected to it, you can just set its pinMode to OUTPUT *before* calling SPI.begin, since it will then not touch its value.

If this clarifies things and resolves your issue, please close this PR. If not, feel free to add more details about why you think this is needed.

GilesBathgate:
I understand how this works now.

- The SS pin represents the slave select when Arduino is a slave, its not the pin intended to be used to select a slave device. 

- The pin you choose to select a slave device is obviously an arbitrary choice but if you happen to use pin 13, then the setting the pin mode to OUTPUT before calling SPI.begin will work.

matthijskooijman:
Exactly! (though it's pin 10 instead of 13 on at least the Arduino Uno). Thanks for reporting back!

PR #5573, opened 2016-11-07
Fix symlinks being replaced with files on save

### LINKED ISSUE #5478: IDE tab as symbolic link is overwritten wit real file on "Save"
setup is:
/project1/project1.ino
/project1/localinclude1.h
/project1/include2.h
/include2.h

where /project1/include2.h is a symbolic link to /include2 and
/project1/localinclude1.h is a "real" file.

project1.ino includes both .h via 
# include "localinclude1.h"
# include "include2.h"

system is Linux Debian Jessie 64bit; arduino ide 1.6.11; file system is ext4
###### 

If I open /project1/project1.ino in arduino ide then all three files (project1.ino, localinclude1.h, include2.h) are shown in ide in tabs (= correctly).
I can edit and save all three files independently. If I save project1.ino or localinclude1.h from the ide all is fine and /project1/include2.h is still a symbolic link.

But if I save include2.h from the ide the symbolic link is replaces by a copy of /include2.h.
Thus, I then still have 
/project1/project1.ino
/project1/localinclude1.h
/project1/include2.h
/include2.h

but now /project1/include2.h is a "real" file (not a link any more) and is a copy of /include2.h.

This bug is really hard to recognize, as at first everything seems to be ok and continues to work.
Only if later you change something in /include2.h it won't show up in /project1/include2.h an v.v.

facchinm:
Fixes #5478

tested on Linux, please confirm that it works in Win and OSX before merging

### DIFF
--- a/arduino-core/src/processing/app/BaseNoGui.java
+++ b/arduino-core/src/processing/app/BaseNoGui.java
@@ -1061,11 +1061,22 @@ public class BaseNoGui {
   }
 
   /**
-   * Spew the contents of a String object out to a file.
+   * Save the content of a String into a file
+   * - Save the content into a temp file
+   * - Find the canonical path of the file (if it's a symlink, follow it)
+   * - Remove the original file
+   * - Move temp file to original path
+   * This ensures that the file is not getting truncated if the disk is full
    */
   static public void saveFile(String str, File file) throws IOException {
     File temp = File.createTempFile(file.getName(), null, file.getParentFile());
     PApplet.saveStrings(temp, new String[] { str });
+
+    try {
+      file = file.getCanonicalFile();
+    } catch (IOException e) {
+    }
+
     if (file.exists()) {
       boolean result = file.delete();
       if (!result) {


matthijskooijman:
The fix itself looks ok, but I'm wondering if we really need this "write to a tempfile, delete the original and move the tempfile"-approach here at all? Just opening the original file for writing would transparently handle symbolic links AFAICS?

Digging through git history shows that it used to work like that, but it was changed in upstream processing, without any comments about why this is: https://github.com/arduino/Arduino/commit/2e26a2d9#diff-b6fda0bf6be0a53c0741a9b02ac16f37R1898

Digging in the Processing repo gives [this commit](https://github.com/processing/processing/commit/d83d947e44b351bd19a5a67102e1c04af023bee3) and [this bug report](https://processing.org/bugs/bugzilla/967.html), which says the tempfile is to prevent truncating the file when the disk is full, which seems like a good reason.

The latest version of Processing has also fixed the symlink problem similar like in this PR (though it uses `File.getCanonicalFile()` which is a bit more compact than the code in this PR): https://github.com/processing/processing/blob/0abee5af6ad3b11cf2b73bb794b8a97c157c4762/app/src/processing/app/Util.java#L166-L203

@facchinm, could you update the PR to use `getCanonicalFile()`, and add a commit that adds a comment about *why* this tempfile approach is used?

facchinm:
Thanks for digging into Processing sources @matthijskooijman ! 
`File.getCanonicalFile()` looks good, I'm testing it and eventually replacing the commit as you suggested

facchinm:
@matthijskooijman , how does it look now? Functionality is ok

arduino-core/src/processing/app/BaseNoGui.java
```diff
@@ -1061,11 +1061,23 @@ static public String sanitizeName(String origName) {
...
+   * - Move temp file to original path
+   * This ensures that the file is not getting truncated if the disk is full
    */
   static public void saveFile(String str, File file) throws IOException {
     File temp = File.createTempFile(file.getName(), null, file.getParentFile());
     PApplet.saveStrings(temp, new String[] { str });
+
+    try {
+      File canon = file.getCanonicalFile();
+      file = canon;
```
matthijskooijman: I think this can just be `file = file.getCanonicalFile();`. If `getCanonicalFile()` throws, the assignment will not be executed.

PR #5578, opened 2016-11-07
Linux: Removed X11 requirement when running in command line mode

### LINKED ISSUE #1981: Command line requires X11

cmaglie:
Fix #1981 

This PR complete the work started in #2328 by me and @bitron and use also some refactorings made by @matthijskooijman on Editor and CommandLineParser.

The graphics elements are now created only when the IDE runs in GUI mode, this allowed to significantly reduce startup time (on my system I measured 25% less, from 1.750s to 1.350s).

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -264,35 +264,6 @@ public class Base {
 
     parser.parseArgumentsPhase2();
 
-    for (String path : parser.getFilenames()) {
-      // Correctly resolve relative paths
-      File file = absoluteFile(path);
-
-      // Fix a problem with systems that use a non-ASCII languages. Paths are
-      // being passed in with 8.3 syntax, which makes the sketch loader code
-      // unhappy, since the sketch folder naming doesn't match up correctly.
-      // http://dev.processing.org/bugs/show_bug.cgi?id=1089
-      if (OSUtils.isWindows()) {
-        try {
-          file = file.getCanonicalFile();
-        } catch (IOException e) {
-          e.printStackTrace();
-        }
-      }
-
-      boolean showEditor = parser.isGuiMode();
-      if (!parser.isForceSavePrefs())
-        PreferencesData.setDoSave(showEditor);
-      if (handleOpen(file, retrieveSketchLocation(".default"), showEditor, false) == null) {
-        String mess = I18n.format(tr("Failed to open sketch: \"{0}\""), path);
-        // Open failure is fatal in upload/verify mode
-        if (parser.isVerifyOrUploadMode())
-          showError(null, mess, 2);
-        else
-          showWarning(null, mess, null);
-      }
-    }
-
     // Save the preferences. For GUI mode, this happens in the quit
     // handler, but for other modes we should also make sure to save
     // them.
@@ -407,6 +378,35 @@ public class Base {
       System.exit(0);
     } else if (parser.isGuiMode()) {
       splash.splashText(tr("Starting..."));
+
+      for (String path : parser.getFilenames()) {
+        // Correctly resolve relative paths
+        File file = absoluteFile(path);
+
+        // Fix a problem with systems that use a non-ASCII languages. Paths are
+        // being passed in with 8.3 syntax, which makes the sketch loader code
+        // unhappy, since the sketch folder naming doesn't match up correctly.
+        // http://dev.processing.org/bugs/show_bug.cgi?id=1089
+        if (OSUtils.isWindows()) {
+          try {
+            file = file.getCanonicalFile();
+          } catch (IOException e) {
+            e.printStackTrace();
+          }
+        }
+
+        boolean showEditor = parser.isGuiMode();
+        if (!parser.isForceSavePrefs())
+          PreferencesData.setDoSave(showEditor);
+        if (handleOpen(file, retrieveSketchLocation(".default"), showEditor, false) == null) {
+          String mess = I18n.format(tr("Failed to open sketch: \"{0}\""), path);
+          // Open failure is fatal in upload/verify mode
+          if (parser.isVerifyOrUploadMode())
+            showError(null, mess, 2);
+          else
+            showWarning(null, mess, null);
+        }
+      }
 
       installKeyboardInputMap();
 
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -161,12 +161,6 @@ public class Base {
 
   }
 
-
-  static protected void setCommandLine() {
-    commandLine = true;
-  }
-
-
   static protected boolean isCommandLine() {
     return commandLine;
   }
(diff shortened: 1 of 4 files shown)


njh:
Just tested PR-5578-BUILD-614 on Mac OS 10.10.5 and Debian 8.6 Linux 32-bit.

* On Mac OS X it is working (output is written to console) but is still showing the splash-screen over the top
* But on Linux it is working great - no X11:
```
arduino --verify --board arduino:avr:uno Minimal/Minimal.ino
Picked up JAVA_TOOL_OPTIONS: 
Loading configuration...
Initialising packages...
Preparing boards...
Verifying...
Sketch uses 6,812 bytes (21%) of program storage space. Maximum is 32,256 bytes.
Global variables use 958 bytes (46%) of dynamic memory, leaving 1,090 bytes for local variables. Maximum is 2,048 bytes.
```

Working building of examples using Jenkins:
https://ci.aelius.com/job/ethersia/84/console

sandeepmistry:
These changes work well on the WiFi101 Travis CI build: https://travis-ci.org/sandeepmistry/WiFi101/builds/174214198 - no more `Xvfb ` needed! (https://github.com/sandeepmistry/WiFi101/commit/45f0f3e12bf80884d660d318e6f460487ee5a574)

app/src/processing/app/Base.java
```diff
@@ -1804,10 +1804,9 @@ static public LibraryList getLibraries() {
   }
 
   public File getDefaultSketchbookFolderOrPromptForIt() {
-
     File sketchbookFolder = BaseNoGui.getDefaultSketchbookFolder();
 
-    if (sketchbookFolder == null) {
+    if (sketchbookFolder == null && !isCommandLine()) {
```
matthijskooijman: Does this not break below, when `sketchbookFolder` is `null` and `sketchbookFolder.exists()` is called? This seems to be a bit of a corner case, since it only occurs when the *default* sketchbook path (as returned by `Platform.getDefaultSketchbookFolder()`) is null. I'm not exactly sure how if this should not normally happen, or if this will happen if the default path does not exist (but judging from the `mkdirs()` a bit further down, it's probably the former).

Looking through the code to see how this works, it seems that the responsibility for resolving the sketchbook path might be too much spread out right now, so perhaps some additional refactoring should be done after this PR in this area.

app/src/processing/app/Base.java
```diff
@@ -349,31 +350,42 @@ public Base(String[] args) throws Exception {
...
-      PreferencesData.set("build.verbose", "" + parser.isDoVerboseBuild());
-      PreferencesData.set("upload.verbose", "" + parser.isDoVerboseUpload());
-      PreferencesData.set("runtime.preserve.temp.files", Boolean.toString(parser.isPreserveTempFiles()));
+      PreferencesData.setBoolean("build.verbose", parser.isDoVerboseBuild());
+      PreferencesData.setBoolean("upload.verbose", parser.isDoVerboseUpload());
 
-      // Make sure these verbosity preferences are only for the
-      // current session
+      // Set preserve-temp flag
+      PreferencesData.setBoolean("runtime.preserve.temp.files", parser.isPreserveTempFiles());
```
matthijskooijman: Perhaps changing `set` to `setBoolean` should end up in a separate commit?

arduino-core/src/cc/arduino/contributions/ConsoleProgressListener.java
```diff
@@ -33,12 +33,19 @@
...
 public class ConsoleProgressListener implements ProgressListener {
   private String lastStatus = "";
+  private double lastProgress = 0.0;
 
   @Override
   public void onProgress(Progress progress) {
-    if (!lastStatus.equals(progress.getStatus())) {
+    // Reduce verbosity when running in console
+    String s = progress.getStatus().replaceAll("[0-9]", "");
+    double p = progress.getProgress();
```
matthijskooijman: This seems like a bit of a convoluted way to reduce verbosity here, but I'm not familiar enough with this Progress stuff to offer a better suggestion.

app/src/processing/app/Base.java
cmaglie: As you noticed the folders handling code is spread around the IDE, and it needs a bit of love :-) Actually, I tried to factor all the folder handling code in a single class while doing this PR, but at some point I've just dropped this task to reduce the possibility of regressions.

The rationale behind this change is:
- if my calculations are correct `BaseNoGui.getDefaultSketchbookFolder()` will never return `null`
- this means that the `promptSketchbookLocation()` will never be called (and could be probably removed altogether)

but since I decided to not touch the folder handling functions in this PR, I opted to just leave it as it is and just add the "guard" to not invoke the file selector in cmd-line mode and let the IDE crash with a NPE (if that code path is ever taken in some obscure ways).

matthijskooijman: Sounds reasonable.

app/src/processing/app/Base.java
```diff
@@ -195,9 +194,20 @@ public Base(String[] args) throws Exception {
 
     BaseNoGui.initPortableFolder();
 
+    // Look for a possible "--preferences-file" parameter and load preferences
     BaseNoGui.initParameters(args);
 
-    splashScreenHelper.splashText(tr("Loading configuration..."));
+    CommandlineParser parser = new CommandlineParser(args);
+    parser.parseArgumentsPhase1();
```
matthijskooijman: Right. Sounds like the commandline parsing needs another refactor after this one then (or, at least a `parseArgumentsPhase0` to just parse the preferences path).

arduino-core/src/cc/arduino/contributions/ConsoleProgressListener.java
cmaglie: The point of this change is to reduce the output of the lines like:
```
Downloading tools 10k of 15680k...
Downloading tools 20k of 15680k...
Downloading tools 30k of 15680k...
Downloading tools 40k of 15680k...
```

While this is nice in a GUI, with a progress bar etc., it's a mess on command line, this commit was the simplest way that comes to mind.

PR #5589, opened 2016-11-11
Updated Steper Motors Circuits URL

martinnm:
A quick update. The URL http://www.arduino.cc/en/Tutorial/Stepper doesn't exist anymore.

The new URL where a user can get the circuits is http://www.arduino.cc/en/Reference/Stepper


agdl:
Thank you!

PR #5623, opened 2016-11-23
Support the Atmel STK500 and JTAGICE3 for ISP

matthijskooijman:
This supersedes PR #2010. Compared to the last version of that PR, this:
 - Is based against master instead of ide-1.5.x
 - Removes the "slow" / "fast" distinction for JTAGICE3 ISP. Instead it now just uses the default bitclock, just like the other programmers (which is 8Khz, so sufficiently slow to work with even slow devices). The JTAG3ICE JTAG option does still have a bitclock option at the fasted possible setting (10Mhz), since for JTAG the max speed does not depend on the target clock, so this should always work.
 - Removes the comments about burn bootloader not working, since avrdude is now fixed to not barf about mismatching unused bits in the fuse and lock bytes.

I believe this resolves all reservations from #2010, so this PR can be merged as it is now.


matthijskooijman:
Thanks! :-)

PR #5635, opened 2016-11-25
Update Resources_pl.po

jacek-c:
Translated 1 line, replaced "kreślarka" -> "ploter"


facchinm:
Hi @jacek-c , to translate the IDE we use Transifex https://www.transifex.com/mbanzi/arduino-ide-15/ , you can fix the wording there and we'll import it before the next release. Thank you!

Issue #5617, opened 2016-11-22
Proposal for a new, more simple Network Upload Notification concept


holgerlembke:
Hi.

I think the current implementation of OTA port detecting is a little bit to buggy. There are a lot of reoccurring issues from people not getting it to work. (/me coming from the esp8266 side) I struggled the better part of some days to get it to work, never had stable success. Opinions about the reason differ, the state of esp8266mdns, jmdns and the Discoverer-concept (why use timers?) inside the IDE...

That said, I think there is way too many and complex software involved, I would like a more basic approach. This will give other hardware a very fast road to OTA update without porting any mdns stuff.

My current solution uses this inside the sketch:
```
String initmsg1 = "mysoftware version 1.0";

void PostMyself(void) {
  IPAddress broadcastIp(WiFi.localIP() | ( ~WiFi.subnetMask()));

  WiFiUDP udp;
  udp.begin(8765);
  udp.beginPacket(broadcastIp, 8531); 
  udp.print("1\nesp8266 "+initmsg1+"\n8266\n");
  udp.endPacket();
}
````
It can be located inside a "if (millis()-ticker>5000)"-Timer thing inside the loop(). IDE sees these broadcasts and adds the menue. 

Changes are (mainly) at https://github.com/holgerlembke/Arduino-IDE/tree/master/arduino-core/src/cc/arduino/packages/discoverers It is the basic outline, sure needs some work, but should be a good demonstration for this proposal.

I tested it with the current esp8266-ArduinoOTA solution, works.

Opinions?

sandeepmistry:
This is possible for third party cores now via the pluggable discovery feature.

PR #5655, opened 2016-11-30
Proposal for a new, more simple Network Upload Notification concept

holgerlembke:
I think the current implementation of OTA port detecting is a little bit to buggy. There are a lot of reoccurring issues from people not getting it to work. (/me coming from the esp8266 side) I struggled the better part of some days to get it to work, never had stable success. Opinions about the reason differ, the state of esp8266mdns, jmdns and the Discoverer-concept (why use timers?) inside the IDE...

That said, I think there is way too many and complex software involved, I would like a more basic approach. This will give other hardware a very fast road to OTA update without porting any mdns stuff.

My current solution uses this inside the sketch:
```
String initmsg1 = "mysoftware version 1.0";

void PostMyself(void) {
  IPAddress broadcastIp(WiFi.localIP() | ( ~WiFi.subnetMask()));

  WiFiUDP udp;
  udp.begin(8765);
  udp.beginPacket(broadcastIp, 8531); 
  udp.print("1\nesp8266 "+initmsg1+"\n8266\n");
  udp.endPacket();
}
````
It can be located inside a "if (millis()-ticker>5000)"-Timer thing inside the loop(). IDE sees these broadcasts and adds the menue. 

Changes are (mainly) at https://github.com/holgerlembke/Arduino-IDE/tree/master/arduino-core/src/cc/arduino/packages/discoverers It is the basic outline, sure needs some work, but should be a good demonstration for this proposal.

I tested it with the current esp8266-ArduinoOTA solution, works.


facchinm:
See #5617

sandeepmistry:
Hi @holgerlembke,

Thank you for taking the time to submit this pull request, at this time we would prefer not to go ahead with your proposal.

However, custom core can now use the new pluggable discovery feature: https://github.com/arduino/Arduino/pull/8038

PR #5701, opened 2016-12-10
Ethernet library: added ability to set/get hostname

mykh:
I've created overloaded Ethernet.begin() function that supports a host name parameter which is used by the DHCP request.
I've also added Ethernet.hostname() function which returns a hostname or an empty string if board was configured with static IP.
A simple example is located in the ./examples/DhcpHostname directory.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.h
+++ b/libraries/Ethernet/src/Ethernet.h
@@ -21,6 +21,7 @@ public:
   // configuration through DHCP.
   // Returns 0 if the DHCP configuration failed, and 1 if it succeeded
   int begin(uint8_t *mac_address, unsigned long timeout = 60000, unsigned long responseTimeout = 4000);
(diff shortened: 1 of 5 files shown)


libraries/Ethernet/src/Ethernet.h
```diff
@@ -31,6 +32,7 @@ class EthernetClass {
   IPAddress subnetMask();
   IPAddress gatewayIP();
   IPAddress dnsServerIP();
+  char* hostname();
```
Ivan-Perez: Shouldn't it be `const char* hostname()` ?

mykh: Fixed.

Misiu:
any updates?

Jozo132 (approved):
I want to see this implemented!

PR #5711, opened 2016-12-12
Improve judgement of Mega 2560 boot-loader entering programming mode

jimaobian:
Improvement for boot-loader of Mega 2560, source code and hex file:
1. Deal with the STK500V2 hanging issue here: 
https://github.com/arduino/Arduino/issues/2101
2. Fix the following problem:
If plugging the Arduino mega 2650 to the linux device, like Raspberry Pi and OpenWrt, it will automatically enter programming mode, waiting 7s for time-out, then starting to run the sketch, which is really annoying.

Tested on Mac, Linux and Windows with Arduino 1.6.13


facchinm:
Hi @jimaobian ,
we normally update the bootloaders once they get updated in production. The fix would need some refactoring and heavy testing (on all OS) before merging but there is no plan to change the current Mega bootloader in the near future. 

If someone lands here because he experienced problem with the Mega bootloader, please leave a message and we'll eventually adopt the issue and update the factory flashed BL.

facchinm:
Closing as per previous comment

PR #5716, opened 2016-12-13
Listening for compiler updates from tool

tomneutens:
Added the possibility to listen for compiler progress updates from a tool plugin. (Has to be tested more thoroughly)


facchinm:
Should fix  #5712
@ArduinoBot build this please

carlosperate:
This sounds really interesting! Any guidance or tips on how it can be integrated with external tools?

PR #5743, opened 2016-12-19
Add support for Arduino UNO Wifi

facchinm:
Both serial upload and OTA are handled.
A new tool is introduced to replace the `curl` based `avrflash` (part of `esp-link`)


facchinm:
Merged manually

PR #5747, opened 2016-12-22
Support printing constant strings containing nulls

eric-wieser:
What's changed
 1. `Serial.print("Hello""\0""World")`
     * Before: prints `Hello`
     * After: prints `Hello\0World`
 2. `Serial.print("Hello")`
     * Before: invokes `strlen`
     * After: does not, is faster. Code size should be unchanged
 3. `Serial.println("Hello")`
     * Before: invokes `strlen`
     * After: does not, is faster. Slightly increased code size, due to inlining

Thoughts?

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.h
+++ b/hardware/arduino/avr/cores/arduino/Print.h
@@ -50,6 +50,10 @@ class Print
       if (str == NULL) return 0;
       return write((const uint8_t *)str, strlen(str));
     }
+    template<size_t N>
+    inline size_t write(const char (str&)[N]) {
+      return write((const uint8_t *)str, N);
+    }
     virtual size_t write(const uint8_t *buffer, size_t size);
     size_t write(const char *buffer, size_t size) {
       return write((const uint8_t *)buffer, size);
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I'm not so sure if this is really a good idea, since now:
```
write("foo\0bar");
```
and
```
char *ptr = "foo\0bar";
write(ptr);
```

Will have different results and I'm not sure that's really a good idea.

Also, I *think* that:
```
char buf[10];
buf[0] = 'X';
buf[1] = '\0';
write(buf);
```

will now print 10 bytes, even though only 2 is intended. Since this sort of thing is likely to occur in sketches, this breaks compatibility.


Regarding handling of strings with embedded zeroes, you might be interested in #1936 which improves the String class to better handle them.

PaulStoffregen:
> Thoughts?

Isn't this what write(buf, length) is supposed to do?

Ideally, both write() functions ought to take (const void *) type, but whether even that can be considered changeable at this late stage in Arduino's development is a good question?

eric-wieser:
> I think that [...] will now print 10 bytes, even though only 2 is intended. Since this sort of thing is likely to occur in sketches, this breaks compatibility.

Yep, you're right - this is not an acceptable change

Issue #2397, opened 2014-10-31
Command line compile + upload - "Unidentified Reference to Setup/Loop"


ghost:
Hi I am trying to compile once and mass upload to multiple microcontrollers (pinoccio's specifically) when I am seeing this weird error of: 

```
core.a(main.cpp.o): In function `main':
/Applications/Arduino.app/Contents/Resources/Java/hardware/arduino/avr/cores/arduino/main.cpp:37: undefined reference to `setup'
/Applications/Arduino.app/Contents/Resources/Java/hardware/arduino/avr/cores/arduino/main.cpp:40: undefined reference to `loop'
collect2: error: ld returned 1 exit status
```

Occasionally, I am seeing this error as well:

```
/Applications/Arduino.app/Contents/Resources/Java/hardware/tools/avr/bin/avr-objcopy: error: the input file 'Bootstrap.cpp.elf' is empty
```

To build, I am using: 

```
/Applications/Arduino.app/Contents/MacOS/JavaApplicationStub --verify --pref build.path=/tmp/oyoroi /Users/vng/OyoroiPinoccio/libraries/pinoccio/examples/Bootstrap/Bootstrap.ino
```

And to upload, I am using: 

```
/Applications/Arduino.app/Contents/MacOS/JavaApplicationStub --upload --pref build.path=/tmp/oyoroi --port /dev/tty.usbmodem145311 /Users/vng/OyoroiPinoccio/libraries/pinoccio/examples/Bootstrap/Bootstrap.ino
/Applications/Arduino.app/Contents/MacOS/JavaApplicationStub --upload --pref build.path=/tmp/oyoroi --port /dev/tty.usbmodem145321 /Users/vng/OyoroiPinoccio/libraries/pinoccio/examples/Bootstrap/Bootstrap.ino
/Applications/Arduino.app/Contents/MacOS/JavaApplicationStub --upload --pref build.path=/tmp/oyoroi --port /dev/tty.usbmodem145411 /Users/vng/OyoroiPinoccio/libraries/pinoccio/examples/Bootstrap/Bootstrap.ino
/Applications/Arduino.app/Contents/MacOS/JavaApplicationStub --upload --pref build.path=/tmp/oyoroi --port /dev/tty.usbmodem145421 /Users/vng/OyoroiPinoccio/libraries/pinoccio/examples/Bootstrap/Bootstrap.ino
```

Does anyone know what's causing all these errors? 

Note that I pause for 5 seconds between compiling and uploading. 

[...]

PR #5749, opened 2016-12-22
Extract Stream methods into an IStream.

eric-wieser:
This allows streams to exist that are unidirectional.

Is this backwards-incompatible?


eric-wieser:
The diff here is largely uninteresting, and boils down to renaming both `Stream`s to `IStream`, and adding a new [`Stream.h`](https://github.com/eric-wieser/Arduino/blob/d4af77cde336a9dd89a92e8ff293a4b3803cb9c4/hardware/arduino/avr/cores/arduino/Stream.h)

Chris--A:
I do not think this is necessary. You can make the Print features private, hiding them from the user.

```C++

class MyStream : public Stream{
  public:
    int available()              { return 0; }  //Fill in these as appropriate.
    void flush()                 { return; }
    int peek()                   { return -1; }
    int read()                   { return -1; }
 
  private:
    using Print::print;
    using Print::println;
    using Print::write;
    size_t write( uint8_t data ) { return (void)data, 0; }
};

MyStream stream;

void setup() {
  stream.read();  //Fine
  stream.println();  //Compile error
}

void loop() {}
```

eric-wieser:
@Chris--A: That's a workaround, but not a good one:

```c++

void print_some_things(Print& p) {
   p.println('Hello world');
}

MyStream stream; // using your definition

void setup() {
    stream.read();
    print_some_things(stream);  // oops. This compiles just fine.
}
```

eric-wieser:
I'm open to other names here instead of `IStream`, such as:

* `InputStream`
* `Read` (to match `Print`)
* `Reader` (which avoids naming classes after verbs)

Chris--A:
That isn't really an issue either. My workaround hides the print features when using the object directly, and for the cases that the object is cast to one of its inherited types, the print functionality has no effect.

 A library shouldn't be left up to assumptions, if you don't support print features, then make it well known, or don't even mention it, then no one would have a reason to use them.

Another option using my original example and make the Stream base private and only expose what you want:

```C++
class MyStream : private Stream{
  public:
    int available()              { return 0; }
    void flush()                 { return; }
    int peek()                   { return -1; }
    int read()                   { return -1; }

    using Stream::find;
    using Stream::findUntil;
    using Stream::parseInt;
    using Stream::readBytes;
    using Stream::readBytesUntil;
 
  private:
    size_t write( uint8_t data ) { return (void)data, 0; }
};

MyStream stream;

void setup() {
  stream.read();  //Fine
  stream.find('a'); //Fine

  Stream &s = stream; //Compile error
  Print &p = stream; //Compile error
}

void loop() {}
```

Similar to the above, you can also have a Stream reference or pointer as a member variable and only expose the input functions.

eric-wieser:
> A library shouldn't be left up to assumptions, if you don't support print features, then make it well known

That's kinda the point of this patch. There is currently no way to make it known that you support read features but not print features, yet there is a way to express the inverse. Why should things be this way?

What if I want to write a function that can read from `Serial` and `stream`? That seems perfectly sane, but your latest response doesn't allow that without templates.

Look at this another way - if `IStream` did exist, can you think of a single reason it should be removed?

eric-wieser:
> `flush()` is part of Stream, but you can also move it to the private section.

Good catch, this probably doesn't belong in `IStream` as I have it here, since it deals with output and not input

matthijskooijman:
>  flush() is part of Stream, but you can also move it to the private section.

Also see #2387.

eric-wieser:
@matthijskooijman: Was tempted to fix that too. What's stopping that being merged?

matthijskooijman:
@eric-wieser, Nothing that I can see, apart from maybe lack of developer time.

eric-wieser:
Now that #2397 is merged, I've rebased this to be even simpler

hardware/arduino/avr/cores/arduino/IStream.h
```diff
@@ -0,0 +1,128 @@
...
+// or a time out occurs due to lack of input.
+enum LookaheadMode{
+    SKIP_ALL,       // All invalid characters are ignored.
+    SKIP_NONE,      // Nothing is skipped, and the stream is not touched unless the first waiting character is valid.
+    SKIP_WHITESPACE // Only tabs, spaces, line feeds & carriage returns are skipped.
+};
+
+#define NO_IGNORE_CHAR  '\x01' // a char not found in a valid ASCII numeric field
+
+class IStream
```
eric-wieser: This is the only line that actually changed in this file other than renaming `Stream` to `IStream` - it was previously `class Stream : public Print`

hardware/arduino/avr/cores/arduino/Stream.h
```diff
@@ -1,6 +1,5 @@
 /*
-  Stream.h - base class for character-based streams.
-  Copyright (c) 2010 David A. Mellis.  All right reserved.
```
eric-wieser: Nothing is left of the original file apart from the name `Stream`, so this copyright notice has moved to `IStream.h`

Issue #5619, opened 2016-11-22
GPG signatures for source validation


NicoHood:
As we all know, today more than ever before, it is crucial to be able to trust
our computing environments. One of the main difficulties that package
maintainers of Linux distributions face, is the difficulty to verify the
authenticity and the integrity of the source code.

The Arch Linux team would appreciate it if you would provide us GPG signatures
in order to verify easily and quickly your source code releases.

**Overview of the required tasks:**
* Create and/or use a 4096-bit RSA keypair for the file signing.
* Keep your key secret, use a strong unique passphrase for the key.
* Upload the public key to a key server and publish the full fingerprint.
* Sign every new git commit and tag.
* Create signed compressed (xz --best) release archives
* Upload a strong message digest (sha512) of the archive
* Configure https for your download server

[GPGit](https://github.com/NicoHood/gpgit) is meant to bring GPG to the masses.
It is not only a shell script that automates the process of creating new signed
git releases with GPG but also comes with this step-by-step readme guide for
learning how to use GPG.

**Additional Information:**
* https://github.com/NicoHood/gpgit
* https://help.github.com/categories/gpg/
* https://wiki.archlinux.org/index.php/GnuPG
* https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work
* https://www.qubes-os.org/doc/verifying-signatures/
* https://lkml.org/lkml/2016/8/15/445
* https://developers.google.com/web/fundamentals/security/encrypt-in-transit/why-https
* https://www.enigmail.net/index.php/en/

Thanks in advance.

NicoHood:
@facchinm @cmaglie Can you please sign the new 1.8 release and the arduino-builder release please? It would really help us to provide secure packages of the arduino software and everyone else can securely check their arduino download.

cmaglie:
We are setting up the keys for the release of 1.8.2

NicoHood:
So much love <3
(Please use 4k RSA)

cmaglie:
The git tag is now signed https://github.com/arduino/Arduino/releases/tag/1.8.2

About the signing of the single packages, instead of generating a signature for every package, what about signing the sha512sum.txt? This way we can sign all the packages at once, something like:

https://downloads.arduino.cc/arduino-1.8.2.sha512sum.txt.asc

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

[...]

PR #5772, opened 2016-12-28
Changed ant build script to use http with launcher4j.

Chris--A:
Due to sourceforge using a 307 temporary redirect, the build script now fails to retrieve the launcher package (windows/linux).
This is because the build script uses the https protocol and the redirect uses http.

Error from ant script:
> [get] Redirection detected from https to http. Protocol switch unsafe, not allowed.

This is a simple fix as it now requests the download using http, so the redirect is allowed.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -954,7 +954,7 @@
   <target name="download-launch4j-windows">
     <antcall target="unzip-with-ant-task">
       <param name="archive_file" value="windows/launch4j-3.9-win32.zip"/>
-      <param name="archive_url" value="https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip"/>
+      <param name="archive_url" value="http://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip"/>
       <param name="final_folder" value="windows/launcher/launch4j"/>
       <param name="dest_folder" value="windows/launcher/"/>
     </antcall>
@@ -963,7 +963,7 @@
   <target name="download-launch4j-linux">
     <antcall target="untar">
       <param name="archive_file" value="windows/launch4j-3.9-linux.tgz"/>
-      <param name="archive_url" value="https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz"/>
+      <param name="archive_url" value="http://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz"/>
       <param name="final_folder" value="windows/launcher/launch4j"/>
       <param name="dest_folder" value="windows/launcher/"/>
     </antcall>


Chris--A:
Just found this commit which changed all downloads to https: https://github.com/arduino/Arduino/commit/1da91087e410bf4b427906661e4feb43e291e342

@NicoHood

NicoHood:
The build always worked for me. Why does this happen? We should not use http as a solution.

cmaglie:
This is my output:

```
download-launch4j-linux:

untar-unzip-check:

untar-unzip-download:

untar-unzip-download-web:
      [get] Getting: https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz
      [get] To: /home/megabug/Code/arduino/build/windows/launch4j-3.9-linux.tgz
      [get] https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz moved to https://netcologne.dl.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ....................................................
      [get] ..............

untar-unzip-download-local:
```

so it seems that sourceforge doesn't redirect to http.
@Chris--A 
can you paste the error message?

Chris--A:
I haven't actually tried the linux build, only windows. However the linux package is also redirected (307).

> The build always worked for me. Why does this happen? We should not use http as a solution.

I didn't notice it until I re-cloned my repo from scratch (previously downloaded launcher deleted).

Here is the output:
```download-launch4j-windows:

untar-unzip-check:

untar-unzip-download:

untar-unzip-download-web:
      [get] Getting: https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip
      [get] To: U:\Arduino\build\windows\launch4j-3.9-win32.zip
      [get] https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip moved to http://internode.dl.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip
      [get] Redirection detected from https to http. Protocol switch unsafe, not allowed.

untar-unzip-download-local:

untar-unzip-checksum:
     [echo] Testing checksum of "windows/launch4j-3.9-win32.zip"
 [checksum] Could not find file U:\Arduino\build\windows\launch4j-3.9-win32.zip to generate checksum for.

BUILD FAILED
U:\Arduino\build\build.xml:103: The following error occurred while executing this line:
U:\Arduino\build\build.xml:1047: The following error occurred while executing this line:
U:\Arduino\build\build.xml:955: The following error occurred while executing this line:
U:\Arduino\build\build.xml:858: Could not find file U:\Arduino\build\windows\launch4j-3.9-win32.zip to generate checksum for.
```

Changing to http allows the redirect to occur.

NicoHood:
We better write sourceforge an email to provide https connections instead of using http.

cmaglie:
It seems that **some** of their mirrors supports only http.

Internode mirror force http:
```
      [get] https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip moved to http://internode.dl.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-win32.zip
      [get] Redirection detected from https to http. Protocol switch unsafe, not allowed.
```

Netcologne mirror insteal allows https:
```
      [get] https://downloads.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz moved to https://netcologne.dl.sourceforge.net/project/launch4j/launch4j-3/3.9/launch4j-3.9-linux.tgz
```

> We better write sourceforge an email to provide https connections instead of using http.

Are you going to write them?
In the meantime what do you suggest?

NicoHood:
@cmaglie I suggest some protest. Such a big site should take their business serious. I can write them, but you as arduino.cc might have some more impact. But I will try. In the meantime you can generate your gpg key #5619 .

It is really important, I cannot say it often enough. Please sign your sources. You do not need to use it yourself in the build script if you dont care, but please give others the chance to do so.

### And once more as an explanation:
The download via https does not secure us from source tampering. The initial download can be tampered. The hash we generate from this initial download is trusted by us, but does not guarantee the origin of the source. Therefore we need to use GPG.

The real solution here would be that launch4j signs their sources and we verify them. Then you can create a hash of that download and include it in your build script. But DO NOT blindly download sources, generate a hash and use this.

The use of https over http helps us a bit in the first place of the download. It ensures the server is hopefully the server. While still a lot people do not trust the CA model. If you can trust the hash (it has to be sha256 at a minimum), then you can also download via http or ftp. The reason why we do not want to use http still is a reason of privacy. Which might not be that critical in europe (yet) but in other countries it is.

So:
* I go ahead and write sourceforge about the https mirrors
* I go ahead and write launch4j about gpg signatures
* You go ahead and sign all your sources. Please sign github releases and the arduino.cc downloads
* Please also create sha512sums and also use them in build.xml if possible (over sha256)

cmaglie:
> The real solution here would be that launch4j signs their sources and we verify them. Then you can create a hash of that download and include it in your build script. But DO NOT blindly download sources, generate a hash and use this.

I still don't understand how can this improve things.
How a user is supposed to verify the signature? Who guarantees that the key I get from a keyserver is the real key used by developers and not a fake one?

Chris--A:
> It is really important, I cannot say it often enough. Please sign your sources. You do not need to use it yourself in the build script if you dont care, but please give others the chance to do so.

Are you asking me to sign my commits? (probably not).
Either way, I'm setting up gnupg anyway, might as well sign 'em. I'll amend this commit later.

PaulStoffregen:
Is this a solution in search of a problem?

In an alternate universe where everyone has an abundance of man-hours to implement everything, great.

But here in reality, where projects are constrained by a fixed number of developer hours, is this really worthwhile?  Should the human time needed to implement even more than the already-good hashes, and continue doing this extra work on every release really be prioritize over hundreds of other goals, which create tangible benefits for users?

Has there ever been any confirmed case of source tampering in Arduino's 10+ year history?

cmaglie:
While we wait for sourceforge to take action, I've mirrored the files to downloads.arduino.cc that have no problems with https. The files are already checksummed.

Chris--A:
Good compromise.

ChristophBWeber:
Hello all, I am Director of Technology at SourceForge. Great discussion here regarding all the facets that make a download safe and trustworthy. At SourceForge we are in the final stage of moving all our mirrors to HTTPS. This is a bit like herding cats, not unlike managing an open source project. Nevertheless, we will be 100% HTTPS on our mirror network by January 15, 2017. This is part of our ongoing push to move our entire platform to the HTTPS protocol. My apologies if you got caught in the middle of the transition unprepared.

Hope this clears things up, and please let me know if there's anything else we can do to help your project.
Christoph Weber,  christoph.weber@bizx.info

NicoHood:
> I still don't understand how can this improve things. How a user is supposed to verify the signature? 

@cmaglie A user can verify the signatures via GPG. But most users are to lazy to verify those. But we use them for packaging our software (not only arch, also debian (+ubuntu), rpm (suse, redhat) and others). It is the only way for us to provide secure packages of the software that we distribute because the author can be authenticated.

> Who guarantees that the key I get from a keyserver is the real key used by developers and not a fake one?

@cmaglie You need to write the [**full** fingerprint](https://lkml.org/lkml/2016/8/15/445) on your website and also on github. You have two sources with https where the fingerprint is shown which should give you a good portion of trust. Also GPG is a public trust database. Users can sign other keys, so they get more trust. For example my key is signed with the ArchLinux master keys. So you can ensure that its me, nicohood. Otherwise they would not have signed my public key/fingerprint.

> Are you asking me to sign my commits? (probably not).
Either way, I'm setting up gnupg anyway, might as well sign 'em. I'll amend this commit later.

@Chris--A The more people sign their commits, the more authenticated the git history becomes. The more people join, the better it is. And its really simple to activate automated commit signing with git.

> In an alternate universe where everyone has an abundance of man-hours to implement everything, regarless of its actual value, great.

@PaulStoffregen You totally missed the point. GPG signing does not even take a minute for every release that happens every month or so. But the value we can gain is authenticated sources. Its the reason why you can trust your system. Security is not something that we do for fun. And yes i think especially the last 3 years showed that is is extremely important to protect the users with all the surveillance and hacks out there. It would not be the first service that got hacked. And if we have GPG signatures this would prevent this risk.

@ChristophBWeber Thanks a lot for this statement, this really clarified things and helps the community. Thanks!

ChristophBWeber:
Just to tie this up: All active SourceForge mirrors are running over HTTPS as of 11:30am PST today. No more HTTP downloads. Hope this helps you folks.

PR #5774, opened 2016-12-29
Const correct eeprom

eric-wieser:
Add `const` keywords to methods where appropriate. This makes the following possible:

```c++

void do_readonly(const EEPROMClass& eeprom) {
    const EERef ref = eeprom[1]; // before this patch, this would error!
    T x = eeprom.get<T>(2); // before this patch, this would error!
    uint8_t val = eeprom.read(2);  // before this patch, this would error!

    //intentionally errors: ref++;

    bool empty = eeprom.begin() == eeprom.end();  // before this patch, this would error!
}
```

A minor but hopefully uncontentious change.

No changes to `EEPtr`, since that would require introducing a new type to mean "pointer to const" (`const EEPtr` just means "const pointer")

### DIFF
--- a/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
+++ b/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
@@ -93,8 +93,8 @@ struct EEPtr{
     EEPtr &operator=( int in )          { return index = in, *this; }
     
     //Iterator functionality.
-    bool operator!=( const EEPtr &ptr ) { return index != ptr.index; }
-    EERef operator*()                   { return index; }
+    bool operator!=( const EEPtr &ptr ) const { return index != ptr.index; }
+    EERef operator*()                         { return index; }
     
     /** Prefix & Postfix increment/decrement **/
     EEPtr& operator++()                 { return ++index, *this; }
--- a/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
+++ b/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
@@ -116,10 +116,11 @@ struct EEPtr{
 struct EEPROMClass{
 
     //Basic user access methods.
-    EERef operator[]( const int idx )    { return idx; }
-    uint8_t read( int idx )              { return EERef( idx ); }
-    void write( int idx, uint8_t val )   { (EERef( idx )) = val; }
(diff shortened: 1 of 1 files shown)


facchinm:
Adding @Chris--A to the thread since he surely knows any (possible) side effect of this patch (being the developer of EEPROM v2 :smile: )

Chris--A:
Thanks, @facchinm I have had a look through the PR.

I could not re-produce any errors with the current library (2.0). I did get appropriate warnings however.

This usage should be allowed, even though its quite *unique*. Seeing as there is a global `EEPROM` variable, passing a const reference to `EEPROMClass` is not a common usage.

However, I built the library with the intention that `EEPtr` and `EERef` can be used if someone wants specific control of the EEPROM.

As it stands, the change should not have any impact. Adding the `const` modifier to the functions really does nothing here as `EEPROMClass` has no data-members, which are the only things the `const` modifier protects.

----

I have been working on version 2.1, which is built off code I released in the [EEPROM thread on the Arduino forum ](https://forum.arduino.cc/index.php?topic=312645.0). It includes some new features I was testing such as bit access, String handling, and better iteration facilities. The code I released has been download 1700+ times and I've had quite a bit of feedback. The number one feature people are wanting in a release is bit access.

@eric-wieser 
Could I ask that you add your changes to: https://github.com/Chris--A/Arduino/tree/EEPROM_2.1 which is the branch with the next revision. I do like your proposal and as there are no side effects that I can see (will do testing) I think it would be worth adding a const interface. I'm happy to add these changes in myself if you'd prefer. The functions you have modified has changed slightly in my updates.

I did a little bit of messing with what you are proposing and, I'm not sure as I wasn't paying too much attention, but I think I did see a tiny decrease in code size. I've yet to work out what test actually caused it though, I will look into it further.

----

The latest code is almost ready to create a PR, I have some more documentation to complete and I have a few examples I need to clean up and add.

Cheers

eric-wieser:
@Chris--A : Done.

Another general eeprom comment replace `operator int()` with `explicit operator int()`, since conversion from a pointer to int is probably not a common operation, and could easily cause confusion.

Chris--A:
Cheers, now its the weekend, I'll spend some time going through the changes you have proposed.

facchinm:
@Chris--A   do you have the chance to test it again :slightly_smiling_face: 
@eric-wieser we are closing all the hardware related PRs in this repo.
Thanks!

Issue #5771, opened 2016-12-28
Unify AVR and SAM cores/arduino


cousteaulecommandant:
Right now hardware/arduino/avr/cores/arduino and hardware/arduino/sam/cores/arduino contain redundant files, for example Print.cpp, WString.cpp, etc.  This makes maintenance complicated, since changes on one won't affect the other (even when those changes are platform-independent), so the developer will have to make the changes on multiple files (and something tells me SAM won't have the same attention as AVR).  I think it would be a good idea to unify them so that all platform-independent sources are common, and then each derivative has a small set of platform-specific sources.

dlabun:
If you look at Print.cpp on both the AVR and SAM cores you'll see they're different... That's not to say there are not identical files shared between the cores but many of the files are tweaked for their specific core.

cousteaulecommandant:
True, but the differences are minimal.  In fact I only found one (that's 1 more than I initially assumed).  Still, I think those small differences should be separated as much as possible so that as many files as possible are identical and can be merged.  (Maybe this suggestion is not as immediate as I would have imagined)

Also, this separation between high-level functionality and low-level details seems logical from an organization point of view.

q2dg:
> something tells me SAM won't have the same attention as AVR

Well, until recent .cc/.org merge, SAM architecture (present only in Due board) was deprecated...

eric-wieser:
> If you look at Print.cpp on both the AVR and SAM cores you'll see they're different

Then put the one different function in `PrintFlashString.cpp`. C++ does not require all of a class to be in a single file.

mbanzi:
This is something we're investigating, how to move hw independent files in a different location than the hw dependent files. We'll bring the discussion to the dev list

@q2dg it was never deprecated as a Core, the Arduino Due as a product was marked as retired because we were not going to maintain it further.

bperrybap:
Now that the IDE is smarter about the way it handles libraries and can automatically locate sub libraries, it should be possible to make Print  its own library and move it to the common bundled libraries area.
The Print class should be core independent.
My guess is that the only reason that certain header files like Print were duplicated in each core was because until recently the IDE was not smart enough to locate sub libraries and the only way to ensure that files like Print.h, Printable.h, Stream.h and a few others were always located was to put them in the core libraries directory. This was because the core libraries directory was always on the include path.
Now that the IDE is smarter, these files could be moved into the bundled libraries area as separate libraries.

eric-wieser:
[...]

PR #5775, opened 2016-12-29
Make avr and sam files more similar to aid #5771

eric-wieser:
Pulls the per-platform code in `Print.cpp` into its own file, so that both `Print.cpp`s are now identical.

Also corrects style divergence between otherwise-identical files. Hopefully this helps with #5771


cmaglie:
the SAM core actually has a `avr/pgmspace.h` placeholder, I think we can just copy the Print.cpp over as-is from AVR (on the SAMD core we already do that).

sandeepmistry:
Closing as the two files should be the same now.

PR #5777, opened 2016-12-29
Added support for decimal RPMs.

ledjohnny:
I've slightly modified the Stepper motor library to add support for fractions of RPMs. I changed the `whatSpeed` variable to be of type `float`.


facchinm:
Hi @ledjohnny , could you move the PR to https://github.com/arduino-libraries/Stepper so we can keep track in the proper repo? Thanks!

PR #5786, opened 2017-01-01
Fix data loss when adding an existing file to sketch

polsys:
The `addFile` function checks for the case where source and destination are equal, for example when the user wants to update a sketch file modified outside the editor. However, the casing logic will delete the 'old' destination file before that, which causes the file to be lost.

### Steps to reproduce (tested on Windows)
1. Create a new sketch.
2. Create a new file (code or data, doesn't matter) outside the editor. Add it to the sketch by dragging and dropping.
3. Navigate to the sketch directory and drop the copied file again to the editor.
4. A replacement confirmation dialog appears. Select OK.
5. (If the file was a data file, a message box about this being a no-op appears.)
6. The file is deleted on the disk. An exception is logged if the file was a code file.

### The changes in this PR:
* Coalesced the file equality checks so that no file operations are performed if source and destination are same.
* Removed the data file no-op message. I didn't touch the localization files, being unfamiliar with them, I guess it should be removed there as well?

### DIFF
--- a/app/src/processing/app/SketchController.java
+++ b/app/src/processing/app/SketchController.java
@@ -495,61 +495,64 @@ public class SketchController {
       isData = true;
     }
 
-    // check whether this file already exists
-    if (destFile.exists()) {
-      Object[] options = { tr("OK"), tr("Cancel") };
-      String prompt = I18n.format(tr("Replace the existing version of {0}?"), filename);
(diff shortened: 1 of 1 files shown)


cmaglie:
Thank you!

There is no problem with localization strings, they are updated automatically with `gettext` (we run this tool periodically to cleanup/update translations).

PR #5790, opened 2017-01-02
Add virtual keyword to virtual method override, for clarity

eric-wieser:
These are already treated as virtual by the compiler anyway. Ideally we'd have
the override keyword here too, but I'm not sure if the lcd compiler supports
this.


facchinm:
Moved to https://github.com/arduino/ArduinoCore-sam/pull/5

eric-wieser:
What's the link between these two repos? Is one subtree merged into the other?

facchinm:
The sam folder was a leftover from AVR-only times. We are moving it in a way like https://github.com/arduino/ArduinoCore-samd, so the core is provided standalone with all the history extracted with `git filter-branch`.
When all PRs and issue will be moved to the right location we'll apply https://github.com/arduino/Arduino/pull/5881

PR #5796, opened 2017-01-03
Add virtual beginMulticast(...) stub to UDP class

sandeepmistry:
This allows libraries to use `udp.beginMulti(..)` when a `UDP` type is passed in. This method is optional and will fail by default if the network library does not support multicast UDP.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Udp.h
+++ b/hardware/arduino/avr/cores/arduino/Udp.h
@@ -41,7 +41,8 @@
 class UDP : public Stream {
 
 public:
-  virtual uint8_t begin(uint16_t) =0;	// initialize, start listening on specified port. Returns 1 if successful, 0 if there are no sockets available to use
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Do some protocols require also specifying a mac address?

sandeepmistry:
@PaulStoffregen good question, so far I've only looked at mDNS. I'll do some more research at other protocols that use multicast UDP.

sandeepmistry:
The local IP would be handy.

I've been experimenting making the [EthernetBonjour](https://github.com/TrippyLighting/EthernetBonjour) compatible with [WiFi101](https://github.com/arduino-libraries/WiFi101). For now, I changed `begin` to pass in the local IP.

Progress can be found on the following fork + branch: https://github.com/sandeepmistry/EthernetBonjour/tree/generic-udp

facchinm:
Partially moved to https://github.com/arduino/ArduinoCore-sam/pull/6

facchinm:
Merged manually 26ae0ac86814be9d949b8912e4c01c05ed19be32

PR #5802, opened 2017-01-04
[AVR] Send USB ZLP if required

### LINKED ISSUE #5732: Revert Subtract one from USB_EP_SIZE in USB_SendSpace
https://github.com/arduino/Arduino/pull/4864 introduced a bug that rawhid and the gamepad is not working anymore. This was reported by a lot of users now and should be reverted. There are lots of others suggestions to fix this bug, please go another way. This is definitely not the correct way and breaks multiple USB modules. The core should not try to fix CDC module bugs.

More details in the pr above and here: https://github.com/NicoHood/HID/issues/91

facchinm:
Fixes #5732 , "reverts" #4864

Tested with all the following problematic code on Linux, WinXp and Win10

https://github.com/NicoHood/HID/blob/master/examples/RawHID/RawHID/RawHID.ino (only on Linux)

```arduino
void setup() {

}

void loop() {
  if (Serial) {
    if (Serial.available() > 0) {
      Serial.read();
      Serial.write(buff,64);
      Serial.flush();
    }
  }
}
```

```arduino
void setup() {

}

void loop() {
  if (Serial) {
    if (Serial.available() > 0) {
      Serial.read();
      char buff[129] = "D_ENDPOINT(USB_ENDPOINT_IN (CDC_ENDPOINT_IN ),USB_ENDPOINT_TYPED_ENDPOINT(USB_ENDPOINT_IN (CDC_ENDPOINT_IN ),USB_ENDPOINT_TYPE";
      Serial.write(buff,128);
      Serial.flush();
    }
  }
}
```

```arduino
void setup() {

}

void loop() {
  if (Serial) {
    if (Serial.read() == 'h'){
      Serial.write("Hello ");
      Serial.flush();
      Serial.write("World!\r\n");
      Serial.flush();
    }
  }
}
```

Comments are really appreciated (as the USB core is still really convoluted, I'm probably missing something important)

@NicoHood @sandeepmistry @matthijskooijman @embmicro

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -256,9 +256,7 @@ u8 USB_SendSpace(u8 ep)
 	LockEP lock(ep);
 	if (!ReadWriteAllowed())
 		return 0;
-	// subtract 1 from the EP size to never send a full packet,
-	// this avoids dealing with ZLP's in USB_Send
-	return USB_EP_SIZE - 1 - FifoByteCount();
+	return USB_EP_SIZE - FifoByteCount();
 }
 
 //	Blocking Send of data to an endpoint
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -273,7 +273,9 @@ int USB_Send(u8 ep, const void* d, int len)
 	int r = len;
 	const u8* data = (const u8*)d;
 	u8 timeout = 250;		// 250ms timeout on send? TODO
-	while (len)
+	bool sendZlp = false;
+
+	while (len || sendZlp)
 	{
 		u8 n = USB_SendSpace(ep);
 		if (n == 0)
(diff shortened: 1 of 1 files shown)


hardware/arduino/avr/cores/arduino/USBCore.cpp
```diff
@@ -273,12 +271,15 @@ int USB_Send(u8 ep, const void* d, int len)
 	}
 
 	int r = len;
+
+	bool sendZlp = (len % USB_EP_SIZE) == 0;
```
cmaglie: I think that this is the wrong way to handle ZLP since the buffer could be filled in multiple rounds, for example this sketch still fails after the patch:

```C++
void setup() { }
void loop() {
  if (Serial) {
    if (Serial.available() > 0) {
      Serial.read();
      char buff[65] = "<--3456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABC-->";
      Serial.write(buff,32);
      Serial.write(buff+32,32);
      Serial.flush();
    }
  }
}
```

PR #5804, opened 2017-01-04
[WIP] Make serial monitor hyperlink-aware

facchinm:
Add hyperlink listener capabilities to the serial monitor.

The links are automatically created if http/https/ftp pattern is found, but a full `<a href ... </a>` style can be used and the output won't be mangled (useful if you want to make clickable links with loooong URLs).

No corner case is considered, cpu usage could ramp to 1000% or similar, so use it with a grain of salt (and report any unexpected behaviour/bug)


cmaglie:
I remember that the serial monitor has been carefully tuned for performance in previous pull-request: https://github.com/arduino/Arduino/pull/2491

Before merging we should check if those changes brings back all the poor behaviours that were solved on #2491 in particular the CPU usage when the serial monitor is flooded with data.

sandeepmistry:
@facchinm @cmaglie do we still want to include this in a future IDE release?

facchinm:
I think it **might** still be valuable for some Serial monitor based IoT tutorials, maybe with a better implementation :slightly_smiling_face:

PR #5813, opened 2017-01-06
Initial delay support in stepper motor library

Buskebam:
When a stepper motor is turned on it is possible the rotor is not lined up with the magnetic field. In these first few moments of initializing the motor will line up again. This process takes some time it is important to wait for this effect to be over to start an acceleration of the motor. Otherwise you can get unexpected behavior of the stepper. 

I had a lot of problems with this effect. I edited this library to have a starting delay after initializing. After that it worked.

### DIFF
--- a/libraries/Stepper/src/Stepper.cpp
+++ b/libraries/Stepper/src/Stepper.cpp
@@ -179,7 +179,7 @@ void Stepper::setSpeed(long whatSpeed)
 /*
 * Sets initial delay in us to prevent unexpected behaviour in the start 
 */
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @Buskebam  , could you move the PR to https://github.com/arduino-libraries/Stepper so we can keep track in the proper repo? Thanks!

Buskebam:
Hello, 

Thank you for your respons, I have moved the PR to https://github.com/arduino-libraries/Stepper/pull/3.

PR #5815, opened 2017-01-06
Change double quotes to single quotes

roncapat:
Observe that "." is a string literal, and so is treated as the char '.' plus the null char '\0'. Single quotes reduces the necessary memory for this literal to only one char instead. 
A string literal as the one actually present may require the use of the method "write(const char *str)", so there could be also a performance overhead. 
Another reason to change quotes style is for consistency with line 235.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -250,7 +250,7 @@ size_t Print::printFloat(double number, uint8_t digits)
 
   // Print the decimal point, but only if there are digits beyond
   if (digits > 0) {
-    n += print("."); 
+    n += print('.'); 
   }
 
   // Extract digits from the remainder one at a time
--- a/hardware/arduino/sam/cores/arduino/Print.cpp
+++ b/hardware/arduino/sam/cores/arduino/Print.cpp
@@ -242,7 +242,7 @@ size_t Print::printFloat(double number, uint8_t digits)
 
   // Print the decimal point, but only if there are digits beyond
   if (digits > 0) {
-    n += print("."); 
+    n += print('.'); 
   }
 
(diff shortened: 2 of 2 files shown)


roncapat:
My commits now appear after @matthijskooijman's and @NicoHood's approvals because I signed-off them only this evening, and so the original hashes have changed. Sorry for the inconvenient. Feel free to double-check content, no variations were made from original commits.

NicoHood:
@roncapat What do you mean by signed off? They are not GPG signed or at least if they are you miss-configured github.

matthijskooijman:
@NicoHood, he added Signed-off-by headers.
@roncapat, for this repo we are not currently requiring or actively using signed-off-by headers, though they won't hurt of course.

facchinm:
First two patches looks really good, while I'd avoid merging the GSM related patch (maybe opening another PR on https://github.com/arduino-libraries/GSM is more appropriate)

roncapat:
Let me know if I have to squash the first two commits and pop-out the last
one in order to have only the patch(es) you indicated, or something else.

In the next hours I'll then submit the last patch to GSM library repo.



Il lun 9 gen 2017, 10:07 Martino Facchin <notifications@github.com> ha
scritto:

First two patches looks really good, while I'd avoid merging the GSM
related patch (maybe opening another PR on
https://github.com/arduino-libraries/GSM is more appropriate)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/arduino/Arduino/pull/5815#issuecomment-271236892>, or mute
the thread

facchinm:
Simply force pushing on the same branch after removing the GSM related patch is ok IMO

roncapat:
Done. Here are only the patches for core library.

facchinm:
Merged, thanks!

PR #5816, opened 2017-01-07
Refactor Print.cpp for consistency

roncapat:
- Uniformed parameters (str, num) across "print" and corresponding "println". 
- Reorder methods to better separate non-ln and ln versions.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -74,48 +74,41 @@ size_t Print::print(unsigned char b, int base)
   return print((unsigned long) b, base);
 }
 
-size_t Print::print(int n, int base)
+size_t Print::print(int num, int base)
 {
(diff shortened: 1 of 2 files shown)


matthijskooijman:
This looks like a sane cleanup to me.

This is related to (mostly in the sense that it touches the same code and will probably conflict) to #5829.

roncapat:
Ok, I'll wait for the #5829 to merge and then fix the patch in this PR

cousteaulecommandant:
#5829 actually removes println from Print.cpp (except for println(void)), so that weird order issue is already handled by that.  The random arg names still persist though; it is true that they're a bit confusing.

roncapat:
I had a look to your PR (#5829), and I think it's a great cleanup. When it's merged, maybe I'll check for arg names refactor. I subscribed to your PR so I'll know when to rebase and start from updated sources.

roncapat:
Closing for obsolescence

PR #5817, opened 2017-01-07
Add support for alternative IPv4 formats

roncapat:
This patch satisfies the TODO that was present in the modified sources,
which requested support for "a", "a.b" and "a.b.c" shortened ipv4 addresses.

This patch zero-fills the remaining bytes of the address, if they are left
untouched by the string-to-number conversion routine (the main while loop of
the method).

### DIFF
--- a/hardware/arduino/avr/cores/arduino/IPAddress.cpp
+++ b/hardware/arduino/avr/cores/arduino/IPAddress.cpp
@@ -45,8 +45,6 @@ IPAddress::IPAddress(const uint8_t *address)
 
 bool IPAddress::fromString(const char *address)
 {
-    // TODO: add support for "a", "a.b", "a.b.c" formats
-
     uint16_t acc = 0; // Accumulator
     uint8_t dots = 0;
 
@@ -76,12 +74,15 @@ bool IPAddress::fromString(const char *address)
             return false;
         }
     }
+    
+    _address.bytes[dots++] = acc;
 
-    if (dots != 3) {
-        // Too few dots (there must be 3 dots)
-        return false;
+    // Fill remaining bytes with zeroes (needed to support a, a.b, a.b.c addresses)
+    while (dots <= 3)
+    {
+       _address.bytes[dots++] = 0; 
     }
-    _address.bytes[3] = acc;
+    
     return true;
 }
 
--- a/hardware/arduino/sam/cores/arduino/IPAddress.cpp
+++ b/hardware/arduino/sam/cores/arduino/IPAddress.cpp
@@ -45,8 +45,6 @@ IPAddress::IPAddress(const uint8_t *address)
 
 bool IPAddress::fromString(const char *address)
 {
-    // TODO: add support for "a", "a.b", "a.b.c" formats
-
     uint16_t acc = 0; // Accumulator
     uint8_t dots = 0;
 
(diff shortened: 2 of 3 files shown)


cmaglie:
Thanks for you contribution, BTW uhis implementation is not correct, the extra bytes **should not** be 0-filled.

https://en.wikipedia.org/wiki/Dot-decimal_notation

> A popular implementation of IP networking, originating in 4.2BSD, contains a function inet_aton() for converting IP addresses in character string representation to internal binary storage. In addition to the basic four-decimals format and full 32-bit addresses, it also supported intermediate syntax forms of octet.24bits (e.g. 10.1234567; for Class A addresses) and octet.octet.16bits (e.g. 172.16.12345; for Class B addresses). It also allowed the numbers to be written in hexadecimal and octal, by prefixing them with 0x and 0, respectively. These features continue to be supported by software until today, even though they are seen as non-standard.[3] But this also means addresses where an IP address component is written with a leading zero digit may be interpreted differently by different programs: some will ignore the leading zero, some will interpret the number as octal.[4]

If I read correctly the following **non-standard** format **should** be supported:

(8.24) 10.123456 -> 10.1.226.64
(8.8.16) 10.10.12345 -> 10.10.48.57
(8.24 octal) 10.0123456 -> 10.0.167.46
(8.8.16 ocatl) 10.10.012345 -> 10.10.20.229
(8.24 hex) 10.0x123456 -> 10.18.52.86
(8.8.16 hex) 10.10.0x1234 -> 10.10.18.52

moreover if an address is not valid it should be solved with a DNS lookup, for example:
"10.10.0x12345" should be solved with a dns query because 0x12345 exceeds the expected 16 bits in the 8.8.16 format.

To be honest, I never seen those really used, I guess we can live without this overly complex and convoluted implementation.

roncapat:
Honestly, I searched around a bit because the interpretation I gave to "a", "a.b" and "a.b.c" was a nonsense for me. Maybe next time I should ask for explanation before committing. Now, all seems a bit more clear. 

Thank you for the link (I learned something new, 👍), and sorry for the bloat code.

You think that notation is not useful. So why not, at least, cleaning that TODO strings? Or you think that a correct implementation would be nice to see in the future (even if with some dubts regarding its usefulness)?

cmaglie:
> You think that notation is not useful. So why not, at least, cleaning that TODO strings? Or you think that a correct implementation would be nice to see in the future (even if with some dubts regarding its usefulness)?

At the beginning I thought it was something nice to have at some point, but after some time and after reading the document linked above (that says it's nonstandard and probably not working consistently on every OS) I changed my mind so... yeah... it's time to remove the TODO. Would you like to submit another PR?

roncapat:
OK, I'll do that in the next minutes.

oqibidipo:
If you ever change your mind again, here is a working version.
Leading zeros are ignored, no octal/hex, empty parts are not allowed.

```C++
bool IPAddress::fromString(const char *address)
{
    uint32_t acc = 0; // Accumulator
    uint8_t dots = 0;
    uint8_t digits = 0;
    
    while (*address)
    {
        char c = *address++;
        if (c >= '0' && c <= '9')
        {
            if (acc > UINT32_MAX/10 || (acc == UINT32_MAX/10 && c > UINT32_MAX%10 + '0'))
            {
                // Adding the next digit would overflow
                return false;
            }
            acc = acc * 10 + (c - '0');
            digits++;
        }
        else if (c == '.')
        {
            if (digits == 0) {
                // Empty part
                return false;
            }
            if (dots == 3) {
                // Too many dots, max 3 allowed
                return false;
            }
            if (acc > 255) {
                // Only the last part can be > 255
                return false;
            }
            _address.bytes[dots++] = acc;
            digits=0;
            acc = 0;
        }
        else
        {
            // Invalid char
            return false;
        }
    }

    // Check the validity of the last part
    if (digits == 0 ||
        (dots == 3 && acc > 0xFF) ||
        (dots == 2 && acc > 0xFFFF) ||
        (dots == 1 && acc > 0xFFFFFF))
    {
        return false;
    }

    for (int_fast8_t d = 3; d >= dots; d--) {
        _address.bytes[d] = acc;
        acc >>= 8;
    }
    return true;
}
```

PR #5831, opened 2017-01-10
Arduino due member function in attachInterrupt callback

straccio:
In arduino due using std::function for a callback
for example:
`std::function<void()> int_func =std::bind(&button::press,this);
attachInterrupt(digitalPinToInterrupt(pin),int_func,FALLING);`


facchinm:
Moved to arduino/ArduinoCore-sam#8

PR #5877, opened 2017-01-19
Master

bpzl:


matthijskooijman:
This doesn't seem like a meaningful PR, what's the point of it?

bpzl:
I am sorry about that. New to arduino and coding. I pressed wrong button, was hoping it wouldn't do something like this. 
I'll be more careful until I get the hang of things. 

Sonny

matthijskooijman:
No problem, thanks for clarifying. I'm closing this, then.

PR #5878, opened 2017-01-19
Fix command line with relative paths

### LINKED ISSUE #5836: --verify with relative filename does not work
It seems that specifying a relative filename to the sketch does not work with `--verify`. I suspect this worked before and was broken in the recent refactor to remove GUI dependencies from the CLI invocation.

To reproduce, just call arduino with a plain filename (it does not matter if the file/sketch exists or not), like:

```
$ arduino --verify foo.ino
Picked up JAVA_TOOL_OPTIONS:
Loading configuration...
Initializing packages...
Preparing boards...
Verifying...
java.lang.NullPointerException
        at processing.app.helpers.FileUtils.listFiles(FileUtils.java:348)
        at processing.app.Sketch.listSketchFiles(Sketch.java:108)
        at processing.app.Sketch.<init>(Sketch.java:54)
        at processing.app.Base.<init>(Base.java:381)
        at processing.app.Base.main(Base.java:140)
```

Note that the above has the following patch applied, since the `--verify` handling code currently eats up exceptions without showing them (which also makes things fail without error when a full path to a non-existent sketch is passed). This should also be fixed.

```
--- a/app/src/processing/app/Base.java
         outputFile = new Compiler(sketch).build(progress -> {}, false);
       } catch (Exception e) {

         // Error during build
-        System.exit(1);
+        throw(e);
       }
```

Furthermore, passing a full filename works:
```
$ arduino --verify $(pwd)/foo.ino
(output hidden)
```

Without `--verify`, a relative pathname works as expected and loads the sketch in the GUI without complaints:
```
$ arduino foo.ino
```

cmaglie:
- Added test for command line with relative paths (issue #5836)
- Fixed cmd line regression

This fix #5836

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -376,7 +376,7 @@ public class Base {
         // Build
         splash.splashText(tr("Verifying..."));
 
-        File sketchFile = new File(parser.getFilenames().get(0));
+        File sketchFile = BaseNoGui.absoluteFile(parser.getFilenames().get(0));
         sketch = new Sketch(sketchFile);
 
         outputFile = new Compiler(sketch).build(progress -> {}, false);
--- a/app/test/processing/app/CommandLineTest.java
+++ b/app/test/processing/app/CommandLineTest.java
@@ -1,0 +1,85 @@
+/*
+ * This file is part of Arduino.
+ *
+ * Copyright 2017 Arduino LLC (http://www.arduino.cc/)
+ *
+ * Arduino is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


matthijskooijman:
@cmaglie, are you also fixing the missing error handling?

cmaglie:
> are you also fixing the missing error handling?

Not yet, I didn't looked the internals yet, but from looking at the code that launch the build:

```java
      try {
          [............]
          outputFile = new Compiler(sketch).build(progress -> {}, false);
       } catch (Exception e) {
         // Error during build
         System.exit(1);
       }
```

it seems that an excpetion is raised also when a compile error occurs (that should not be an exception at all...) and in that case we should not print the exception/stacktrace.
I think that `Compiler.build(..)` needs to be refactored to not throw exceptions for build errors to properly fix this. I'll leave this change for now.

matthijskooijman:
> I think that Compiler.build(..) needs to be refactored to not throw exceptions for build errors to properly fix this.

Yeah, probably indeed. Perhaps showing just the exception message and not the stacktrace is a compromise for now? Then at least *something* is shown on errors, instead of just nothing when some internal exception happens?

cmaglie:
> Then at least something is shown on errors, instead of just nothing when some internal exception happens?

I'll cover this in another PR.

Issue #5890, opened 2017-01-20
AppStream upstream metadata fails appstreamcli validation


probonopd:
AppStream upstream metadata fails appstreamcli validation:

```
E - arduino.appdata.xml:<root>
    The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher).

Validation failed.
```

(cc'ing @ximion fyi)

cmaglie:
Should be solved with #5892

PR #5892, opened 2017-01-20
Make metainfo and .desktop files spec compliant

ximion:
This resolves bug #5890 

The AppStream/Metainfo spec has changed a bit, and the file shipped with Arduino doesn't work at all currently, since it was apparently written against a draft version of the initial spec which was never fully supported (congrats, looks like you were one of the earliest adopters!).
This patch updates the metainfo file to match the current version of the specification as found at https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

Additionally, I also updated the .desktop file to follow the reverse-DNS scheme. This is a new requirement of the desktop-entry specification for modern applications to be D-Bus activatable and work well on Wayland.
The AppStream spec highly recommends using that scheme too, so I adapted it to be used in the .desktop and metainfo files.
You can read more about this at https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html

If you have any questions, let me know!
You can validate the current metadata using `appstreamcli validate <metainfofile>`.

NOTE: I haven't tested if the script works yet (but it should).
If the .desktop filename change is not wanted, we can also go without it - just let me know and I'll adapt the patch.

### DIFF
--- a/build/linux/dist/appdata.xml
+++ b/build/linux/dist/appdata.xml
@@ -1,8 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- See https://wiki.gnome.org/GnomeGoals/AppDataGnomeSoftware -->
-<application>
-  <id type="desktop">arduino.desktop</id>
-  <licence>CC-BY-SA</licence>
+<!-- See https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html -->
+<component type="desktop-application">
+  <id>cc.arduino.arduinoide.desktop</id>
+  <metadata_license>CC-BY-SA-3.0</metadata_license>
+
+  <name>Arduino IDE</name>
+  <summary>Open-source electronics prototyping platform</summary>
+
   <description>
     <p>
       Arduino is an open-source electronics prototyping platform based
@@ -15,10 +19,13 @@
       to develop and upload code to compatible microcontrollers.
     </p>
   </description>
+
   <screenshots>
     <screenshot type="default" width="624" height="351">http://mavit.fedorapeople.org/appdata/arduino-screenshot.png</screenshot>
     <screenshot width="704" height="396">http://mavit.fedorapeople.org/appdata/arduino-photo.jpg</screenshot>
   </screenshots>
+
   <url type="homepage">http://www.arduino.cc/</url>
-  <updatecontact>arduino.appdata.xml@mavit.org.uk</updatecontact>
-</application>
+
+  <update_contact>arduino.appdata.xml@mavit.org.uk</update_contact>
+</component>
--- a/build/linux/dist/appdata.xml
+++ b/build/linux/dist/appdata.xml
@@ -3,6 +3,7 @@
 <component type="desktop-application">
   <id>cc.arduino.arduinoide.desktop</id>
   <metadata_license>CC-BY-SA-3.0</metadata_license>
+  <developer_name>Arduino LLC</developer_name>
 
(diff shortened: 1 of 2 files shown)


build/linux/dist/appdata.xml
```diff
@@ -15,10 +19,13 @@
...
   <screenshots>
     <screenshot type="default" width="624" height="351">http://mavit.fedorapeople.org/appdata/arduino-screenshot.png</screenshot>
     <screenshot width="704" height="396">http://mavit.fedorapeople.org/appdata/arduino-photo.jpg</screenshot>
   </screenshots>
+
   <url type="homepage">http://www.arduino.cc/</url>
-  <updatecontact>arduino.appdata.xml@mavit.org.uk</updatecontact>
-</application>
+
+  <update_contact>arduino.appdata.xml@mavit.org.uk</update_contact>
```
matthijskooijman: I wonder if we should update this to an arduino.cc address (developers mailing list perhaps)? @cmaglie?

build/linux/dist/install.sh
```diff
@@ -141,10 +144,18 @@ simple_uninstall_f() {
     rm "${HOME}/.local/share/applications/arduino.desktop"
   fi
 
+  if [ -f "${HOME}/.local/share/applications/arduino-arduinoide.desktop" ]; then
+    rm "${HOME}/.local/share/applications/arduino-arduinoide.desktop"
+  fi
+
```
matthijskooijman: This should have a comment stating this was the name used in older scripts (perhaps add an OLD_RESOURCE_NAME var?)

build/linux/dist/install.sh
```diff
@@ -6,7 +6,7 @@
 # If called with the "-u" option, it will undo the changes.
 
 # Resource name to use (including vendor prefix)
-RESOURCE_NAME=arduino-arduinoide
+RESOURCE_NAME=cc.arduino.arduinoide
```
matthijskooijman: Is the `.desktop` suffix intentionally missing here?

ximion: Yes, it gets added later in the script (so the ID can be used by multiple things)

build/linux/dist/appdata.xml
```diff
@@ -1,8 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- See https://wiki.gnome.org/GnomeGoals/AppDataGnomeSoftware -->
-<application>
-  <id type="desktop">arduino.desktop</id>
-  <licence>CC-BY-SA</licence>
+<!-- See https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html -->
+<component type="desktop-application">
+  <id>cc.arduino.arduinoide.desktop</id>
+  <metadata_license>FSFAP</metadata_license>
```
ximion: Yes, it is well-defined and recommended for these texts, as it does not require a large license text to be shipped with the metainfo file and because it is very short. See https://spdx.org/licenses/FSFAP.html

In any case, I did not intend to change the license, that was by accident - CC-BY-SA-3.0 is totally fine and I changed it back in the new revision.

ximion:
I updated the patch with all changes from @mavit included.

build/linux/dist/appdata.xml
mavit: I no-longer maintain the Fedora package of Arduino, so certainly wouldn't object to this.

facchinm:
Merged, thanks!

PR #5921, opened 2017-01-28
More useability for serial monitor / editor

MichaelSy:
added "clear output" button to serial monitor 
added "show timestamp" checkbox to serial monitor
added configurable shortcut for comment / uncomment
added setting to show always file extensions

### DIFF
--- a/app/src/processing/app/AbstractTextMonitor.java
+++ b/app/src/processing/app/AbstractTextMonitor.java
@@ -128,13 +128,14 @@ public abstract class AbstractTextMonitor extends AbstractMonitor {
 
     pane.add(autoscrollBox);
     pane.add(addTimeStampBox);
-    pane.add(clearButton);
     pane.add(Box.createHorizontalGlue());
     pane.add(noLineEndingAlert);
     pane.add(Box.createRigidArea(new Dimension(8, 0)));
     pane.add(lineEndings);
     pane.add(Box.createRigidArea(new Dimension(8, 0)));
     pane.add(serialRates);
+    pane.add(Box.createRigidArea(new Dimension(8, 0)));
+    pane.add(clearButton);
 
     mainPane.add(pane, BorderLayout.SOUTH);
   }
--- a/app/src/processing/app/AbstractTextMonitor.java
+++ b/app/src/processing/app/AbstractTextMonitor.java
@@ -8,6 +8,8 @@ import java.awt.Dimension;
 import java.awt.Font;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 import javax.swing.Box;
 import javax.swing.BoxLayout;
@@ -20,7 +22,9 @@ import javax.swing.JScrollPane;
 import javax.swing.JTextField;
 import javax.swing.SwingUtilities;
 import javax.swing.border.EmptyBorder;
+import javax.swing.text.BadLocationException;
 import javax.swing.text.DefaultCaret;
+import javax.swing.text.Document;
 
 import cc.arduino.packages.BoardPort;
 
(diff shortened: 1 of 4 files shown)


facchinm:
LGTM! Did you measure if there is any performance issue with the timestamp enabled, sing a sketch like this one
```arduino
void setup() {
SerialUSB.begin(9600);
while (!SerialUSB) ; // wait
delay(250);
}
unsigned long count=1;
void loop() {
SerialUSB.print("Printing simple text, count # ");
SerialUSB.println(count++);
}
```
on a board with native USB serial?

PaulStoffregen:
@ArduinoBot build this please

MichaelSy:
@facchinm, I tested it with a NodeMCU. All works fine without performance issues...

MichaelSy:
added a custom title setting:
editor.custom_title_format=Arduino {version} | {folder}\{file}

If the setting is not available, the default style will be used.

possible options:
{file} = filename
{folder} = absolute folder name
{path} = absolute folder name
{project} = project (sketch) name
{version} = IDE Version

facchinm:
@MichaelSy unfortunately nodeMCU uses an USB to serial converter so the speed is capped (instead, on boards with serial over USB, the speed is *much* faster) . I'll test your PR on a Zero, just wondering why @ArduinoBot is still asleep...

facchinm:
With the timestamp enabled, on the sample sketch I posted earlier with a Zero, the heap usage ramps to 900MB but then the garbage collector acts good. CPU usage increases from 2-3% to 20% but this is an hardcore test so overall it looks good from a performance POW.
I'm attaching a screenshot for @00alis so we can get some suggestions on UI elements placements 
2017-01-31-151940_863x338_scrot

00alis:
Hi,
All good UI wise for the timestamp! I think the option should be unchecked by default.

I am wondering about the 'Clear Output' feature, is it really needed? Closing the serial monitor will clear the output anyway.
If we feel that is needed the button will be better place on the bottom right corner, pushing the 2 dropdowns on the left, since now it is too attached to the Timestamp option and it feels somewhat related to it.

MichaelSy:
@00alis the default is unchecked. It will be saved in the preferences. I think the button don't disturb the user and its easier to clear the output then "closing window and reopen".  I will change the position to the right side.

PaulStoffregen:
I'm very concerned about this PR's performance.  Testing is really needed with a fast native USB board like Arduino Due or Teensy 3.x.  Arduino Micro & Zero are native USB, but they lack the optimized USB code and/or CPU performance to fully utilize even 12 Mbit/sec USB.

facchinm:
@PaulStoffregen I'll test again with a Due and report the results here :wink:

MichaelSy:
I moved the button from left to right. @00alis 
If the performance is a big issue. Maybe we can show the checkbox only for NodeMCU?
tmp

PaulStoffregen:
Anyone know what happened to @ArduinoBot ?

facchinm:
> Anyone know what happened to @ArduinoBot ?
We are investigating, it seems that github changed something so the jenkins plugin is unable to be triggered by @mentioning it. I'd prefer to avoid updating the plugin since the last time it replied to ~120 PRs but maybe it's the only way :smile:

facchinm:
@ArduinoBot build this please.

facchinm:
@ArduinoBot build this please.

FernandoGarcia:
Hi!

How about add a reset button to serial monitor to prevent open and close the window to reset the board?

Best regards.

cmaglie:
I've merged (cherry picked) into master the commits for:

- configurable shortcut for comment / uncomment
- setting to show always file extensions
- setting for custom title

(I've also slightly amended them to remove extra whitespaces)

I'm now rebasing the remaining changes to master and resolve conflicts.

Anyway, it's better to not have multiple changes in a single PR, please push them into many different branches/PR so we can review and merge each one separately, even if it seems cumbersome it definitely makes things easier.

cmaglie:
The last commit in this PR, that has been added later I presume, is:

```
CHANGED: use cache for editor tools (don't create new instance for each menu item)
ADDED: setting for autoformat before saving (editor.autoformat_currentfile_before_saving")
CHANGED: move "clear ouput" button from left to right side (Serial Monitor)
```

This kind of "cumulative patch" commits must be absolutely avoided because, without extra editing/amending, it forces to pick the patch altogether.
For example I'm not sure I understand the usefulness of the "use cache for editor tools", but this is another unrelated patch that would block the serial monitor enhancement.

BTW this is just a note for the future, now I've spitted the commit into three smaller commits in #6178 so let's continue the discussion there.

PaulStoffregen:
> I am wondering about the 'Clear Output' feature, is it really needed?

Github has a saying "anything added dilutes everything else".

PR #5960, opened 2017-02-07
Update rsyntaxarea to 2.6.1

facchinm:
Fixes #5888

PR #5990, opened 2017-02-14
Update RSyntaxTextArea to 2.6.1 and remove need for Arduino-specific patches

matthijskooijman:
This applies some cleanups to the undo and copy-paste handling, that make things simpler and remove the need for modifying RSTA, allowing the upstream compiled jar files to be used. This PR can replace #5960.


facchinm:
Looks REALLY good to me. The cleanup is really nice, and I was somehow concerned about the translations/shortcuts but they all work well (on Linux and OSX at least).
I'll keep testing it for a couple of days, but it can easily supersede #5960.
@PaulStoffregen would you like to take a stress test "journey" also on this PR :smile: ?

app/src/processing/app/Editor.java
```diff
@@ -1265,43 +1260,58 @@ public void actionPerformed(ActionEvent e) {
...
     menu.setName("menuEdit");
     menu.setMnemonic(KeyEvent.VK_E);
 
-    undoItem = newJMenuItem(tr("Undo"), 'Z');
+    // Create a dummy RTextArea, to force it to create the static list
+    // of actions accessible through RTextArea.getAction().
+    new RTextArea();
+
+    RecordableTextAction undoAction = RTextArea.getAction(RTextArea.UNDO_ACTION);
+    JMenuItem undoItem = new JMenuItem(new RstaActionWrapper(undoAction));
```
cmaglie: The dummy object creation to obtain the `UNDO_ACTION` makes me really uneasy.
Also, we already have a `EditorTab.handleUndo()` method, why not just call that instead of trying to fit the RSTA action?

Moreover the RSTA action is wrapped inside another ActionWrapper class that is made just to route the message to the correct EditorTab through getCurrentTab(), while it may be much more simple to do something like:

```java
undoItem.addActionListener(() -> getCurrentTab().handleUndo());
```

Or maybe I'm missing something?

I remember that I've already solved this in another refactoring experiment I did some time ago I'll try to recover this attempt from my old branches.

app/src/processing/app/Editor.java
```diff
@@ -1610,8 +1526,6 @@ public int getCurrentTabIndex() {
    */
   public void selectTab(final int index) {
     currentTabIndex = index;
-    undoAction.updateUndoState();
-    redoAction.updateRedoState();
```
cmaglie: Removing these calls produce the correct enable/disable of menu actions?
Something to check if this PR is merged.

app/src/processing/app/Editor.java
matthijskooijman: The reason to use RSTA's actions, is that they are automatically updated to become enabled/disabled when the undo state changes, and to automatically show "Undo" or "Can't undo". If we do that manually, as is done currently, we need access to the UndoManager which needs a patch (at least for the `getUndoPresentationName()`, IIRC RSTA does export a `canUndo()` or similar). Also, we need to keep track of changes within the document with change listeners and other stuff, which is a bit messy. Having said that, this approach isn't 100% elegant either, so I can understand your being hesitant.

app/src/processing/app/Editor.java
matthijskooijman: Yeah, RSTA handles this internally already.

app/src/processing/app/Editor.java
cmaglie: thanks for the clarification, inedeed the fact that RSTA handle automatically the enable/disable is a point.

Another thing that comes to mind is how the localization is handled? is it something already in the RSTA resources? about `getUndoPresentationName()` it seems that the menu is always presented as "Undo" or "Redo", IMHO it's something that can be easily dropped in case.

cmaglie:
I've already pushed just this commit into master:
87c917f (Remove unused CaretAwareUndoableEdit class)

SasmitoMotion:
Thanks mr Maglie, it would be help for me and my friend

cmaglie:
fixed by #6022

PR #6023, opened 2017-02-27
Add input and output function

OtacilioN:
input and output functions can make Arduino Language easier for beginners to learn, being more clean, they will able to choose if they want to use pinMode or input/output function

also basic examples and the button example have been updated to use input and output functions

### DIFF
--- a/build/shared/examples/01.Basics/Blink/Blink.ino
+++ b/build/shared/examples/01.Basics/Blink/Blink.ino
@@ -18,13 +18,16 @@
   
   modified 8 Sep 2016
   by Colby Newman
+  
+  modified 27 Feb 2017
+  by Otacilio Neto
 */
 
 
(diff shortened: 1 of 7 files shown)


Ivan-Perez:
In order not to duplicate code, shouldn't it be better to call `pinMode` from `input` and `output` functions?

```cpp
void output(uint8_t pin)
{
    pinMode(pin, OUTPUT);
}
```

facchinm:
Hi @OtacilioN ,
we are not willing to modify/add basic IO APIs unless it's really necessary. `pinMode` works really well as it is, and providing another function with the same effect (wrapper) has no pros. If you find that wrappers useful in your sketches, you can create a library in no time and use it in every sketch (or, you can take a look at something like https://pin.fenichelar.com/ )

PR #6041, opened 2017-03-05
Fix "save as" operation for the data folder of a sketch

### LINKED ISSUE #5998: Version 1.8.1 doesn't copy 'data' folder on save as
When using the Arduino IDE with an ESP8266, for example, I use the data folder in the sketch folder (the folder that opens when you hit CTRL+K) for files to upload to the SPIFFS. (HTML, CSS, JS, icons etc.) In version 1.6.x, this data folder was copied when doing a 'save as', but using the new 1.8.1 version, I have to copy the folder manually.
Why was this feature removed?

I'm using Ubuntu 16.04.

Pieter

jeroenoverman:
This pull fixes #5998.

When performing a save as operation the data folder of the sketch was not copied to the new location.

What has changed:
- switching to the new folder is now the last operation in the save as method.
  - this caused not finding a data folder, since the folder was already changed to the new location
- If the data folder does not exists yet, it is created before the files are copied to the new location

I have tested this on my Windows machine and it solves the issue.

### DIFF
--- a/arduino-core/src/processing/app/Sketch.java
+++ b/arduino-core/src/processing/app/Sketch.java
@@ -351,13 +351,22 @@ public class Sketch {
         file.saveAs(new File(newFolder, file.getFileName()));
     }
 
-    folder = newFolder;
 
     // Copy the data folder (this may take a while.. add progress bar?)
     if (getDataFolder().exists()) {
       File newDataFolder = new File(newFolder, "data");
+      // Check if data folder exits, if not try to create the data folder
+      if (!newDataFolder.exists() && !newDataFolder.mkdirs()) {
+        String msg = I18n.format(tr("Could not create directory \"{0}\""), newFolder.getAbsolutePath());
+        throw new IOException(msg);
+      }
+      // Copy the data files into the new folder
       FileUtils.copy(getDataFolder(), newDataFolder);
     }
+    
+    // Change folder to the new folder
+    folder = newFolder;
+    
   }
 
   /**


arduino-core/src/processing/app/Sketch.java
```diff
@@ -351,13 +351,22 @@ public void saveAs(File newFolder) throws IOException {
         file.saveAs(new File(newFolder, file.getFileName()));
     }
 
-    folder = newFolder;
 
     // Copy the data folder (this may take a while.. add progress bar?)
     if (getDataFolder().exists()) {
       File newDataFolder = new File(newFolder, "data");
+      // Create the data folder
+      if (!newDataFolder.mkdirs()) {
```
facchinm: Should we also check for `newDataFolder.exists()` before bailing out? The check could be probably turned into `if(!(newDataFolder.exists() || newDataFolder.mkdirs()))` . Do you see any drawback in this approach?

jeroenoverman: We can check if the folder exists, but since we just created the parent folder at line 341, it is very unlikely the folder already exits.

Never the less it won't harm if we implement this extra check. So I will add the check and test if it still works.

matthijskooijman:
Changes look good to me. The first and last commit should probably be dropped, no point in polluting the history with them. As for the second commit (moving the `folder` update), it would probably be good to expand the commit message a bit to point out that `folder` is an instance variable used by `getDataFolder()` and that the old code would never see the old data folder for that reason?

arduino-core/src/processing/app/Sketch.java
```diff
@@ -355,8 +355,8 @@ public void saveAs(File newFolder) throws IOException {
     // Copy the data folder (this may take a while.. add progress bar?)
     if (getDataFolder().exists()) {
       File newDataFolder = new File(newFolder, "data");
-      // Create the data folder
-      if (!newDataFolder.mkdirs()) {
+      // Check if data folder exits, if not try to create the data folder
+      if (!(newDataFolder.exists() || newDataFolder.mkdirs())) {
```
matthijskooijman: I would think that `if (!newDataFolder.exists() && !newDataFolder.mkdirs())` is a bit easier to read, would you agree?

jeroenoverman: For readability you might even want to separate the condition into two if statements. (In case you don't know how the lazy and/or operator works).

But I think it is better to keep them into one if statement. I will change it to the and operator for better readability.

jeroenoverman:
I have fixed the comments you gave me.

jeroenoverman:
@facchinm  Is this Pull Request ready to be merged or should something still be changed or more tested?

facchinm:
Looks fine, thanks for the ping :wink:

Issue #5827, opened 2017-01-09
Package an up-to-date version for Debian


JulienPalard:
Look like current deb package has not been updated in years, leading to users use a deprecated version of Arduino, and reporting very old bugs like https://github.com/arduino/Arduino/issues/5822.

Looks like the deb upgrade is blocked by licence clarity: https://github.com/arduino/Arduino/pull/2703

NicoHood:
If you split the packages properly, the licenses are clear to me. Look at the archlinux packages and propose your debian package suggestions. I dont think its a license issue, just nobody wants to package the stuff.

maqifrnswa:
this new folder layout (post 1.8) is clear. Packaging it is a pain, but it was already done. I think that debian can be updated.

cmaglie:
@JulienPalard @maqifrnswa 
do you know who is actually the maintainer of the debian package?

You should probably ask him to update the debian package and not here.

NicoHood:
@cmaglie Please sign the sources #5619 so we can safely package arduino.

cmaglie:
> Please sign the sources

@NicoHood thanks for the reminder, there is already #5619 to track the issue no need to duplicate here ;-)

PanderMusubi:
Please, see also https://bugs.launchpad.net/ubuntu/+source/arduino/+bug/1425677

PR #6053, opened 2017-03-10
adding arduino-linux-setup.sh script

artynet:
The shell script helps the linux user to quickly configure the OS environment to work with the Arduino IDE. It mainly performs these tasks :

* writes **UDEV** rules for accessing the serial ports
* uninstalls **modemmanager** from the system if installed (preventing the use of the serial ports)
* add the user to the **tty** and **dialout** groups

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -629,6 +629,7 @@
 
     <copy todir="linux/work" file="linux/dist/arduino" />
     <copy todir="linux/work" file="linux/dist/install.sh" />
+    <copy todir="linux/work" file="linux/dist/arduino-linux-setup.sh" />
     <copy todir="linux/work" file="linux/dist/uninstall.sh" />
 
     <chmod perm="ugo+x">
--- a/build/linux/dist/arduino-linux-setup.sh
+++ b/build/linux/dist/arduino-linux-setup.sh
@@ -1,0 +1,194 @@
+# arduino-linux-setup.sh : A simple Arduino setup script for Linux systems
+# Copyright (C) 2015 Arduino Srl
+#
+# Author : Arturo Rinaldi
+# E-mail : arturo@arduino.org
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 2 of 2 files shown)


NicoHood:
You should rather provide a proper debian package for debian/ubuntu users and not install the files directly inside the filesystem. This was never a good idea. #5827

rkost:
Hmm. I don't know if it was a good idea to automatically uninstall any packages on the system. This is what dependencies and conflicting packages are for. This does not seem useful to me. Maybe you should print some advice that if modemmanager is installed, the user should take care of it.
But I think the way to go was described by @NicoHood already.

NicoHood:
Why was this merged? As I said software should not automate any installation or even uninstall software. A proper distribution package should be used instead. Otherwise people with less knowledge will mess up their system and be unable to undo all the changes. I think it was not a wise idea to follow this "concept".

ArduinoCC itself should think of providing debian packages via PPA, as its the most often used distribution at the moment (debian/ubuntu) for end consumer. ArchLinux already has proper packages.

facchinm:
@NicoHood this has been merged since it condenses a lot of FAQs into a "simple" bash script that the user MAY launch if needed. The script is not being launched automatically during installation so it doesn't hurt anyone :smile: 
About DEB/PPA, I'm sorry I have to repeat it again: it's not something we are willing to provide directly. You did an awesome job with the Arch packages, if someone wants to prepare a deb recipe/a PPA we can endorse it but nothing more.

NicoHood:
Those bash commands have to be put into a readme with instruction details and reasons. Users will just try to run it, possibly break their system, dont know what they did and stop. Just write one good doc, but dont automate this. Its evil. But still your opinion.

sk8nfool:
Actually there are a number of things wrong with arduino-linux-setup.sh!
It is actually dangerous. 

1. It shouldn't remove ANY package. It should warn that there may be a problem and suggest how to mitigate it.

2. It very nicely creates a bunch of udev rule files in /tmp but doesn't meet the rule definition - the filename for avrisp.rules is wrong! It is lacking the number at the beginning (should be something like 99-avrisp.rules)

3. It copies ANY file ending in ".rules" from /tmp to /etc/udev/rules.d. WRONG - it should only copy the files it created. Using "*.rules" is a bit too permissive.

4. There is no need to reboot the system. Kicking udev (as done excessively in refreshudev()) is sufficient.

Lastly - there is no documentation regarding arduino-linux-setup.sh! Nothing that tells the user whether or not to run the scrip and if so when. Personally I don't think this file should be part of the distribution at all.

facchinm:
@sk8nfool you are right in all your points.
Now we only need to understand if leaving the file there (and fixing its shortcomings) is worth the effort or if we'd better remove it entirely.
Poll time :slightly_smiling_face: ?

sk8nfool:
It kind of depends on how it's fixed. If the final result has benefit then
keep it and document it. If not toss it.

Issue #6033, opened 2017-03-02
Flaw in changes to save/restore bootloader magic key RAM


MLXXXp:
The changes added to save and restore the data at the bootloader "magic key" location in RAM are flawed. These changes were made in a [commit on Apr 6, 2016](https://github.com/arduino/Arduino/commit/4c901d3f9f347177b3f263c451885ddc3b1b6ab0) and discussed in issue #2474.

The code is located in file _CDC.cpp_, which as of this writing is [here](https://github.com/arduino/Arduino/blob/1f35bfca531312bc9bae49ad596c57bd3eae9c38/hardware/arduino/avr/cores/arduino/CDC.cpp).

The problem is that data can be restored to the magic key location even if it hasn't been saved in the first place, possibly overwriting data that the sketch has put there. This is because received USB information can cause the _else_ at line 132 to execute before the _if_ statement has ever become true.

The following sketch demonstrates the problem, when compiled under IDE 1.8.1 with core version 1.6.17, for a Leonardo or other ATMega32u4 based board. _overlapArray[]_ has been set to sit over the magic key location at RAM address 0x800. If the IDE serial monitor is opened, the printed data will show that the array data has been corrupted with the value that was at RAMEND - 1.

```cpp
// Bootloader magic key overlap test

// allocate some global variable space so that...
uint8_t globalArray[1698];
// this array will overlap the bootloader magic key
uint8_t overlapArray[8];

void setup() {
  Serial.begin(9600);
  // fill the array to make sure it's not optimised out
  for (uint16_t i = 0; i < sizeof(globalArray); i++) {
    globalArray[i] = i; 
  }
  // fill the overlap array with F0 F1 F2 F3 F4 F5 F6 F7 F8
  for (uint8_t i = 0; i < sizeof(overlapArray); i++) {
    overlapArray[i] = i + 0xF0;
  }
  delay(5000);
}

void loop() {
  Serial.print("\noverlapArray =");
  for (uint8_t i = 0; i < sizeof(overlapArray); i++) {
    Serial.print(" 0x");
    Serial.print(overlapArray[i], HEX);
  }
  Serial.print("\n*(RAMEND-1) = 0x");
  Serial.println(*(uint16_t *) (RAMEND-1), HEX);
  delay(2000);
}
```

I've come up with a proposal to address the problem:

1. Only inintiate the boot sequence (signaled by baudrate = 1200 and DTR off) if DTR was previously on.

[...]

PR #6055, opened 2017-03-10
[AVR] Fix problems with USB CDC auto-reset

MLXXXp:
Address a problem where the bootloader magic key location could be restored before it was saved, and other issues, with a bootloader auto-reset invoked by USB CDC.

Discussed in issue #6033


facchinm:
@ArduinoBot build this please

az0uz:
I like the Idea of having the possibility to use Watchdog with 32u4 based arduino.

This patch doesn't work if you call `wdt_reset()` in the user code faster than one time every `> 120ms` (reset time in the CDC driver).
If you do so,  the mcu will never reset to let you upload a new sketch.

MLXXXp:
@az0uz 
> This patch doesn't work if you call wdt_reset() in the user code faster than one time every > 120ms

This is being discussed in PR #6064.

facchinm:
Closing as merged in https://github.com/arduino/ArduinoCore-avr

PR #6060, opened 2017-03-11
Fix path to preferences.txt on Windows

terrillmoore:
The documentation was wrong for Windows, for locating the preferences file. This update matches modern Windows (7..10), but is probably not right for XP.

### DIFF
--- a/build/shared/manpage.adoc
+++ b/build/shared/manpage.adoc
@@ -212,7 +212,8 @@ EXIT STATUS
 
 FILES
 -----
-*~/.arduino15/preferences.txt*::
+*~/AppData/Local/Arduino15/preferences.txt (Windows)::
+*~/.arduino15/preferences.txt* (Mac OS X and Linux)::
 	This file stores the preferences used for the IDE, building and
 	uploading sketches.
 


terrillmoore:
Sorry, I see that I missed a trailing '*' in the markdown, which messes up formatting. I'll resubmit.

PaulStoffregen:
Actually, on Macintosh the path is ~/Library/Arduino15/preferences.txt

PaulStoffregen:
And on Windows, the unix "~" syntax doesn't really apply.  The correct syntax is: %LocalAppData%/Arduino15/preferences.txt

terrillmoore:
Well... actually, if you're using git bash on WIndows, which practically everybody does these days, ~/AppData .... is the easiest way to get to the file.

If not using Git Bash, then we should replace `/` by `\` for all Windows paths, because `/` is sometimes rejected by overzealous parsers.

matthijskooijman:
@terrillmoore, for future reference, you could have force-pushed to your original branch and have this PR updated instead of creating a new one.

PaulStoffregen:
> using git bash on WIndows, which practically everybody does these days

citation needed

terrillmoore:
> @terrillmoore, for future reference, you could have force-pushed to your original branch and have this PR updated instead of creating a new one.

Thanks, @matthijskooijman I'll bear that in mind.

PR #6081, opened 2017-03-16
During compilation, scroll the console to the first error

matthijskooijman:
While adding messages to the console, this keeps track of the position
of the first bit of stderr output. Then, whenever
`EditorConsole.scrollDown()` is called, the scroll position is changed
so that the line above the first error line is the first one in the
console. If no errors are printed yet, the console scrolls down as
before.

To know what messages are coming from stderr, the ConsoleOutputStream
that generated the message is passed to `EditorConsole.insertString()`,
though this might not be the best solution for this.

This PR is still a bit of proof-of-concept, especially the extra argument to `insertString()` doesn't sound like the perfect solution yet. Perhaps the console code should be refactored to be a bit simpler in the first place. But I needed this myself (it will be saving me a lot of time in my current work), so might as well put it up for discussion :-)

### DIFF
--- a/app/src/cc/arduino/ConsoleOutputStream.java
+++ b/app/src/cc/arduino/ConsoleOutputStream.java
@@ -95,7 +95,7 @@ public class ConsoleOutputStream extends ByteArrayOutputStream {
     if (editorConsole != null) {
       SwingUtilities.invokeLater(() -> {
         try {
-          editorConsole.insertString(text, attributes);
+          editorConsole.insertString(this, text, attributes);
         } catch (BadLocationException ble) {
           //ignore
         }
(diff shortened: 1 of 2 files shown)


facchinm:
The intention looks very good, but it seems that the console doesn't scroll anymore if there is no error but only warnings. This probably happens because everything from stderr is interpreted as en error (I have lots of warnings about broken libs etc at the very beginning of the compilation output)

matthijskooijman:
Yeah, that's a tricky thing. Currently, the error regex matching happens in a completely separate part of the code, which I couldn't easily connect to the scrolling code. Also, I'm not sure if skipping warnings is really wanted, especial compiler warnings are usually relevant.

matthijskooijman:
This also interacts badly with avrdude uploading, since that outputs all of its output on stderr, so when compilation is successful, it sticks around the top of the avrdude output without scrolling on to the progress bar. To properly implement this, I think arduino-builder should output more structured info segmenting the stdout and stderr output per command run, and communicating the first failed command, so the IDE can scroll to the output of that failed command, or something like that.

On additional problem is that if you manually scroll down, the current code continues to scroll back up until compilation is complete.

PR #6086, opened 2017-03-17
Disable serial monitor for non-SSH network ports

sandeepmistry:
This avoids displaying the password prompt for network ports that don't support SSH.


sandeepmistry:
@cmaglie this is ready for review.

Issue #5849, opened 2017-01-14
Add "Clear" button to Serial Monitor UI.


rlaferla:
Many times, your Arduino is spitting out a ton of data and you want to manually clear the monitor.  The only way currently (AFAIK) is to restart the app.  I'd like to see a "Clear" button on the GUI that when pressed clears the monitor.

I looked at the code and this could be easily added to SerialMonitor and AbstractTextMonitor with some code similar to below:

```
protected JButton clearButton;

protected void onCreateWindow(Container mainPane) {
  // after adding send button
  clearButton = new JButton(tr("Clear"));
  upperPane.add(clearButton);
}

  public void onClearCommand(ActionListener listener) {
    clearButton.addActionListener(listener);
  }

protected void onEnableWindow(boolean enable) {
  // other code
    clearButton.setEnabled(enable);
}

// and in Serial Monitor.java

    onClearCommand(new ActionListener() {
          public void actionPerformed(ActionEvent e) {
              textArea.setText("");
          }
    });

```

lpasqualis:
+1

agdl:
@rlaferla you are right there is no button, anyway if you click again on the baud rate the Serial monitor is cleared (faster than close/open again the monitor). Anyway on boards like the UNO this means a board reset

cmaglie:
Fixed by https://github.com/arduino/Arduino/commit/99006fc5c68115a55b01b796a50d6a904c9b716c

Thank you!

PR #6092, opened 2017-03-18
Issue #5849: Added clear button to serial monitor UI

rlaferla:
https://github.com/arduino/Arduino/issues/5849


FernandoGarcia:
@facchinm It's duplicated from https://github.com/arduino/Arduino/pull/5921?

facchinm:
@FernandoGarcia , correct, it's indeed a duplicate. I added both on the next milestone so we can compare them and avoid losing some code

cmaglie:
Fixed by https://github.com/arduino/Arduino/commit/99006fc5c68115a55b01b796a50d6a904c9b716c that is almost the same implementation, excecpt for the button that is in the right lower corner.

PR #6124, opened 2017-03-26
Use usbasp-clone protocol for programmer "usbasp".

jbtronics:
Avrdude won't check for usb name and vendor of the programmer, so even non-original usbasp clones (but with correct usb IDs) can be used for uploading. "Real" usbasp programmers are not affected by this change.
All in all more programmers are supported and problems with usbasp-clones are prevented.


facchinm:
Hi @jbtronics  ,
Thanks!

Issue #6128, opened 2017-03-27
editor.indent setting in 'preferences.txt' has NO effect in 1.8.1


bombasticbob:
Because it is SO irritating (like it also was for THIS guy:  http://forum.arduino.cc/index.php?topic=247973.0 ), I want to SHUT OFF any "help" the IDE *FEELS* I need when typing things.

This PARTICULARLY includes ANY form of "autocomplete".

I type SO fast, I don't NEED it.  In fact, my 'finger memory' assumes it WILL NOT BE THERE.

So, when it _IS_ there, it's an EXTREME IRRITANT.

Well, I tried the solutions offered in the above forum link, that is to set 'editor.indent' to FALSE.

I did this in the 'preferences.txt' in '.arduino15' and ALSO the one in the 'lib' directory [where I'd modified the colors to be SANE so I could actually READ the errors and warnings in the build output] so I'm pretty sure I was adjusting the CORRECT FILES.

And the result?  NOTHING!  I _STILL_ get that EXTREMELY IRRITATING BEHAVIOR.

So, PLEASE FIX the ability to SHUT IT OFF.  Apparently, it USED to work (shutting it off I mean).  And please don't say "use a different editor".  That's not the point.  The point is to FIX THIS ONE.

facchinm:
Fixed with #6129. Next time, please avoid using caps lock and this harsh tone or the issue will be closed automatically, no matter how much right you are. If you dislike the IDE so much, there are thousands of alternatives out there, just pick one and enjoy it (or, if you like our IDE in the end, try being more polite please :wink: )

bombasticbob:
thanks for getting a test-build out there quickly.

bombasticbob:
the test build in #6129 appears to fix the general problem; that is, the auto-indent behavior is now controlled by the editor.indent true/false assignment.  _However_, this does not address my reason for having created this issue in the _first_ place.  As such, I shall create a _new_ issue describing the nature of the problem in full, and recommend the addition of a new setting to control it.  Editing 'preferences.txt' is acceptable to me, which _should_ make this a minimal change.  I may even do it myself with a pull request, if I have time.

bombasticbob:
related to this issue, I found the code that inserts the ending curly brace, see latest comment added to #6132 (with recommended change)

PR #6129, opened 2017-03-27
Apply editor.indent preference

facchinm:
Solves #6128

### DIFF
--- a/app/src/processing/app/EditorTab.java
+++ b/app/src/processing/app/EditorTab.java
@@ -148,7 +148,7 @@ public class EditorTab extends JPanel implements SketchFile.TextStorage {
     textArea.setMarginLineEnabled(false);
     textArea.setCodeFoldingEnabled(PreferencesData.getBoolean("editor.code_folding"));
     textArea.setAutoIndentEnabled(PreferencesData.getBoolean("editor.indent"));
-    textArea.setCloseCurlyBraces(PreferencesData.getBoolean("editor.autocomplete"));
+    textArea.setCloseCurlyBraces(PreferencesData.getBoolean("editor.auto_close_braces", true));
     textArea.setAntiAliasingEnabled(PreferencesData.getBoolean("editor.antialias"));
     textArea.setTabsEmulated(PreferencesData.getBoolean("editor.tabs.expand"));
     textArea.setTabSize(PreferencesData.getInteger("editor.tabs.size"));
--- a/app/src/processing/app/EditorTab.java
+++ b/app/src/processing/app/EditorTab.java
@@ -148,6 +148,7 @@ public class EditorTab extends JPanel implements SketchFile.TextStorage {
     textArea.setMarginLineEnabled(false);
     textArea.setCodeFoldingEnabled(PreferencesData.getBoolean("editor.code_folding"));
     textArea.setAutoIndentEnabled(PreferencesData.getBoolean("editor.indent"));
+    textArea.setCloseCurlyBraces(PreferencesData.getBoolean("editor.autocomplete"));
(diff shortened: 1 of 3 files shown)


bombasticbob:
tested the linux64 version.  it now honors the true/false state of the 'editor.indent' setting for indenting.  however, the _unwanted_ behavior of adding a '}' for me when I type '{' is still there.

Thank you for fixing this part of the issue.  I have noted that the '{' '}' "autocomplete" behavior is no longer directly associated with indenting. It happens whether or not indenting is enabled.

I shall add a new issue to use a 'preferences.txt' string to _*disable*_ the unwanted (irritating) behavior, and reference these issues as part of it.

bombasticbob:
related to this issue, I found the code that inserts the ending curly brace, see latest comment added to #6132 (with recommended change)

app/src/processing/app/EditorTab.java
```diff
@@ -148,6 +148,7 @@ private SketchTextArea createTextArea(RSyntaxDocument document)
     textArea.setMarginLineEnabled(false);
     textArea.setCodeFoldingEnabled(PreferencesData.getBoolean("editor.code_folding"));
     textArea.setAutoIndentEnabled(PreferencesData.getBoolean("editor.indent"));
+    textArea.setCloseCurlyBraces(PreferencesData.getBoolean("editor.indent"));
```
matthijskooijman: Wouldn't this be a separate preference? I personally dislike autocompletion, but I do want (same-level) auto indent (i.e. pressing enter preserves the indentation of the previous line, it should not increase indentation after a { for me).

facchinm: Agree, it should be moved to its own preference. I added it this way to check if the functionality was ok and provide a build with it

bombasticbob:
@facchinm
@matthijskooijman

Thanks for the activity on this.  Yes, I'm in total agreement with your earlier comments about making these two functionalities separate.  Since the forum post from a couple of years ago (mentioned at the beginning) suggested that getting rid of auto-indent also got rid of auto-closing-brace, I had to weigh getting rid of something I want (auto-indent) against getting rid of something I _really_ do _not_ want (auto-closing-brace).  Having the two settings instead (i.e. #6132) would be the best overall fix for this.

So thanks for the mention.  Also thanks for testing how it _can_ be done.

bombasticbob:
Here's a patch file that I have created based on a 'git clone' I did on the repository.  It includes the change that fixes 'indent' (noted above) as well as adding 'editor.autocomplete' (issue #6132).  I tested it and it works perfectly.  It requires manually editing the local 'preferences.txt' to turn OFF the feature, but that's fine with me and probably everyone who wants this feature on by default.


[issue_6132.patch.txt](https://github.com/arduino/Arduino/files/896350/issue_6132.patch.txt)

bombasticbob:
@facchinm 

please incorporate #6155 as well.  then at least 3 additional people would be happier (and most likely _many_ others).

cmaglie:
@bombasticbob 
the option `editor.autocomplete`, as it is now, disables "auto close braces" but not "auto indent", while I expect it to override both (even if `editor.indent` is true).

Anyway, the editing-aids are just "auto indent" and "auto close braces" and IMHO is too early to think about an option to disable the aids altogether, I'd add such option later when/if more editing-aids are added.

So here my proposal:

- Change `editor.autocomplete` (that seems too generic to me) to `editor.auto_close_braces` and merge this PR as is
- Postpone the option to disable all editing-aids to a future PR when more editing-aids are added

bombasticbob:
@cmaglie 

changing the text of the setting to 'auto_close_braces' is ok with me.  I just want a way to control the behavior.  I'm happy that the only question is what actual name to use [something I had already guessed might happen].

PR #6147, opened 2017-04-03
Update to USB v2.1

bromagosa:
In order to stay compatible with the upcoming WebUSB API, as specified at https://github.com/webusb/arduino/blob/gh-pages/README.md#getting-started

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.h
+++ b/hardware/arduino/avr/cores/arduino/USBCore.h
@@ -127,7 +127,7 @@
 #define MSC_PROTOCOL_BULK_ONLY					0x50 
 
 #ifndef USB_VERSION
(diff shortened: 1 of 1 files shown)


facchinm:
There is no need to change it for everyone (which would bring a lot of side effects). Simply clone all the project in your sketchbook and restart the IDE. In this way, two boards will appear in the Boards menu, already declaring the correct USB version -
 https://github.com/webusb/arduino/blob/gh-pages/hardware/webusb/avr/boards.txt (the final paths must resemble `$sketchbook//hardware/webusb/avr/boards.txt` or the boards won't appear)

bromagosa:
@facchinm I see, thanks!

I guess the WebUSB README need updating then :)

PR #6176, opened 2017-04-10
Use version 0.9.1 of the WiFi101 firmware updater

sandeepmistry:
cc/ @agdl 

Let's wait for @ArduinoBot to make a new build ...


sandeepmistry:
@agdl this looks good to me, can you test it out and merge if all is good?

Issue #849, opened 2012-11-15
adding support of code completion [imported]


cmaglie:
This is [Issue 849](http://code.google.com/p/arduino/issues/detail?id=849) moved from a Google Code project.
Added by 2012-03-07T13:26:22.000Z by [Wilfried...@gmail.com](http://code.google.com/u/104914328976362983260/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Enhancement, Priority-Medium
### Original description

**What change would like to see?**
I would like to see a code completion in the IDE like other IDE Editors have. Especially it would be nice, when this feature could interact with the #include keyword. 

**Why?**
Because of simpler programming.

**Would this cause any incompatibilities with previous versions?  If so, how**
**can these be mitigated?**
I thing, there will be no incompatibilities.

ricardojlrufino:
As we talked on the list of developers, did the new implementation based on the structure in branche 1.5.x.
The current implementation status is:
 - Autocomplete (90%)
 - Real-time Parser and Error Marker(90%)
 - Documentation in Editor (in progress - 20%)

How was launched from the IDE version 1.6, I believe you may have a few more adjustments.
But must first of a feeedback than has already been done to trace a plan for the other implementations (such as documentation on Editor)

ricardojlrufino:
Was made the merge with the 1.6.1 version of the master and created a new branche to facilitate integration:
https://github.com/OpenDevice/Arduino/tree/1.6.x-autocomplete-neweditor

lars-sorensen:
So... Has this been cancelled?

PureSpider:
Would hate if this was cancelled. :-1:

ricardojlrufino:
I did not have any further feedback from the Arduino team. It seems to me that is not as priority

PureSpider:
Can you provide a windows build for the current version of the Arduino IDE with your extensions? I would love to use that!

Ede123:
Always the useful stuff is considered low priority :disappointed: 

Would be great if this could be implemented, since it would make coding a lot easier.

@arduino team: Please give some feedback on whether you're willing to merge this and what needs to be done if you think there are any issues! All the hard work was obviously done by @ricardojlrufino already...

lmihalkovic:
IMHO, the code as it is somwhow _forces_ a choice between cpp/c & completion OR multiple file types (like editing keywords.txt or the lib descriptor). There might also be athe need for a legal advice on distributing the cdt code inside arduino.

[...]

PR #6235, opened 2017-04-29
Base implementation of autocomplete #849

ricardojlrufino:
Start with the minimum base for autocomplete support.
The functionality is initialized with the key: Ctrl + Space
A test class has been created: FakeCompletionProvider, to enable usability testing on other platforms

Implementation can be added through the class extension:
Cc.arduino.autocomplete.CompletionProvider

### Tests:
[X] Ubuntu 16.04 64

### TODO:
[ ] Discovery of implementations (using preferences or using Java SPI ) ?


facchinm:
@ArduinoBot build this please

tigoe:
I find this confusing, and couldn't get it to autocomplete. When i press control-space (an unusual key combo for OSX users), I get a popup box that interrupts my typing, and when I dismiss it with the enter key, it puts the line number in my code. 

What I assumed when I saw "autocomplete" was something like what Atom (https://atom.io/) or Coda (https://panic.com/coda/) do: when the editor knows the context from the file type (e.g. .html, .js) it auto-completes from keywords when you tab or hit enter, there's no additional popup box, but there is a very small hovering text tip that doesn't block your view, which you can turn on or off.. It's not invasive, you can ignore it if you choose, or you can complete it with one keystroke. To me, that is what I would expect in an auto-complete. I don't see how that relates to what this is doing. Can you explain further what you're intent is?

cmaglie:
@tigoe 
this PR is aimed to do the **only** the GUI part first, so the popup is not populated with actual code-completions but with some placeholders. @ricardojlrufino made this after my request here #849, because his first submission of a full implementation of autocomplete (here https://github.com/arduino/Arduino/pull/3484) is way too complex to handle all at once and touch many aspects of the IDE that are unrelated with the autocompletion.

So this is a try to go with an incremental approach, once done the GUI part we can focus on the code-completion engine that will be the tricky part to get right.

tigoe:
I see, thank you for the explanation. In which case, my thoughts on the GUI are as follows: the popup window is too big and invasive. I would do something smaller, that takes less screen real estate. And the key combination is uncomfortable for OSX users. If it requires a command-key combo, I would use control as the modifier on windows and Linux, but command as the modifier on OSX, as is standard. Though ideally, I would not require a modifier except to turn the mode on or off; I'd use enter or tab. like other auto-corrects fo.

ricardojlrufino:
@cmaglie  My idea was to follow their guidelines, to have a cleaner interface.
But in fact, I think it's best to use what we already have ...

> Is this the autocomplete library here?
> https://github.com/bobbylight/AutoComplete

YES

@tigoe In that first moment, I find it interesting to focus on the functional aspects. The visual part and usability can have its specific moment.
These changes could be made in the bobbylight / AutoComplete project, I believe. I could try doing them (although I'm not an OSX user .. =])

tigoe:
OK. When you're dealing with the visual/usability, I'll pay attention again, because that is the most important aspect to me.

ricardojlrufino:
Now, basically, there have been two classes ...

SketchCompletionProvider, is the basis for providing autocomplete services for various contexts: Code, Strings, documentation.

Specifically the 'code' that is what interests us at that moment, will be offered by classses that extend: BaseCCompletionProvider

Currently the two classes do not have much, but they will be necessary as we implement new functionalities.

For reference, look at my previous code:

ricardojlrufino:
Hi, @cmaglie. Did you have time to do the review?

app/src/cc/arduino/autocomplete/BaseCCompletionProvider.java
```diff
@@ -0,0 +1,17 @@
+package cc.arduino.autocomplete;
+
+import org.fife.ui.autocomplete.DefaultCompletionProvider;
+
+/**
+ * Base completion provider for C/C++.
+ * @author Ricardo JL Rufino (ricardo@criativasoft.com.br)
```
cmaglie: May you add a GNU GPL license on top of all the files and move the credits there?
You can copy the header from any of the other files.

app/src/processing/app/EditorTab.java
```diff
@@ -106,6 +107,8 @@ public EditorTab(Editor editor, SketchFile file, String contents)
     file.setStorage(this);
     applyPreferences();
     add(scrollPane, BorderLayout.CENTER);
+    
+    textarea.setupAutoComplete(editor.getSketch(), new FakeCompletionProvider());
```
cmaglie: Ok some thoughts about this:

- this call could be commented out, so we can merge this PR without affecting the IDE behavior for now.

- what about adding a method like `Editor.setAutocompletionProvider(CompletionProvider)`? this way we can implement the completion-engine into a tool...

- moreover I think that this call can be inlined (and the method `textarea.setupAutoComplete` removed).

ricardojlrufino:
Hi @cmaglie , I followed your suggestions, what's the next step?

ricardojlrufino:
@cmaglie PING ! =)

facchinm:
The relevant bits of this PR has been merged in beta-1.9 branch, closing as merged

PR #6261, opened 2017-05-05
Fix move constructor and operator= in String.

### LINKED ISSUE #6260: String move constructor and move assignment don't move
The String::String(String&&) and String::operator=(String&&) do not perform expected operation i.e. they don't move, but instead they copy.

leg0:
The move constructor String::String(String&&) and String::operator=(String&&) now perform move instead of copy.
This fixes #6260.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/WString.cpp
+++ b/hardware/arduino/avr/cores/arduino/WString.cpp
@@ -43,16 +43,15 @@ String::String(const __FlashStringHelper *pstr)
 	*this = pstr;
 }
 
-#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__)
+#if STRING_HAS_MOVE
 String::String(String &&rval)
+	: buffer(rval.buffer)
+	, capacity(rval.capacity)
+	, len(rval.len)
 {
-	init();
-	move(rval);
-}
-String::String(StringSumHelper &&rval)
-{
-	init();
-	move(rval);
+	rval.buffer = NULL;
+	rval.capacity = 0;
+	rval.len = 0;
 }
 #endif
 
@@ -189,25 +188,22 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length)
 	return *this;
 }
 
-#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__)
+#if STRING_HAS_MOVE
 void String::move(String &rhs)
 {
-	if (buffer) {
-		if (rhs && capacity >= rhs.len) {
-			strcpy(buffer, rhs.buffer);
-			len = rhs.len;
-			rhs.len = 0;
-			return;
-		} else {
+	if (this != &rhs)
+	{
+		if (buffer != NULL)
 			free(buffer);
-		}
+
+		buffer = rhs.buffer;
+		len = rhs.len;
+		capacity = rhs.capacity;
+
+		rhs.buffer = NULL;
+		rhs.len = 0;
+		rhs.capacity = 0;
 	}
-	buffer = rhs.buffer;
-	capacity = rhs.capacity;
-	len = rhs.len;
-	rhs.buffer = NULL;
-	rhs.capacity = 0;
-	rhs.len = 0;
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


hardware/arduino/avr/cores/arduino/WString.cpp
```diff
@@ -43,16 +43,15 @@ String::String(const __FlashStringHelper *pstr)
...
-	init();
-	move(rval);
-}
-String::String(StringSumHelper &&rval)
-{
-	init();
-	move(rval);
+	rval.buffer = NULL;
+	rval.capacity = 0;
+	rval.len = 0;
```
eric-wieser: No need for these last two lines - all that matters is that `buffer` is NULL to prevent `rval::~String()` from freeing it

hardware/arduino/avr/cores/arduino/WString.cpp
```diff
@@ -43,16 +43,15 @@ String::String(const __FlashStringHelper *pstr)
 	*this = pstr;
 }
 
-#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__)
+#if STRING_HAS_MOVE
 String::String(String &&rval)
+	: buffer(rval.buffer)
+	, capacity(rval.capacity)
+	, len(rval.len)
```
eric-wieser: Really these should all have a call to `std::move`, but of course `std::move` is not available on the arduino.

Perhaps it would be sensible to provide `std::move` and `std::forward` if arduino is going to start using rvalue references internally

hardware/arduino/avr/cores/arduino/WString.cpp
leg0: According to c++ standard, object needs to be in valid state after being moved from. Having null buffer and non-zero capacity and /or length does not seem like a valid state to me.

hardware/arduino/avr/cores/arduino/WString.cpp
leg0: Even if std::move was available it would not make much sense to use it here - all it does is to cast is argument to rvalue reference. For integral types and pointers move construction is same as copy construction.

I agree that it would be useful to have move and forward available in lib.

hardware/arduino/avr/cores/arduino/WString.cpp
matthijskooijman: Keep the state sane is probably a good idea, so I'd suggest leaving these in.

hardware/arduino/avr/cores/arduino/WString.cpp
matthijskooijman: I guess having `std::move` on all these assignments would allow move semantics / optimizations on each of the fields, but since we know these are just a pointer and two integers, it should not make any difference here. So I guess the code is good as the PR proposes it.

leg0:
I removed `String(StringSumHelper&&)` constructor because having it made no sense at all. `String(String&&)` takes care of it - you can pass either `String&&` or `StringSumHelper&&` to this constructor. `StringSumHelper` is derived from `String` and has no data members other than those inherited from `String`. Even if it did have some extra data members, truncation would have to happen during move, and normally that is something you don't want.
The only reason (that I can think of right now) to have the removed constructor would be to prevent move construction from `StringSumHelper` to avoid truncation if that were a problem. I.e. declare it `String(StringSumHelper&&) = delete;`

hardware/arduino/avr/cores/arduino/WString.cpp
```diff
@@ -189,25 +188,22 @@ String & String::copy(const __FlashStringHelper *pstr, unsigned int length)
...
-		if (rhs && capacity >= rhs.len) {
-			strcpy(buffer, rhs.buffer);
-			len = rhs.len;
-			rhs.len = 0;
-			return;
-		} else {
+	if (this != &rhs)
+	{
+		if (buffer != NULL)
 			free(buffer);
```
BillyDonahue: FYI: The `if` is redundant since `free(nullptr)` is defined to be a safe no-op.
http://www.cplusplus.com/reference/cstdlib/free/

leg0: Thanks. Removed the null checks before free.

facchinm:
Hi @leg0  ,
Thanks!

Issue #6359, opened 2017-06-07
IDE window: CTRL-MouseWheel for zoom


snobu:
The inability to zoom in/out the code window is slightly annoying, i'm sure i'm not the only out there bothered by it. What's the development effort required to implement either `CTRL-MouseWheel` and/or `CTRL-+`, `CTRL-minus`? Do we have to move away from Java? :) Or just bind to some events?

I am aware a font size can be specified under Preferences, but that's `CTRL-comma - click-select - type - click`, not exactly the best UX.

bengtmartensson:
Possibly Java 9, due next month, will bring something here?

facchinm:
That's a good suggestion , thanks for posting! It should be fairly easy to trap the event inside the Editor view and apply the preference. Hot-reloading could be a bit more difficult but feasible indeed. If anyone wants to try coding a solution we'll be super happy to test it :smile:

skishida:
Now, font size may become a negative value in #6551 change and at that time, the IDE will not start.
The temporary fixing approach is edit `editor.font` in `preferences.txt`.

PR #6394, opened 2017-06-18
Add font size shortcuts (menu and keyboard)

gmichael225:
Having a stab at #6359.
(Scaling font size using keyboard shortcuts.)

I haven't implemented Ctrl-Scroll, but have implemented keyboard shortcuts using the existing menu item pattern.

Let me know if I should go about this a different way!

screen shot 2017-06-18 at 7 17 06 pm

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1374,6 +1374,24 @@ public class Editor extends JFrame implements RunnerListener {
 
     menu.addSeparator();
 
+    JMenuItem increaseFontSizeItem = newJMenuItem(tr("Increase Font Size"), '=');
+    increaseFontSizeItem.addActionListener(new ActionListener() {
+        public void actionPerformed(ActionEvent e) {
+          base.handleFontSizeChange(1);
+        }
+    });
+    menu.add(increaseFontSizeItem);
+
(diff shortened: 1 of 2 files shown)


facchinm:
@ArduinoBot build this please

app/src/processing/app/Editor.java
```diff
@@ -1374,6 +1374,24 @@ public void actionPerformed(ActionEvent e) {
 
     menu.addSeparator();
 
+    JMenuItem increaseFontSizeItem = newJMenuItem(tr("Increase Font Size"), '=');
```
facchinm: Why didn't you choose `ctrl + '+' ` as increase shortcut?  The `=` version doesn't seem to work with an Italian keyboard (since `=` can only be reached with `Maiusc` modifier)

gmichael225: Ah! I hadn't thought about other keyboards.
One convention I'm familiar with is to use - and = to avoid having to use a shift key to increase but not decrease the font size. (Plus can only be used with a shift modifier on a US keyboard.)

Perhaps `ctrl + '+'` should be the primary shortcut and `ctrl + '='` be secondary?

facchinm: Could work, I'm ccing our UX designer for an advice :slightly_smiling_face:  @00alis

facchinm:
Superseded by https://github.com/arduino/Arduino/pull/6551

Issue #6350, opened 2017-06-05
Version 1.8.2 three times slower to launch


technoblogy:
I've been comparing the time to launch 1.6.8, 1.8.1, and 1.8.2 on a MacBook Pro (Mac OS X 10.11.6).

Both 1.6.8 and 1.8.1 take 4 seconds, but 1.8.2 takes 13 seconds, over three times slower.

Any suggestions why this might be?

PaulStoffregen:
Maybe the newer JRE?

gilbertoverg:
I have the same issue, for me the launch time for 1.8.2 and 1.8.3 is 27 seconds

bperrybap:
Some of the problem (maybe most of it) is due to something in the IDE that ends up scanning the users sketchbook libraries and cores multiple times. It does this on startup or any time you switch cores.
I don't understand multiple scans would ever be necessary.

This has existed for many years but seems to have gotten worse recently.
It gets particularly slow when your sketchbook directory is across a network mount.

In the past it was very obvious for me as certain IDEs had some foolish naming restrictions on library directory names and I would have to answer the same prompt for a given library directory name up to 9 times for each directory it didn't like on startup or whenever I switched cores.
i.e. I'd get a slew warnings for directory names it didn't like - each of which must be individually acknowledged, and then this same sequence would repeat over and over again until the IDE finally started. 
It was a total pain.
The new IDEs no longer have the restrictions on directory names and hence don't toss up the warnings that you must acknowledge but they seem to still do the multiple scans of the users sketchbook libraries.

gilbertoverg:
@bperrybap I agree with you, I can see Arduino scanning libraries during startup with:
```sudo fs_usage pid <Arduino pid>```

technoblogy:
Thank you for confirming that I'm not the only one who has found this. I definitely think it would be worth the developers spending a bit of time to fix this.

cmaglie:
The multiple scanning of the libraries can be surely optimized, but it has been like this well before 1.8.1, IMHO the cause of the slow down should be searched elsewhere.

bengtmartensson:
I have found that "phony" serial ports (e.g. Bluetooth virtual such) can be startup time culprit. Can this be an issue here?

f3rland:
Thanks for pointing that out @bengtmartensson 
[...]

PR #6419, opened 2017-06-23
Only rescan libraries folders when really needed

facchinm:
Scanning libraries is an heavy task if the sketchbook becomes huge;
This patch targets two points:

- remove the rescan() after setLibrariesFolders(), which already performs a rescan
- call setLibrariesFolders() only when the folder list has changed
  - This ensures that no scan is performed when changing board in the same architecture

Could mitigate #6350


cmaglie:
Manually merged.

Issue #6442, opened 2017-06-29
Add 9bit support to HardwareSerial


svatoun:
I stumbled over a few different implementations / tweaks how Arduino could read 9bit data using Serial objects; the core library does not support 9bit.
I am aware of issues #2292 and #2291 - however I'd like to take a differen timplementation path: use guard macro to conditionally enable the extra code (and twice as long buffer). This way, normal user will not see any performance/size penalty, but developer who needs 9bit support could enable it at the documented cost.

Diffs will be attached shortly as PR...

PR #6443, opened 2017-06-30
Initial implementation of 9bit data support

svatoun:
Pull request to fix issue #6442.

Initial implementation of conditional 9bit support. Since 9bit transfers are rather niche, the implementation does not (almost) add any overhead for standard usage, buffer expansion and more complicated processing is guarded by SERIAL9 macro. Users must put -DSERIAL9 macro into their platform.txt in order to use the support - I didn't found other suitable way. Should be easier in the future if support for -D defines is added to the arduino IDE.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
@@ -82,9 +82,18 @@ void HardwareSerial::_tx_udr_empty_irq(void)
 {
   // If interrupts are enabled, there must be more data in the output
   // buffer. Send the next byte
-  unsigned char c = _tx_buffer[_tx_buffer_tail];
+  serial_data_t c = _tx_buffer[_tx_buffer_tail];
   _tx_buffer_tail = (_tx_buffer_tail + 1) % SERIAL_TX_BUFFER_SIZE;
-
+#ifdef SERIAL9
+  if (use9bit) {
+      if (c & 0x100) {
+        sbi(*_ucsrb, TXB80);
+      } else {
+        cbi(*_ucsrb, TXB80);
+      }
+      c = c & 0xff;
+  }
+#endif // SERIAL9
   *_udr = c;
 
   // clear the TXC bit -- "can be cleared by writing a one to its bit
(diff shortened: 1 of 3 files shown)


svatoun:
any news / progress / estimate when the PR could be reviewed(or rejected) ?

mikaelpatel:
What are the performance implications? How much does the effective baud-rate drop due to the additional handling of configuration? At lower baud-rates there is no impact but I guess that above approx 115Kbps the drop will start. A benchmark is required. 

As 9-bit mode is not a very common use case for Serial I would suggest considering a Serial sub-class. Also typical 9-bit mode usage is to signal the start and end of frames. This could be implemented without modifying the current Serial API. 

Cheers! Mikael

PaulStoffregen:
If this is merged (a pretty big "if"), I would like to suggest a longer, more descriptive define than SERIAL9.  Perhaps SERIAL_9BIT_SUPPORT could work?

basprins:
Any news on this? Multidrop bus is a 9bit uart protocol. I could sure benefit if this branch would ever be merged!

svatoun:
@basprins  I admit that I was lazy delivering the requested benchmarks; partially because given the nature of changes, the performance impact cannot be too high: one conditional jump based on register data operation, one IO `write` per-byte (not bit!). Similar for `read`.

IMHO worse effect is doubling the rx/tx buffer because of `unsigned char` to `unsigned short` change. Can be seen as bad on smaller Arduinos (I use Nanos ;)). Can be probably changed at the expense of some shift/and to use a bitfield for 9th bits - pls. give your opinions on whether you see that as worth the increased complexity.

Also note that if the guard macro is not defined, the change simply will not exist at all, leaving original code as it was - performance impact zero.

 But now I have two A-Mega boards available, so I can make some measurement. I assume that we need at least to: prove that 16MHz is able to tx / rx 115kbps. That's trivial - but do you have some suggestion to actually measure 'net consumed time' for (say) the write operation ? Sequential write or read operations are limited by UART speed (when it becomes avialable again) - how to measure the really tiny CPU time that one write opreation takes ?
Thanks for the advice.

mikaelpatel:
@svatoun After the latest @CLAassistant notice I revisited this pull request. I believe that if the purpose of 9-bit mode is to allow multi-drop protocols with the 9th bit for indicating address or data this could be done by simply adding a sendAddress(uint8_t addr) function to the HardwareSerial API. This function would set/reset the 9th bit accordingly.

PR #6446, opened 2017-07-01
Fix compilation warnings

jtentor:
Remove compilation warnings
- Add virtual destructor in HardwareSerial.h, Stream.h and Print.h
- Add attribute initialization in Stream.h

### DIFF
--- a/hardware/arduino/avr/cores/arduino/HardwareSerial.h
+++ b/hardware/arduino/avr/cores/arduino/HardwareSerial.h
@@ -118,6 +118,7 @@ class HardwareSerial : public Stream
       volatile uint8_t *ubrrh, volatile uint8_t *ubrrl,
       volatile uint8_t *ucsra, volatile uint8_t *ucsrb,
       volatile uint8_t *ucsrc, volatile uint8_t *udr);
+    virtual ~HardwareSerial() {}
     void begin(unsigned long baud) { begin(baud, SERIAL_8N1); }
     void begin(unsigned long, uint8_t);
     void end();
--- a/hardware/arduino/avr/cores/arduino/Print.h
+++ b/hardware/arduino/avr/cores/arduino/Print.h
@@ -44,7 +44,7 @@ class Print
     void setWriteError(int err = 1) { write_error = err; }
   public:
     Print() : write_error(0) {}
-  
+    virtual ~Print() {}
     int getWriteError() { return write_error; }
     void clearWriteError() { setWriteError(0); }
   
--- a/hardware/arduino/avr/cores/arduino/Stream.h
+++ b/hardware/arduino/avr/cores/arduino/Stream.h
@@ -60,8 +60,9 @@ class Stream : public Print
     virtual int read() = 0;
     virtual int peek() = 0;
 
-    Stream() {_timeout=1000;}
-
+    Stream() {_timeout=1000, _startMillis=0;}
+    virtual ~Stream() {}
+    
 // parsing methods
 
   void setTimeout(unsigned long timeout);  // sets maximum milliseconds to wait for stream data, default is 1 second


facchinm:
Hi @jtentor , thanks for the contribution! Did you check if the addition increases the generated binary size? Second question: how did these warnings appear? Did you enable `--pedantic` or similar?

jtentor:
1st question:
========
Yes, the addition increase the generated binary size.
Compiling this simple code with Arduino IDE
```
// -------------------
void setup() {
  Serial.begin(9600);
}
void loop() {
}
// -------------------
```
Without arduino original HardwareSerial.h, Print.h and Stream.h we have:
Sketch uses 1390 bytes (4%) of program storage space. Maximum is 32256 bytes.
Global variables use 182 bytes (8%) of dynamic memory, leaving 1866 bytes for local variables. Maximum is 2048 bytes.

With the addition to fix four warnings, three for virtual destructor and one for attribute not initialized we have:
Sketch uses 2042 bytes (6%) of program storage space. Maximum is 32256 bytes.
Global variables use 196 bytes (9%) of dynamic memory, leaving 1852 bytes for local variables. Maximum is 2048 bytes.

With the addition to fix only three virtual destructor warnings we have:
Sketch uses 2034 bytes (6%) of program storage space. Maximum is 32256 bytes.
Global variables use 196 bytes (9%) of dynamic memory, leaving 1852 bytes for local variables. Maximum is 2048 bytes.

It means we need 8 bytes for attribute initialization and 644 for virtual destructor.

In this trivial code represents relatively a lot of code, but it becomes less important as code grows.


2nd question:
=========
I am editing and compiling with Eclipse IDE for C/C++ Developers Version: Neon.3 Release (4.6.3) Build id: 20170314-1500 with Jan Baeyens plugin (http://eclipse.baeyens.it/).

Is possible Eclise have the "pedantic" option as default, but it is not set in my compilation options.

**In Arduino IDE we have -MMD option wich is not set in Eclipse IDE.**

The Eclipse IDE have a "Problems" windows where I found the four warnings, three for virtual destructor and one for attribute not initialized.

Possible this warnings are not relevant for hobbyist products, but I am a software engineering and checking for no warnings in my compilations is a quality factor.

Thanks @facchinm

facchinm:
You are welcome :slightly_smiling_face: 
Adding (virtual) destructors to a couple of important base classes (like Print or Stream) have been discussed in the past, but it only increases the final binary size and they are mostly useless in embedded projects, thus the idea has been discarded. In boards like the UNO every single byte is important so adding an useless feature to make the static analyzer happy is a no-go. 
About the command line, `-MMD` is used to generate the intermediate dependency files, so it shouldn't affect the compiler output in any way.

jtentor:
You are correct, **embedded systems do not need unnecessary code bytes**.

I'm going to work on some -D options that enable or disable these virtual constructors

As soon as I have them I will share them.

Thank you.

jtentor:
Hi @Ivan-Perez , from your question **Is not it important that the library behaves correctly?**

For software engineering is very important, but in small embedded systems or that who not use virtual destructor this adds unnecessary code.

Remember in first instance, Arduino is for little embedded systems; it is a stakeholder decision.

PaulStoffregen:
> fixes memory leaks ... on bigger projects which extend and add functionality of any of the Print subclasses.

Can you give some specific examples where this has been a problem?

Ivan-Perez:
Hi @jtentor and @PaulStoffregen, I showed an example on one comment in the PR a linked in my previous message. Here you have the link to that coment: https://github.com/arduino/Arduino/pull/4466#issuecomment-194230165. I think it is a very simple example.

Another person show me a workaround to fix my issue, so no complains, but it would be easier to just delete object without using such workarounds. In my project, adding the `Print` virtual destructor only adds a few bytes (less than 100), as I use them everywhere else.

facchinm:
Hi @jtentor ,
Thanks!

Issue #6455, opened 2017-07-03
stk500v2 bootloader corrupts data when reading EEPROM


mlasevich:
When reading EEPROM via avrdude using wiring (via bootloader) - results are corrupted. Every other 8 bytes are lost. Writing to EEPROM works, but, as expected, validation fails as what is written does not match corrupted version that is read back.

You can reproduce on Mega2560(and probably others using this bootloader) using following sketch to initialize EEPROM to known state:

```c++
#include <EEPROM.h>

void eepromDump() {
  char buff[8];
  int counter = 0;
  byte data = 0;
  Serial.write("EEPROM contents:\n");
  for (int i = 0; i < 4096 ; i++) {
    counter = counter % 16;
    if (counter == 0) {
      snprintf(buff, 10, "%07X", i);
      if (i != 0) {
        Serial.println();
      }
      Serial.print(buff);
    }
    data =  EEPROM.read(i);
    snprintf(buff, 4, " %02X", data);
    Serial.print(buff);
    counter ++;
  }
  Serial.println();
}

void eepromErase(int start){
  Serial.println("Erasing EEPROM");
  for (int addr=start; addr <4096; addr++){ EEPROM.update(addr, 0x00); }
}

void setup() {
  Serial.begin(115200L);
  const char data[63]="0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
  EEPROM.put(0, data);
  eepromErase(64);
  eepromDump();
}

void loop() {
}
```

After that if executing a read via a programmer you get expected correct output:
```
$ avrdude -c dragon_isp -P usb -p atmega2560 -Ueeprom:r:eeprom.bin:r
$ hexdump -C eeprom.bin
00000000  30 31 32 33 34 35 36 37  38 39 61 62 63 64 65 66  |0123456789abcdef|
00000010  67 68 69 6a 6b 6c 6d 6e  6f 70 71 72 73 74 75 76  |ghijklmnopqrstuv|
00000020  77 78 79 7a 41 42 43 44  45 46 47 48 49 4a 4b 4c  |wxyzABCDEFGHIJKL|
00000030  4d 4e 4f 50 51 52 53 54  55 56 57 58 59 5a 00 ff  |MNOPQRSTUVWXYZ..|
00000040  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00001000
```

But using the bootloader for same thing you get corrupted output:

```
[...]

PR #6456, opened 2017-07-03
Fixed READ EEPROM in stk500v2 bootloader

mlasevich:
As per #6455 

CMD_READ_EEPROM_ISP was addressing EEPROM as 16bit instead of 8bit data.
Changed to mimic CMD_PROGRAM_EEPROM_ISP workaround.

### DIFF
--- a/hardware/arduino/avr/bootloaders/stk500v2/stk500boot.c
+++ b/hardware/arduino/avr/bootloaders/stk500v2/stk500boot.c
@@ -1063,10 +1063,13 @@ int main(void)
 						else
 						{
 							/* Read EEPROM */
+							uint16_t ii = address >> 1;
 							do {
-								EEARL	=	address;			// Setup EEPROM address
-								EEARH	=	((address >> 8));
-								address++;					// Select next EEPROM byte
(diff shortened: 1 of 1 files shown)


mlasevich:
What is the status of this? Will this be merged into the official bootloader or is there an issue with it?

Thanks.

facchinm:
@mlasevich we usually update the bootloaders as soon as they are used in production. Any bugfix is well accepted but will be merged only if the generated binary has been validated

mlasevich:
Thanks for the reply. I mostly wanted to know if I needed to do anything else to get this accepted and it sounds like that is not the case.  It is no rush, we can perfectly well use the custom built bootloader for our purposes, but it always feels better to use the official build :-) 

For my education, what does "used in production" mean - does that mean it will be tested once a new board using it will be released or something else? 

Thanks :-)

facchinm:
Exactly, as soon as a board is being factory-flashed with a given bootloader we update the corresponding source and binary :wink:

mlasevich:
Closing as the relevant file is no longer in this repo. Opening new PR here: https://github.com/arduino/ArduinoCore-avr/pull/24

Issue #4753, opened 2016-03-24
Can't access buttons on Preferences window on screens with vertical resolution of 600 px


yvescordier:
Arduino IDE 1.6.8
The preference window has no vertical scrollbar. So it is impossible to access the OK button on the bottom of the window on a PC (old recycle for courses) with lower resolution (1024x600 or 800x600).

Would it be possible to reorganize the items of this window to be able to access then on 10'' old notebook? There is this huge useless empty gap between the end of the last preference and the buttons. There is a lot of room unused between lines and another way could be to put the OK and Cancel buttons on the top of the box.
![preferences](https://cloud.githubusercontent.com/assets/18050289/14016794/b0e44142-f1c2-11e5-9f86-9f2c43158d91.jpg)

lmihalkovic:
@yvescordier this was my own thinking on the preferences: #4421, #4417, [#4102/comment](https://github.com/arduino/Arduino/pull/4102#issuecomment-171289157): the single page has been split into more logical smaller pages accessible via the left hand-side list. I think the arduino team's original argument that 'users are so non-tech that they can't be overwhelmed with too many choices' has over the years turned into an efficient justification for doing the bare minimum, especially considering how the IDE (not the libraries) could be construed as having entered _de-facto sunset mode_.

yvescordier:
Sorry but english is not my mother language. My problem is not to split the informations and checkboxes in two or three pages but the problem is to be able to access all informations of the preference window on a screen with low resolution (1024*600). On such screen, it is not possible to acces the OK button to validate the choices. It is why I propose to have a slimerst box of about 500 point heigh (this could be done by elimating usless zones).

regards
yves cordier

lmihalkovic:
```
this could be done by elimating usless zones
```
... or by avoiding to cram together things that have nothing in common, but result in a window too big for everything to be visible on a smaller resolution monitor. But the arduino designers likely has a plan for what they want to do.
Cheers.

Reminds me of this architect friend who used to say nobody could pay him enough to make him design a toilet in the center of a living room, without a very serious reason...

facchinm:
Fixed by #6459

yvescordier:
sorry to have been late to try the modification but i Had'nt my computer avilable. 

due to function bar of  XP on the bottom of the screen, it sill be difficult to press the buttons

only 10 pixels of the button are visible

50 pixels less should be apeciate

is it possible to remove 2 or 3 trailing lines more at the bottom of the preference windows

 

[...]

PR #6459, opened 2017-07-05
Remove gaps on confirmation button to fit 600px height

facchinm:
Remove some unused space from Preference panel to be able to click OK and Cancel on 800x600px monitors. 
On some configurations the OS could decorate the window and make it bigger than 600px; in this case, we should split the preferences into two tabs (argh) or make the window scrollable (even worse). Let's try this approach and see if it works.

Should solve #4753


rin67630:
Why did  cmaglie delete the facchinm:smaller_preference_panel branch. We DO need it!

PaulStoffregen:
This fix was merged more than 1 year ago.

If there is *still* a problem with 1.8.7 or the latest nightly build, maybe open a new issue.  Wouldn't hurt to reference this one and mention the fix wasn't enough for some displays.  A screenshot of the actual problem and specific info about which operating system and desktop environment would also go a long way towards persuading anyone that this deserves more attention.

cjdg:
Happening again in 1.8.9

PR #6580, opened 2017-08-02
Turn off BOD to so that MCU will run at 3v

alustig3:


matthijskooijman:
I'm not so sure we want to turn the BOD off entirely. IIRC it normally triggers at 2.7V (typical), which is fine for the Pro Mini that runs at 3.3V normally. If you want to run your board at different settings, you should just apply your own fuses settings, no need to apply these to everyone else, I'd think.

Additionally, I'd prefer these settings to be the same with the factory-defaults for the Pro Mini, so it doesn't matter if you re-burn the bootloader or not.

alustig3:
Very well. Thanks for the quick response!

PR #6601, opened 2017-08-10
Program unused lock bits to 1

Osambezy:
When burning the bootloader, write access to the Boot Loader section is first unlocked and then locked again by setting the respective Lock bits in the Lock Bit Byte. Bits 7 and 6 are unused according to the datasheets of ATmega168, ATmega328P, ATmega32U4 and ATmega2560. They should therefore always be programmed to 1.

Currently they are programmed to 0. It still works, but could lead to undesired behavior. Avrdude issues a warning, that will turn into an error in the future.

This patch fixes all Lock bytes such that the unused bits will be programmed to 1.

### DIFF
--- a/hardware/arduino/avr/boards.txt
+++ b/hardware/arduino/avr/boards.txt
@@ -31,8 +31,8 @@ yun.bootloader.high_fuses=0xd8
 yun.bootloader.extended_fuses=0xfb
 yun.bootloader.file=caterina/Caterina-Yun.hex
 yun.bootloader.noblink=caterina/Caterina-Yun-noblink.hex
-yun.bootloader.unlock_bits=0x3F
-yun.bootloader.lock_bits=0x2F
+yun.bootloader.unlock_bits=0xFF
+yun.bootloader.lock_bits=0xEF
 
 yun.build.mcu=atmega32u4
 yun.build.f_cpu=16000000L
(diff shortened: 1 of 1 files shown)


matthijskooijman:
IIRC there is a related issue. One reason for not setting them to 1 originally is that, on some commonly used programmers, avrdude reads these bits back as 0, so setting them to 1 makes verification fail. This was fixed in the Arduino version of avrdude by excluding the unused bits from the verification comparison. From a quick look at the patch, it seems this only applies to fuse bits, lock bits might have been left out.

In any case, the patch was discussed here: https://github.com/arduino/avrdude-build-script/issues/2 and originally here: https://github.com/arduino/Arduino/pull/5202
Setting the unused fuse bits to 1 happened here: https://github.com/arduino/Arduino/pull/5182

matthijskooijman:
Note that on some programmers (such as FTDI bitbang), avrdude will actually *start* showing the warning with this change. This is something to be fixed upstream in avrdude, see https://savannah.nongnu.org/patch/?8996

Osambezy:
Yes, fixing this in avrdude seems to be the way to go, thanks for the info! If the datasheet says that unused bits are 1, then it should be kept like that all the way through the toolchain.

matthijskooijman:
I came across this PR again, and I think the changes in it should maybe be applied? Looking at the current avrude patch (https://github.com/arduino/avrdude-build-script/blob/master/avrdude-6.3-patches/80-Avoid-failing-fuse-check-if-different-bits-are-reserved.patch) it does seem to apply to the lock bits as well, so I'm not sure why @descampsa only made this change for fuse bits and not lock bits in https://github.com/arduino/Arduino/pull/5182

As for introducing new avrdude warnings, those should be fixed upstream, but in the meanwhile updating the avrdude patch to the one I submitted upstream (see https://github.com/arduino/avrdude-build-script/issues/2#issuecomment-323438775) would at least make the warning shown for programmers returning 0 more appropriate.

If this is done, this PR should be resubmitted to https://github.com/arduino/ArduinoCore-avr, since that's where this boards.txt file lives now. Probably needs to be retested on a bunch of different programmers too...

@facchinm, what would you think?

Issue #6408, opened 2017-06-22
the find window not always on top of the main window


czghzc:
the find window not always on top of the main window

facchinm:
As far as I can tell, the "find" window always appears at the center of the active IDE window. 
Are you experiencing something different? Can you provide a screenshot? Thanks!

czghzc:
![2017-06-22_110300](https://user-images.githubusercontent.com/23645596/27422868-4d1f5c4e-5762-11e7-947c-15666d612d27.gif)

facchinm:
Mmmmh, it looks like a problem with Windows' window manager  (and probably Java). I'll try investigating before updating to the next Java version.

czghzc:
My system is Win7-32Bit.
My previous version of Java is 1.7.0_80-b15, there is such a problem.
Then I updated the Java version to 1.8.0_131-b11, the problem is still not resolved.
If you have no problem with the test, it may be my system.

czghzc:
![2017-10-07_092746](https://user-images.githubusercontent.com/23645596/31303575-957ea1c8-ab42-11e7-9d30-7f0c71bdbf53.png)

I'm sorry, but now I understand your reply, it is ashamed.
I tested it, and now the find window is above all windows, but it finds another problem.
If there are other windows in the system, such as a browser, the find window will also be on top of the browser window, which will interfere with the operation of the browser window, and we hope that the find window is just above the Arduino main window.

kenneth558:
Agreed that the find window drops behind main window.  Linux Mint 18, IDE 1.8.5, but has always been the case in IDE on Linux

facchinm:
Would you mind testing https://github.com/arduino/Arduino/pull/6603 (or the latest beta)? The wrong behaviour should be gone for good.

czghzc:
![pic2](https://user-images.githubusercontent.com/23645596/37202019-56b78d10-23c4-11e8-8436-fd2f649f2467.gif)
The beginning is good, there is no problem, but after clicking the window's title bar with a mouse is not normal, the search window goes to the back of the main window.

When programming in C++, setting the search window to the child window of the main window can achieve the desired effect. However, I will not use Java. I do not know if Java has the same operation method.

PR #6603, opened 2017-08-10
Make "find and replace" window always appear on top of the parent editor

facchinm:
Fixes https://github.com/arduino/Arduino/issues/6408

### DIFF
--- a/app/src/cc/arduino/view/findreplace/FindReplace.java
+++ b/app/src/cc/arduino/view/findreplace/FindReplace.java
@@ -73,10 +73,24 @@ public class FindReplace extends javax.swing.JFrame {
 
     Base.setIcon(this);
 
+    editor.addWindowListener(new WindowAdapter() {
+      public void windowActivated(WindowEvent e) {
+        findField.requestFocusInWindow();
+        findField.selectAll();
+        setAlwaysOnTop(true);
+      }
+      public void windowDeactivated(WindowEvent e) {
+        setAlwaysOnTop(false);
+      }
+    });
+
     addWindowListener(new WindowAdapter() {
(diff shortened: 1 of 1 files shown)


czghzc:
I'm sorry, but now I understand your reply, it is ashamed.
I tested it, and now the find window is above all windows, but it finds another problem.
If there are other windows in the system, such as a browser, the find window will also be on top of the browser window, which will interfere with the operation of the browser window, and we hope that the find window is just above the Arduino main window.

What we want is like this picture below:

facchinm:
Hi @czghzc ,
I had the chance to test it on Windows and indeed it behaves exactly as you described.
I just pushed 09f5196 which should make the behaviour consistent across all OSes. The build by @ArduinoBot should arrive soon :wink:

czghzc:
Hi @facchinm ,
I tested the windows version because i only had windows system, just as I wanted it, it worked very well and thank you so much thanks!

kenneth558:
With 6603 the behavior is noticeably better, for doggone sure!  Only thing is, still, I when I file-save via the menu dropdown, the find window still goes behind right when I select the final menu selection.

czghzc:
pic2
The beginning is good, there is no problem, but after clicking the window's title bar with a mouse is not normal, the search window goes to the back of the main window.

When programming in C++, setting the search window to the child window of the main window can achieve the desired effect. However, I will not use Java. I do not know if Java has the same operation method.

In addition, I also have a suggestion that when the search window comes out, the content of the search text box is selected, which facilitates us to modify the content to be searched.

kenneth558:
I agree with causing the text to be selected when re-pressing ctrl-f.  I just thought you had a reason for not having it that way in the first place.

PR #6608, opened 2017-08-13
Add support for 64 bit integer printing

RobTillaart:
Added support for printing **unsigned long long** and **long long**

solves issue https://github.com/arduino/ArduinoCore-avr/issues/194

Included is a reference implementation (commented out) which is a replica of the print code for **long**.
The code is 2-3x faster with a slightly larger footprint. The performance is gained by working in "16 bit" chunks. Tested on UNO and discussed "long long" ago here - https://forum.arduino.cc/t/serial-print-of-a-64-bit-double/140151

(minor) Cleanup trailing spaces

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -1,21 +1,21 @@
 /*
  Print.cpp - Base class that provides print() and println()
  Copyright (c) 2008 David A. Mellis.  All right reserved.
- 
+
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
  License as published by the Free Software Foundation; either
  version 2.1 of the License, or (at your option) any later version.
- 
+
  This library is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


Ivan-Perez:
Might this be added to the `String` class, so it also supports 64-bit integers?

RobTillaart:
@Ivan-Perez 
The String Class could be extended with two functions (one given here)

String::String(uint64_t value, unsigned char base)
{
	init();
	char buf[1 + 8 * sizeof(unsigned long)];
	ul64toa(value, buf, base);
	*this = buf;
}

The AVR stdlib does not have a ul64toa() so you need something like this

// straightforward implementation
void ui64toa(uint64_t v, char * buf, uint8_t base)
{
  int idx = 0;
  uint64_t w = 0;
  while (v > 0)
  {
    w = v / base;

    buf[idx++] = (v - w * base) + '0';
    v = w;
  }
  buf[idx] = 0;
  // reverse char array
  for (int i = 0, j = idx - 1; i < idx / 2; i++, j--)
  {
    char c = buf[i];
    buf[i] = buf[j];
    buf[j] = c;
  }
}

A similar function for a signed version void i64toa()

facchinm:
This PR was merged in https://github.com/arduino/ArduinoCore-API/commit/3c3b27e94be308184054fb49a0fc329d3c88801e , closing here

Issue #6335, opened 2017-05-29
Please add a keyboard shortcut to open the Library Manager


feikname:
It's a GUI used often by me, and having a option to open it by keystrokes instead of navigating through the menu would greatly improve my workflow (and of many people, I hope).

Suggested shortcut: <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>K</kbd> (No idea why, it's just what comes to my mind)

feikname:
Fixed in 38e23a7.

PR #6638, opened 2017-08-19
Add "Manage libraries..." to the "Tools" menu too and add a shortcut for it.

### LINKED ISSUE #3635: Add sketch path to the compile include path

feikname:
**Disclaimer:** The original intent of this pull request was to move the "Manage Libraries..." item to the Tools Menu". The original text is available below.

This pull request adds the Menu+Shift+I shortcut for easy acess to the Library Manager, thus fixing #6335

It also adds the "Manage Libraries..." item to be available in the Tools menu, while also keeping the old entry.

The result can be seem in the last image of the original text.

------ 

Original:

~Description from the commit message:~
>It's easier to access it there
>
>Also, a shortcut was added to it: <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>O</kbd>
>
>  My original plan was for it to be "Control + Shift + K", but for some
>reason it doesn't work on Xubuntu 17.04

This is my attempt at fixing #6335

If you think changing menu location is too big of a change, I can close this pull request and open a new one that only adds the shortcut.

Before:\
image

Now:\
image

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1070,6 +1070,10 @@ public class Base {
     importMenu.removeAll();
 
     JMenuItem menu = new JMenuItem(tr("Manage Libraries..."));
+    // Ctrl+Shift+I on Windows and Linux, Command+Shift+I on macOS
+    menu.setAccelerator(KeyStroke.getKeyStroke('I',
+        Toolkit.getDefaultToolkit().getMenuShortcutKeyMask() |
+        ActionEvent.SHIFT_MASK));
     menu.addActionListener(e -> openLibraryManager("", ""));
     importMenu.add(menu);
     importMenu.addSeparator();
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -158,7 +158,7 @@ public class Editor extends JFrame implements RunnerListener {
 
   static volatile AbstractMonitor serialMonitor;
   static AbstractMonitor serialPlotter;
-  
+
   final EditorHeader header;
   EditorStatus status;
   EditorConsole console;
@@ -248,7 +248,7 @@ public class Editor extends JFrame implements RunnerListener {
 
     //PdeKeywords keywords = new PdeKeywords();
     //sketchbook = new Sketchbook(this);
-    
+
     buildMenuBar();
 
     // For rev 0120, placing things inside a JPanel
@@ -399,8 +399,7 @@ public class Editor extends JFrame implements RunnerListener {
         statusNotice(tr("One file added to the sketch."));
 
       } else {
-        statusNotice(
-	    I18n.format(tr("{0} files added to the sketch."), successful));
+        statusNotice(I18n.format(tr("{0} files added to the sketch."), successful));
       }
       return true;
     }
@@ -947,14 +946,14 @@ public class Editor extends JFrame implements RunnerListener {
     } finally {
       if (zipFile != null) {
         try {
-          zipFile.close();
-        } catch (IOException e) {
-          // noop
-        }
-      }
-    }
-    return null;
-	}
+           zipFile.close();
+         } catch (IOException e) {
+           // noop
+         }
+       }
+     }
+     return null;
+   }
 
   public void updateKeywords(PdeKeywords keywords) {
     for (EditorTab tab : tabs)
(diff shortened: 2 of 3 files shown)


facchinm:
Hi @feikname , the shortcut change is really needed but I can't find the relevant commit. Did you forget to push it?
I'd also be fine with moving the menu but I'd ask confirmation to @00alis about this topic.

00alis:
I think we should not move it, but have it in both places, probably a lot of people are used to find it there. I agree that it could be useful under Tools too :)

feikname:
@facchinm The shortcut addition was placed in the same commit as the one that moves the item:
```java
JMenuItem item = newJMenuItemShift(tr("Manage Libraries..."), 'O'); // newJMenuItemShift is a wrapper function inside Editor.java for creating Ctrl Shift shortcuts in modern operating systems
item.addActionListener(e -> base.openLibraryManager("", ""));
toolsMenu.add(item);
```

@00alis I'd rather not have duplicate entries, but I will make a commit to re-add the "Manage Libraries..." item to it's old place soon.

feikname:
The new commit (ad84772c27c418c63867d27ecca2237f4d58bac2) should re-add the item back into its old place, although I didn't test it yet.

OtacilioN:
I really agree with @00alis, if we move it we will loose a lot of tutorials and documentations that says to users that you can find the Arduino Library Manager in the Sketch Menu.

But yeah, i do believe that the tools menu is right Menu for this tool, so instead moving i do believe that has no problem for the users duplicate this section in both menus, so we can acess even easier this part wich is so important.

And for finishing the idea in #6335 to make keyboard shortcut it is really necessary to make the Arduino Library Manager even more accessible.

feikname:
Good point, I didn't think about all the old tutorials. 

So now that the entry is present in both its old place and the Tools menu, and the shortcut was added, I will rebase the commits.

The only thing left is to confirm if the Ctrl Shift O shortcut is okay.

OtacilioN:
Crtl Shift I does something? Because I could be from 'I'nclude, wich renember libraries, maybe could be a better shortcut, but Crtl Shift O i dont see any problem in the Crtl Shift O

feikname:
Okay, this pull request should be ready now.

I changed the shortcut from Menu+Shift+O to Menu+Shift+I because the former option seemed awfully similar to the Open File shortcut to me.

feikname:
- True
- Sure, no problem. 
- I didn't even think about this, as I don't use the Board Manager myself much. It can be added (given that the code is trivial, I hope), I am just not sure on how to rename the pull request doing it./
Shortcut suggestions are welcome.
- I have no idea

feikname:
Can someone test bd4d9c5 please? I just need to know if the shortcut still works and if the tip is shown on both entries.

feikname:
I did a forced push again, the pull request should be ready now. 

> You added the shortcut in only one of the two places, so it is shown only there. Can't you just add it in both places? Or will the command then execute twice?

Yes, it's possible. The b57d2fa6e9399aba8ca3814d341e5896613cf5bb commit (previously bd4d9c548293821e968c4306be5541e9cc265573, I only renamed it) does exactly that. The entry that gets called is the one from `Base.java` only.

> Wouldn't it be consistent to also allow managing the installed boards from the tools menu?

IMO, it's already easily reachable, unlike the "Manage Libraries..." item. Maybe a shortcut could be added to it, but that's probably better to be included in another PR to keep things clean.

I see there's [this commit](38e23a7530806ec7af3a2543116f6c930337f1a8) in the `ide-1.9.x-beta` branch that was included from the [previous state](4f3e2d96f7b94aac7e629e127ca8078edfae9929) of this pull request. If you plan on merging the `ide-1.9.x-beta` branch completely, I recommend to change this before doing so.

PR #6750, opened 2017-09-21
Added support for HEX leading zero in print

Tijgerd:
Added the leading zero for the print funcionality

if the argument PRINT_LEADINGZERO is added after the base definition, a leading zero is added to the hex

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Print.cpp
+++ b/hardware/arduino/avr/cores/arduino/Print.cpp
@@ -69,22 +69,22 @@ size_t Print::print(char c)
   return write(c);
 }
 
-size_t Print::print(unsigned char b, int base)
+size_t Print::print(unsigned char b, int base, int arg)
 {
-  return print((unsigned long) b, base);
+  return print((unsigned long) b, base, arg);
 }
 
-size_t Print::print(int n, int base)
+size_t Print::print(int n, int base, int arg)
 {
-  return print((long) n, base);
+  return print((long) n, base, arg);
 }
 
-size_t Print::print(unsigned int n, int base)
+size_t Print::print(unsigned int n, int base, int arg)
 {
   return print((unsigned long) n, base);
 }
 
-size_t Print::print(long n, int base)
+size_t Print::print(long n, int base, int arg)
 {
   if (base == 0) {
     return write(n);
@@ -92,18 +92,18 @@ size_t Print::print(long n, int base)
     if (n < 0) {
       int t = print('-');
       n = -n;
-      return printNumber(n, 10) + t;
+      return printNumber(n, 10, arg) + t;
     }
-    return printNumber(n, 10);
+    return printNumber(n, 10, arg);
   } else {
-    return printNumber(n, base);
+    return printNumber(n, base, arg);
   }
 }
 
-size_t Print::print(unsigned long n, int base)
+size_t Print::print(unsigned long n, int base, int arg)
 {
   if (base == 0) return write(n);
-  else return printNumber(n, base);
+  else return printNumber(n, base, arg);
 }
 
 size_t Print::print(double n, int digits)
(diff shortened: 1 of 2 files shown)


Tijgerd:
@clivepengelly, @fake-name, technikadonis, @cmaglie is this a solution for the problem?
Anyone willing to test this?

fake-name:
I still stand that the current implementation is broken, and the correct option is to have print `HEX` **always** emit a leading zero, but that's probably not a winning opinion at this point.

Tijgerd:
@fake-name ,
 I do support your opinion, I'm using HEX because I want it to be with leading zero..

BUT

HEX, DEC, etc. are a bunch of predefined values indicating the base of the value (e.g. HEX = 16)
This base is used as a divider for the value, so for each "letter" printed, the value is divided by the base.

adding a strange value for the base, like ACTUAL_HEX, will be a bit odd because it will be something like: 42 which is not a base value.

I added a OPTIONAL input value which defaults to PRINT_NOARG when not assigned.
if it IS assigned it checks if the given HEX value is smaller than 16, if so, it will add a leading zero.

With this solution, old scripts can still be used :) 
I think that's the whole point.. they want it to be backward compatible.

fake-name:
> HEX, DEC, etc. are a bunch of predefined values indicating the base of the value (e.g. HEX = 16)
This base is used as a divider for the value, so for each "letter" printed, the value is divided by the base.

So change that to an enum?

Tijgerd:
That's what my initial plan was BUT there may be some users that did actual write the base number and so.. It won't be backward compatible.

I know, it doesnt feel right for the HEX not printing the leading zero but this was once implemented and changing it would make a small group of arduino users unhappy ;)

fake-name:
> That's what my initial plan was BUT there may be some users that did actual write the base number and so.. It won't be backward compatible.

That's not documented, and the documentation makes no statements about *what* `HEX`, `DEC`, etc... actually *contain*, so they're relying on undefined behaviour *now*. Their code is already broken, it just coincidentally works.

Tijgerd:
https://www.arduino.cc/en/Serial/Print
https://www.arduino.cc/en/Serial/Println

It's kind of documented...

fake-name:
Ah, it's kind of crappily documented on https://www.arduino.cc/en/Serial/Println, the contents of which actually *contradict* https://www.arduino.cc/en/Serial/Print.

Siiiiiiiigh, Arduino code in a nutshell.

Tijgerd:
True :+1:

fake-name:
Anyways, `print()` explicitly enumerates the valid values for `format` when `val` is integral, so I'm still sticking with just change the function of `format`. If you really want to keep existing behaviour, just place the new enum values in the top bits of `format`. If anyone is printing in > base 16, they're sufficiently nuts that I think we can ignore them.

Tijgerd:
@facchinm any news on this?

cousteaulecommandant:
Some ideas/opinions/thoughts:
1. Arduino's `print()` is meant as a simple printing function rather than a fully featured formatting one, so I'm not sure adding more arguments is a good idea.
2. For more advanced formatting specifications, I think using `printf()` is a better idea. See #5938 which adds support for `Serial.printf()`.
3. Back in the day I thought of including more fields in the format specifier, like
    ```arduino
    #define WIDTH(x) ((x)<<8) // up to 127
    #define SIGN (1<<7) // include +/-
    #define ZERO_PAD (1<<6) // pad with zeros instead of spaces
    #define BASE(x) (x) // up to 36
    Serial.print(x, WIDTH(2) | ZERO_PAD | BASE(16));
    ```
    which would allow cramming multiple format specifiers on a single argument, thus being compatible with the proposal in #5829.
4. Maybe it is a good idea to just print UNSIGNED integers (byte, word, unsigned long) as fixed width, and SIGNED integers (signed char, int, long) as regular numbers with a sign (e.g. `Serial.print(-15,HEX)` should print `-F`, not `FFFFFFF1`; see #4535). This way, a simple conversion would be enough in most cases.

hardware/arduino/avr/cores/arduino/Print.cpp
```diff
@@ -209,14 +209,16 @@ size_t Print::printNumber(unsigned long n, uint8_t base)
 
   // prevent crash if called with base == 1
   if (base < 2) base = 10;
-
-  do {
+  //the argument is only valid for HEX values smaler than 0x10
+  if(base != 8 || n>=16) arg = 0;
```
retfie: if(base != **HEX** || n>=16) arg = 0;

Tijgerd:
@retfie , if I remember correctly, **HEX** is defines as 16 (By default in the original arduino code)

Issue #6669, opened 2017-08-30
Arduino Micro USB Serial cannot receive more than 255 data once


LeoYan:
Issue:  Arduino Micro USB Serial is crashed if send more than 255 data once to Serial.  when it is occurred , the USB register is:  1) UEINTX = 0b 10010100  2)UEBCLX = 0b 0. That means the RX FIFO is empty but FIFOCON is not released, so the host has to wait until timeout.

sandeepmistry:
Hi @LeoYan, Could you please try out the changes I proposed in pull request https://github.com/arduino/Arduino/pull/6886, so see if they resolve this issue?

LeoYan:
@sandeepmistry  I have tested it on MAC,   it works:)  Thank you very much.

PR #6886, opened 2017-11-03
Handle receiving a ZLP in USB_Available

sandeepmistry:
To address https://github.com/arduino/Arduino/issues/6669.

If a ZLP is received the FifoCount will be zero, but `RXOUTI` will still be triggered.

@NicoHood could you please try this change out to ensure it's ok HID API wise. Thanks.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/USBCore.cpp
+++ b/hardware/arduino/avr/cores/arduino/USBCore.cpp
@@ -171,6 +171,11 @@ static inline u8 FifoFree()
 	return UEINTX & (1<<FIFOCON);
 }
 
+static inline u8 HasOUT()
+{
+	return UEINTX & (1<<RXOUTI);
+}
+
 static inline void ReleaseRX()
 {
 	UEINTX = 0x6B;	// FIFOCON=0 NAKINI=1 RWAL=1 NAKOUTI=0 RXSTPI=1 RXOUTI=0 STALLEDI=1 TXINI=1
@@ -214,7 +219,12 @@ public:
 u8 USB_Available(u8 ep)
 {
 	LockEP lock(ep);
-	return FifoByteCount();
+	u8 n = FifoByteCount();
+
+	if (!n && HasOUT())
+		ReleaseRX(); // handle ZLP
+
+	return n;
 }
 
 //	Non Blocking receive


NicoHood:
@sandeepmistry I currently do not have the hardware nor time to test this, I am sorry. Thanks for mentioning me though. If the official arduino hid libraries work, mine should work too.

sandeepmistry:
@kervinck would you be able to open a new pull request for your suggested changes?

kervinck:
That would take me one or two weeks of study, cleanup and testing. Sorry.

obra:
I've just tested this patch against a longstanding issue I've had where, only on macOS, sending 384 or more bytes to the AVR USB Serial device from the host causes USB Serial to lock up.

...and it appears to solve the issue.


Without this patch, sending the following string (or any string of equal or greater length) from the Classic Arduino IDE's serial console will cause the USB Serial connection to stop reading input, but only on macOS.

```
0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef01234567890abcde
```

The trivial repro sketch for this failure is 
```

void setup() {Serial.begin(9600);}
void loop()  {if (Serial.available()) {Serial.write(Serial.read());}}
```

@cmaglie Apologies for tagging you directly, but who's the right code owner to see about getting this PR either merged or rejected?

tlyu:
Confirmed the observation by @kervinck that this fix is incomplete. If the application calls `USB_Recv` before `USB_Available` when a ZLP has been read, it will stop reading new input until it calls `USB_Available`. (The USB hardware might accept and ACK an additional packet, due to double buffering, but the application can't read it.)

This additional patch appears to fix it in our use case (diff is against the keyboardio fork of the core):

```diff
--- a/avr/cores/keyboardio/USBCore.cpp
+++ b/avr/cores/keyboardio/USBCore.cpp
@@ -245,6 +245,10 @@ int USB_Recv(u8 ep, void* d, int len)

        LockEP lock(ep);
        u8 n = FifoByteCount();
+       if (!n && HasOUT()) {
+               ReleaseRX(); // handle ZLP
+               n = FifoByteCount();
+       }
        len = min(n,len);
        n = len;
        u8* dst = (u8*)d;
```

I also agree that the datasheet says to always clear `RXOUTI` before `FIFOCON`, but this doesn't seem to cause problems in practice. It also doesn't seem to be the problem here. The AVR core doesn't use interrupts for non-control endpoints, so it's less important to clear `RXOUTI` as soon as possible. (`FIFOCON` seems to be what controls whether the USB hardware will send NAK to the host.)

tlyu:
This probably needs to be moved to https://github.com/arduino/ArduinoCore-avr

PR #6913, opened 2017-11-10
Improve accessibility in the Preferences dialog

coreyknapp:
This commit makes the Preferences dialog more usable by accessibility devices like screen readers.

Previously, a screen reader would only read the content of the text boxes. By using JLabel's setLabelFor() feature (and in some extreme cases, setAccessibleName()), a screen reader will know that a label is assigned to a text field. For example, instead of just reading "12", it will now read "Editor font size: 12".

### DIFF
--- a/app/src/cc/arduino/view/preferences/Preferences.java
+++ b/app/src/cc/arduino/view/preferences/Preferences.java
@@ -174,7 +174,8 @@ public class Preferences extends javax.swing.JDialog {
     jTabbedPane1.setRequestFocusEnabled(false);
 
     sketchbookLocationLabel.setText(tr("Sketchbook location:"));
-
+    sketchbookLocationLabel.setLabelFor(sketchbookLocationField);
+    
     sketchbookLocationField.setColumns(40);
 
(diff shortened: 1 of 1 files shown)


facchinm:
I absolutely LOVE this kind of contributions. Thank you very very much @coreyknapp .
If you could help us in finding the remaining inaccessible parts of the IDE (I bet there are a lot) it would be incredibly useful. Thanks again!

coreyknapp:
I actually have a few more updates planned, this was just low-hanging fruit that I could go ahead and knock out.

As of right now, the biggest accessibility obstacle is with the Board and Library Managers. There's two problems: The screen reader can't read what's inside the table cells, and the buttons inside the table cells cannot be accessed with the keyboard. I can fix the first problem pretty easily by taking advantage of Swing's accessibility tools but the second problem is rooted in how the table is being rendered. I don't think there's an easy solution.

facchinm:
Got it! Thank you very much for the effort. Let us know if there is anything we can do on the Board/Lib manager UI side to ease your development!

PR #6927, opened 2017-11-14
Use const pointers in Stream::find() and Stream::findUntil()

bblanchon:
Changed the signatures of `Stream::find()` and `Stream::findUntil()` to take const pointers, as they don't change the target.

| Old signature | New signature |
|----|---|
| `find(char *target)` | `find(const char *target)` |
| `findUntil(char *target, char *terminator)` | `findUntil(const char *target, const char *terminator)` |

All the overloads have been changed accordingly.

This should not cause any code to break.


bblanchon:
👋 Can someone have a look?

facchinm:
Hi @bblanchon  ,
Thanks!

PR #7029, opened 2017-12-16
Let serial monitor send newline by default

matthijskooijman:
The serial.line_ending value was not present in the default
preferences file, so the default was implicitly 0, meaning "No line
ending". Since users often do not realize that they can even choose a
line ending, and "no line ending" is only rarely useful, it seems better
to default to sending a newline. Using both CR & NL would be more
consistent with Stream.println on the Arduino side, but just sending a
newline is probably easier to parse on the Arduino side.

### DIFF
--- a/build/shared/lib/preferences.txt
+++ b/build/shared/lib/preferences.txt
@@ -272,6 +272,7 @@ serial.databits=8
 serial.stopbits=1
 serial.parity=N
 serial.debug_rate=9600
+serial.line_ending=1
 
 # I18 Preferences
 


matthijskooijman:
@cmaglie @facchinm, would you consider merging this PR? It's fairly simple, but offers a nice improvement for new users.

matthijskooijman:
Ah, that's probably a manual cherry-pick or rebase. I think github only detects manual merges if you push the exact same commit, not a rebased version of it. Great, so it's merged already, I'll close this one then.

PR #7036, opened 2017-12-19
Removed explicit check for true/false

arnavb:
The title says it all.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/Stream.cpp
+++ b/hardware/arduino/avr/cores/arduino/Stream.cpp
@@ -115,10 +115,10 @@ bool Stream::findUntil(char *target, size_t targetLen, char *terminator, size_t 
 {
   if (terminator == NULL) {
     MultiTarget t[1] = {{target, targetLen, 0}};
-    return findMulti(t, 1) == 0 ? true : false;
+    return findMulti(t, 1) == 0;
   } else {
     MultiTarget t[2] = {{target, targetLen, 0}, {terminator, termLen, 0}};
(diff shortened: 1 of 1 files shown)


arnavb:
Ok, should I make the changes on those files in this PR?

matthijskooijman:
The other cores live in different repositories, so they'd need different PRs.

@facchinm, @cmaglie, what's the plan for these shared files? I remember some plans to extract them into a shared space, but I'm not sure how far along those plans are? Regardless, what's the best course to take here? Preparing separate PRs, or just copying over the entire Stream.cpp file at some later point?

arnavb:
@matthijskooijman You said this was safe to merge a few months ago. Any chance of this actually happening? (Sorry to pester you)

matthijskooijman:
@arnavb, I'm not in a position to merge things, I usually only review stuff. You'll have to pester @cmaglie and @facchinm instead :-p

arnavb:
Well, since you've pinged them, that's good.

facchinm:
Hi @arnavb , sure, we are in the middle of a transition to https://github.com/arduino/ArduinoCore-avr so I'd prefer not to merge the PR here but to merge it there. Furthermore, as you can see from the "Print and Stream class" tag, it is not an hardware related change so it's going to be merged as part of Chainsaw project. Sorry for the confusion about all these repos but we are trying to provide on a more modern way to contribute to the project.

arnavb:
@facchinm Makes sense!  Thanks for the info!

PR #7064, opened 2017-12-31
Deprecate binary.h in favor of binary literals

cousteaulecommandant:
Replace macros in binary.h with a single enum.
Make all Bxxxx constants trigger a "deprecated" warning (if supported), 
indicating to use 0bxxxx instead (supported since C++14 and GCC 4.3).
Addresses issue #3561

(It seems that `__attribute__ ((__deprecated__))` on enum values is not supported until GCC 6.1; in this case the warning will not be displayed)


cousteaulecommandant:
I have split this PR into two commits:
1. Replace the macros with a single `enum`.
2. Add the `deprecated` attribute to these constants in order to promote the use of `0bxx` literals.

In my opinion, this PR could be pulled directly.  However, if there are plans to slowly move away from macros in the Arduino library, and you don't think it is a good idea to mark `Bxx` constants as deprecated yet, **I could make a separate PR** with just the first commit.

PS: the second commit will have no effect with the default settings, since all warnings are suppressed.
PS2: `Bxx` constants are not used anywhere in this repository, so there should be no risk of getting a lot of warnings when compiling a sketch with warnings enabled if that sketch doesn't use these constants.

facchinm:
Already applied in https://github.com/arduino/ArduinoCore-API/commit/8bcc446b918f8579a39d493a830668a14e6cf93e , closing here

PR #7124, opened 2018-01-13
[IDE] Add support for selectable, user-defined and pre-packaged themes contained in zip files

Mumfrey:
### Summary

As mentioned in #7115, this PR adds support for selectable, user-defined themes packaged in zip files. This is intended to make it easier for users to share their themes with others.

### Description

This PR adds the following features:

* Much like the previous PR, user-defined themes can be installed in a `theme` folder within the user's sketchbook folder. However unlike resources placed directly in the `theme` folder, zip-packaged themes are not automatically activated in the IDE. Instead, available theme zips are enumerated at startup (and when opening the preferences dialog) and can be selected via the preferences dialog.
* Zips are selected for candidacy if they contain a `theme.txt`, though the `theme.txt` itself can be empty if the theme only overrides other resources such as images.
* Zips can optionally contain a `theme.properties` file which is a standard Java properties file which currently supports the following keys:
  * `name` can be used to specify a display name for the theme (used in the properties dialog)
  * `version` can be used to specify a version string
  
  Defining a standard properties file makes it easy to add additional functionality in the future, such as defining supported IDE version ranges for a theme, update URLs, author information, etc. to be determined as required.
* Zip resources override default resources, but can themselves be overridden by resources placed in the `theme` folder directly. This means there's a simple heirarchy of selection for resources with user-defined resources being highest, zip-packaged resources next, and default resources being lowest.
* In addition to the `<sketchbook>/theme` folder, the default `theme` folder in the application directory is also scanned for zipped themes. This is intended to make it possible for the IDE to package some additional selectable themes in the standard distribution, for example high contrast theme for visually impaired, dark theme, etc.
* If the same theme file exists in both locations, the user can select either one. Selected file names are prefixed with a "namespace" when stored in the `preferences.txt` which determines the location of the file.

### Screenshots

This PR adds a single entry to the preferences dialog, which allows the theme to be selected:

Preferences dialog

### Notes

* I have only tested this functionality on Windows and Linux because I don't have access to a Mac computer. Accessing resources in the zip is achieved by compiling a URL which points to the file within the zip, the same URL format as used by the JRE for resources within jars. This should *hopefully* be fully supported across platforms (I already took care of the mangling required for windows paths) but may need some quick checking on a Mac just be sure.

* This PR alters the external contract of `Theme` to return a `Resource` instead of a `File` when a file is needed directly. Since `Resource` instances are aware of resources they override (via their `getParent()` method), future consumers can leverage the "supercedes" relationship of theme resources if they wish. An example of this is already provided in the `loadFromResource` convenience method provided for `PreferencesMap`, whereby the preferences are loaded "bottom up" allowing preferences in higher priority resources to override those specified by lower priority ones.

Any changes or comments just let me know.


facchinm:
@ArduinoBot build this please

DemonWav:
Tested on macOS High Sierra, version 10.13.2, works fine as far as I can tell.

Mumfrey:
Am I okay to close this PR since it seems like the changes were merged into the beta branch, github just hasn't autoclosed it.

facchinm:
Ah, yes, due to the rebase every information about the actual PR got lost but the commit is merged. Closing, thank you!

PR #7157, opened 2018-01-25
Added delete operators to comply with new gcc [C++17]

kpochwala:
Earlier, when compiled with `-fsized-deallocation` warning enabled, this file caused problems. I also added `noexcept` specifier. Now it  compiles with gcc 7.2 and `-std=gnu++17` without problems.

### DIFF
--- a/hardware/arduino/avr/cores/arduino/new.cpp
+++ b/hardware/arduino/avr/cores/arduino/new.cpp
@@ -34,3 +34,11 @@ void operator delete[](void * ptr) {
   free(ptr);
 }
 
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Is noexcept needed? We're always running with `-fno-exceptions`, so `noexcept` might not be needed, or even wrong (not sure)? There is some previous discussion on a related subject in #108.

Can you be a bit more explicit about what problems you were running in without this delete operator? Is this operator new in C++17, or is there something in C++17 that makes it needed while it was not before?

facchinm:
Hi @kpochwala ,
Thanks!

PR #7193, opened 2018-02-06
Add StringView

steve-weinrich:
Add the StringView class.  A StringView is a read-only, non-owning reference to a C-string or a String. 
Although the name is based on the C++17 std::string_view, this implementation parallels the String interface and ignores the std::string_view interface.  

I also have a sketch that contains validation code for this class, but I don't see anywhere to put it!


Martin-Kraus:
https://github.com/arduino/Arduino/commit/4ac1fd1672135a5ad9dbce2b6333ced7d8ad4d5e

steve-weinrich:
HI Martin,
I am still learning the process for changes.  I see that this was committed.  Should I create the WiKi documentation or am I misunderstanding the process?
Thanks,
Steve

Issue #7139, opened 2018-01-19
String relational operators are members.


BillyDonahue:
Member relational operators are no good because they cause the lhs and rhs to be treated differently. This causes problems such as making == appear non-commutative. Such as:

```c++
void setup() {
  String("hi") == "world";
  "world" == String("hi");
}
```

```
ketch_jan19a:2: error: expected ';' before ')' token
   String("hi") == "world");
                          ^
sketch_jan19a:3: error: no match for 'operator==' (operand types are 'const char [6]' and 'String')
   "world" == String("hi");
           ^
exit status 1
expected ';' before ')' token
```

It would be less user-surprising to bring these operators out of the class.

I propose defining them at namespace scope as symmetric pairs of operators, between a String on one side, and everything String accepts as an implicit conversion on the other side, per:
    http://en.cppreference.com/w/cpp/string/basic_string/operator_cmp

PR #7194, opened 2018-02-06
String relational operators

BillyDonahue:
The String class has an incomplete set of relational operators with
`const char*`. The 6 operators `==`,`!=`,`<`,`>`,`<=`,`>=` must be
defined between String and String, but also between String and `const
char*`.

The motivation is to treat left and right operands symmetrically.  str
== "hello" works, but swapping arguments to "hello" == str doesn't,
which is a potential user surprise.

As String is implicitly convertible from `const char*`, but this incurs
a "strdup". So it also makes sense to provide relational operators
directly with const char* to avoid the expensive promotion of const
char* to temporary String, just to do a string value comparison.

Resolves: #7139


BillyDonahue:
signed

PR #7364, opened 2018-03-22
Allow println to send `\n` and not `\r\n`

hmaarrfk:


matthijskooijman:
I'm not sure if adding this to the API really is a good idea (to avoid cluttering the API). Doing a manual `print("\n")` can already do what you want, of course.

hmaarrfk:
I understand not wanting clutter. I can reimplement this with a switch.

Maybe an option to choose between `\n` and `\r\n`.
```
Serial.setln("\n");
```
maybe?


I am working with large code bases many times new programmers where at times the difference betwen `\r\n` and `\n`. The code works, but the serial line endings are all over the place. 

Refractoring the code is hard because it is challenging to test hardware with automated tests.

Would you consider alternative ways to changing the line ending?

hmaarrfk:
@matthijskooijman I changed the implementation now to be less polluting. Default behaviour is the same, but it should be much easier to change the line endings if the user wants to.

let me know what you think

matthijskooijman:
This looks a lot better to me. I do still see a lot of whitespace changes in your PR, those must certainly be removed before merging.

One other question is the performance impact of this change. I wonder if the compiler is smart enough to  optimize the line ending variable away.

Finally, it's not my call to merge this, but @cmaglie's or @facchinm's.

hmaarrfk:
@matthijskooijman, Thanks for considering this. I will remove the whitespace if the PR is considered.

hmaarrfk:
I guess I found the new home for this code 

https://github.com/arduino/ArduinoCore-avr/pull/18

matthijskooijman:
Correct, thanks!

Issue #6395, opened 2017-06-19
Digital Potentiometer code example needs delay added


mlfergy:
https://www.arduino.cc/en/Tutorial/DigitalPotControl

The code needs a delay to work:

void digitalPotWrite(int address, int value) {
  // take the SS pin low to select the chip:
  digitalWrite(slaveSelectPin, LOW);
**delay(100);**
  //  send in the address and value via SPI:
  SPI.transfer(address);
  SPI.transfer(value);
**delay(100);**
  // take the SS pin high to de-select the chip:
  digitalWrite(slaveSelectPin, HIGH);
}

matthijskooijman:
Are you using the exact same chip as mentioned in that tutorial? Did you look up the amount of delay needed in the datasheet? 50ms sounds like a lot more than a typical chip would require.

mlfergy:
Exact same chip...  I edited the post.  It actually doesn't work reliably
unless delay is 100 ms

SimonePDA:
Fixed the sketch, thanks a lot for your testing.

PR #7404, opened 2018-04-02
Update DigitalPotControl.ino

SimonePDA:
Fixing SPI communication with a delay as pointe out in #6395


sandeepmistry:
This change should also be ported to other cores:
* SAMD: https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/SPI/examples/DigitalPotControl/DigitalPotControl.ino
* SAM: https://github.com/arduino/ArduinoCore-sam/blob/master/libraries/SPI/examples/DigitalPotControl/DigitalPotControl.ino
* ARC32: https://github.com/arduino/ArduinoCore-arc32/blob/master/libraries/SPI/examples/DigitalPotControl/DigitalPotControl.ino

PR #7422, opened 2018-04-05
Add clean button to SerialPlot

chromhelm:


facchinm:
Hi @chromhelm ,
thanks for the contribution! Could you cleanup it a bit?
Since you implement three functionalities (clear button, labels and configurable buffer) it would be better to have three different commits, hence removing f8db03e from the history and splitting c76a8f9 into two (or more) commits.

app/src/processing/app/SerialPlotter.java
```diff
@@ -287,19 +316,81 @@ public void message(final String s) {
...
       int validParts = 0;
+      int validLabels = 0;
       for(int i = 0; i < parts.length; ++i) {
-        try {          
-          double value = Double.valueOf(parts[i]);
+        Double value = null;
+        String label = null;
+        
+        // all commands start with #
+        if(parts[i].startsWith("#") && parts[i].length() > 1) {
```
facchinm: I'm not a big fan of hardcoded commands but I appreciate the idea. 
Probably a standalone commit (like "add CLEAR and SET_SIZE runtime commands to Serial Plotter") could do the trick and make discussion easier

chromhelm:
Hi
I made a new pull request for label implementation #7453

app/src/processing/app/SerialPlotter.java
chromhelm: Done in pull request #7461

madsdyd:
This PR has a label of "IDE 1.9.x Beta". What does this mean? 

I think the relationship between #7452, #7454 and #7461 is confusing. Does anyone know what the status is?

PR #7452, opened 2018-04-10
Add clear button to SerialPlotter

chromhelm:


app/src/processing/app/SerialPlotter.java
```diff
@@ -255,6 +256,8 @@ public SerialPlotter(BoardPort port) {
 
     messageBuffer = new StringBuffer();
     graphs = new ArrayList<>();
+
+    clearButton.addActionListener(ae -> {graphs.clear();});
```
sceniclife, line 260: You can argue to also reset xCount to 0 here as well.

PR #7459, opened 2018-04-12
Add multithreding support to HardwareSerial

slavaza:
This fix will allow use HardwareSerial driver in simple multithreading applications.

### DIFF
--- a/hardware/arduino/avr/libraries/Thread/README.md
+++ b/hardware/arduino/avr/libraries/Thread/README.md
@@ -21,7 +21,7 @@ See example for how to.
 
 `grab(..)`	implement mutual exclusion barrier lock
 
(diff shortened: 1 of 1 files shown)


facchinm:
https://github.com/arduino/Arduino/pull/7459/commits/583e7d1408f63e9cba0342a21c849565093ffc31 could make sense, while including a whole new library (Thread in this case) is not going to happen :slightly_smiling_face: 
If you could move the PR (or, better, only the first commit) to https://github.com/arduino/ArduinoCore-avr it would help, thanks!

slavaza:
Excuse me, please. I discovered strange unexpected editing error what I removed in commits to ArduinoCore-avr. My code works good and I planning use it for toys in future.

PR #7461, opened 2018-04-12
Serial plotter remote commands

chromhelm:
This adds remote command to set the size and clear the plot area.
The commands are send like this:
#CLEAR
#SIZE:<size>
size should be between 10 and 5000. This are hard coded limits for the plot size right now.

### DIFF
--- a/app/src/processing/app/SerialPlotter.java
+++ b/app/src/processing/app/SerialPlotter.java
@@ -258,9 +258,8 @@ public class SerialPlotter extends AbstractMonitor {
 
     serialRates.setMaximumSize(serialRates.getMinimumSize());
 	
-	clearButton = new JButton(tr("Clear output"));
+	  clearButton = new JButton(tr("Clear output"));
 	
-
     graphWidth = new JSpinner(new SpinnerNumberModel(
                                BUFFER_CAPACITY_DEFAULT, //initial value
                                BUFFER_CAPACITY_MIN, //min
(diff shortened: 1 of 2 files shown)


madsdyd:
Is there a consensus about adding "remote commands" (this patch include them)", vs. e.g. the ability to be able to ignore some lines? Traditionally, "#" is often used to indicate a comment in text based data files. An initial suggestion could be to expand the SerialPlotter protocol to 
* Ignore any line starting with "#"
* Interprent any line starting with "$" as a command to the serial plotter.

The reason for "$" being that it is often used indicate a command prompt.

madsdyd:
Status of the QA check is that it fails because 
* The method `setNewBufferCapacity` assigns to one of its parameters.
* Line 326 needs to be changed from `if (commands.equals("CLEAR")) {` to `if ("CLEAR".equals(commands)) {`

Both are trivial to fix

stefan123t:
Thanks for adding the Serial Plotter with Labels and Commands to the Arduino IDE!

* I am trying to find the Serial Plotter protocol documentation time and again. Here is [PDF documentation](https://github.com/chromhelm/Arduino/blob/cdd66b63f686ebe3535dd5fc5f6b9cc592ce8263/ArduinoSerialPlotterProtocol.pdf) by @chromhelm and here are the official [ markdown docs](https://github.com/arduino/Arduino/blob/master/build/shared/ArduinoSerialPlotterProtocol.md) as well as a suggestion by @per1234 to [enhance / update existing examples with labels](https://github.com/arduino/Arduino/pull/7453#issuecomment-469207719)

Regarding remote commands and comments as described in this PR: 

> Is there a consensus about adding "remote commands" (this patch include them)", vs. e.g. the ability to be able to ignore some lines? Traditionally, "#" is often used to indicate a comment in text based data files. An initial suggestion could be to expand the SerialPlotter protocol to
> 
>     * Ignore any line starting with "#"

* I agree that `#` is almost ubiquitious for comments along with optional `//` line comments and `/* */` multiline comments.

* The command line should be part of a comment, e.g. something like a shebang: `# $ CLEAR;` or `#% $ SIZE 5000,500;`

* There are feature requests to ignore lines #9462 and adding a filter / regex for plotting only specific interesting lines #10226 .

PR #7486, opened 2018-04-18
Allow "update" to be used for EERef's operator=

Artoria2e5:
As one of the new convenience features introduced in EEPROM 2.0, `EEPROM[x] = newval` stands out from the `put` function by not using an `update` behavior. This commit allows users to use the feature while saving EEPROM lifetime by using an `update` behavior instead. To enable it, use `#define EEPROM_REF_ASSIGN_MODE update` before including the Arduino header.

For disambiguation and ease of switching, an explicit `write` function is added to EERef. Like `update`, the separate function is accepted by the compiler with an implicit `inline`. As a result, it is unlikely to adversely impact performance by going through this function instead of a direct `eeprom_write_byte`.

forum-thread: https://forum.arduino.cc/index.php?topic=542365

### DIFF
--- a/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
+++ b/hardware/arduino/avr/libraries/EEPROM/src/EEPROM.h
@@ -25,12 +25,6 @@
 #include <avr/eeprom.h>
 #include <avr/io.h>
 
-//! The action associated with the EERef assignment operator.
-//! Possible values are "write" and "update", unquoted.
-#ifndef EEPROM_REF_ASSIGN_MODE
-#define EEPROM_REF_ASSIGN_MODE write
-#endif
-
 /***
     EERef class.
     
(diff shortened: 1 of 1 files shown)


Artoria2e5:
One currently pending change is the replacement of the get/put loop with corresponding avr-libc [`eeprom_OP_block` functions](https://www.nongnu.org/avr-libc/user-manual/group__avr__eeprom.html), and potentially the replacement of the `EERef::update` method with `eeprom_update_byte`. All these routines are just wrappers around `eeprom_write_r18` (body section of `eeprom_write_byte` after moving a register), so the increase in code size should be minimal. On the other hand this also means such a change would be mostly aesthetic.

facchinm:
Hi @Artoria2e5 ,
If you are still interesting in getting this merged could you reopen in https://github.com/arduino/Arduinocore-avr ? [Here](https://github.com/arduino/Arduino/tree/master/hardware#migrating-commits) you can find some hints about reapplying the patch to that repo easily.
Thanks!

Issue #6361, opened 2017-06-08
Linenumbers on by default


septillion-git:
I just noticed line numbers are not displayed by default. I think this is a mad thing! Line numbers are the most useful thing, especially for a newbie, to see where an error is. Yes, the line highlighting also helps but especially when the compiler returns multiple errors line numbers are so useful. Also to be able to point to a line when helping on the forum. Any reason not to turn them on by default?

PR #7536, opened 2018-04-30
Line Numbers

ayohenry:
Alternative solution to issue #6361, added keyboard shortcut to toggle line numbers on/off; ctrl + shift + L.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Arduino uses
 [avrdude](http://www.nongnu.org/avrdude/),
 [bossac](http://www.shumatech.com/web/products/bossa),
 [openOCD](http://openocd.org/)
-and code from [Processing](http://www.processing.org)
+and code from [Processing](https://www.processing.org)
 and [Wiring](http://wiring.org.co).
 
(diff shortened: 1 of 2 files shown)


app/src/processing/app/Base.java
```diff
@@ -1877,6 +1877,15 @@ public void handleFontSizeChange(int change) {
     PreferencesData.set("editor.font", StringUtils.join(pieces, ','));
     getEditors().forEach(Editor::applyPreferences);
   }
+  
+  public void toggleLineNumber() {
+    if (PreferencesData.getBoolean("editor.linenumbers")) {
```
spacegaier: Could be shortened to:
`PreferencesData.setBoolean("editor.linenumbers", !PreferencesData.getBoolean("editor.linenumbers"));`

spacegaier:
The option to toggle the line numbers does already exist in the preferences. I am not sure what the benefit is of moving that into the menu. Not sure about a scenario where you want to consistently switch them on and then off again.
Adding a shortcut on top, is then even going one step further which I am not sure about. It blocks the shortcut also for some more usuful feature in the future.

PR #7544, opened 2018-05-01
Use "{archive_file_path}" in link command

jantje:
In IDE 1.6.6 "{archive_file_path}"  was introduced as a FQN archive file name but in avr it is not used in the link command which will lead to failing build if archive_file_path is not defined as {build.path}/{archive_file}


jantje:
I also updated the documentation at https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification/_compare/4601bc4542eb69dbbd115df66136424bed038fa3

jantje:
As the hardware moved to a different place I created a issue there https://github.com/arduino/ArduinoCore-avr/issues/14

PR #7549, opened 2018-05-02
CLI: Add --version to CLI option

yohsuke:
I added to get the Arduino IDE version from the command line
It will allow to check easily if the new Arduino is already installed.

This feature makes it easier to build external systems linked to specific versions of Arduino.

1. I added `--version` action, which shows version name and exit
    1. Currently, VERSION_NAME_LONG (like `1.8.5`, `1.9.0-beta`, `1.8.6 Hourly Build XXX`, etc...) is used. Because I want to know its version number and stable/beta/hourly.
    2. Finish with `0`. Because it is `SUCCESSFLLY FINISHED`.
2. Updated man page.

### DIFF
--- a/arduino-core/src/processing/app/helpers/CommandlineParser.java
+++ b/arduino-core/src/processing/app/helpers/CommandlineParser.java
@@ -85,10 +85,6 @@ public class CommandlineParser {
           }
           libraryToInstall = args[i];
         }
-        if (a == ACTION.VERSION) {
-          BaseNoGui.showMessage("Arduino", BaseNoGui.VERSION_NAME_LONG);
-          System.exit(0);
-        }
         action = a;
         continue;
       }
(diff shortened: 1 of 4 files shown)


arduino-core/src/processing/app/helpers/CommandlineParser.java
```diff
@@ -84,6 +85,10 @@ public void parseArgumentsPhase1() {
           }
           libraryToInstall = args[i];
         }
+        if (a == ACTION.VERSION) {
+          BaseNoGui.showMessage("Arduino", BaseNoGui.VERSION_NAME_LONG);
+          System.exit(0);
```
matthijskooijman: I'm not sure I like this exit here, this bypasses any cleanup that would normally (perhaps) run after this function returns. Also, this function is expected to parse arguments only, not act on them. Also, passing `--version --verify` now just acts on `--version`, instead of showing an error that only one of them can be passed.

Better would seem to set `action` here, and act on it elsewhere.

yohsuke: You're right!
Thank you for your comment.

shepting:
@matthijskooijman Is this better with the changes?

Issue #5012, opened 2016-06-04
Autoformat ignores tabsize preference


scottdarch:
I have `editor.tabs.size=4` in my preferences.txt which works fine if I hit <tab> but the autoformatter (cmd+T on OSX) ignores this setting and uses 2 spaces instead.

Verified this bug on both 1.6.8 and 1.6.9

JAG0991:
I'm looking within 1.6.10. The Auto Format tool uses a different configuration file called formatter.conf. It will look for a user defined formatter.conf within the same directory as your preferences.txt before using the default that can be found in the arduino/lib/ folder. The default formatter recommends that you copy it to the same directory as your preferences.txt and then edit it. Inside the formatter.conf is "indent=spaces=2". Changing this to 4 should hopefully make the Auto Format indent size the same as your editor.tabs.size.

jacob-jarick:
JAG0991 set me on the correct path. I use tabs always and 8 spaces (dont hate).

Exit arduino.
Create formatter.conf in your arduino conf directory (same location as preferences.txt
Add: indent=tab=8 (or your preferred size) then save.
edit preferences.txt

set the following 2 lines:

editor.tabs.expand=false
editor.tabs.size=8

done.

really should be an option in the GUI but its no hassle to edit a few config files.

PR #7550, opened 2018-05-02
Added comment to address A-Style Autoformatter

Alogitt:
Addresses issue #5012 which discusses auto format and preferences.txt. This will help others find the correct file to edit for A-Style.

### DIFF
--- a/build/shared/lib/preferences.txt
+++ b/build/shared/lib/preferences.txt
@@ -24,8 +24,8 @@
 # You'll have problems running Processing if you incorrectly 
 # modify lines in this file.
 
-# DO NOT CHANGE THIS FILE if you want to edit autoformatter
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Sounds like a good addition. However, you now write "DO NOT CHANGE THIS FILE if you want to edit autoformatter", which suggests that `preferences.txt` contains autoformatter settings, but you shouldn't change them. Rather, I think that `preferences.txt` does not contain such settings, and it might be better to write something like "This file does not contain autoformatter settings. Instead... " etc.

If you want to update this PR, note that you can amend your commit and do a force push to replace the current contents of this PR, there is no need to create a new PR.

Alogitt:
Thank you for your suggestion! I have edited the file with the altered message. Is there anything else I should do?

PR #7551, opened 2018-05-02
Change error output color to brighter orange

Alogitt:
Addressed issue #7451 by changing the color of error output to a brighter orange color.

### DIFF
--- a/build/shared/lib/theme/theme.txt
+++ b/build/shared/lib/theme/theme.txt
@@ -29,7 +29,7 @@ console.font = Monospaced,plain,11
 console.font.macosx = Monaco,plain,10
 console.color = #000000
 console.output.color = #eeeeee
-console.error.color = #E34C00
+console.error.color = #FF8300
 
 # GUI - BUTTONS
 buttons.bgcolor = #006468


matthijskooijman:
Looks good. Care to add a screenshot of how this looks (and let's try to prevent too much bikeshedding :-p).

Alogitt:
Sure! I'm including a screenshot with three different colors for error output. The leftmost one is the original color, the middle is an intermediate color I tested, and the rightmost one is the final color I committed. 
comparison

facchinm:
@00alis what about it?

Alogitt:
Hello, I just went back to the issue request that requested this fix and realized that the problem was not the brightness of the orange color, but certain colors in general. I am including a picture with the error output color in white (what the person is now asking for). I originally did not stray from the color orange because I did not want to break the traditional look, but maybe the change to white could help make the IDE more accessible. 
ffffff

per1234:
I put this PR on hold because feedback indicates it does not at all resolve https://github.com/arduino/Arduino/issues/7451.

00alis:
I agree the colors are hard to read on certain screens, it is quite an hard task to make it work everywhere. On the [Web IDE](https://create.arduino.cc/editor) we are using the following palette:
Green (debug content, does not appear if you turn off Show Verbose output) #1da086
White (normal output)
Yellow (warning) #F1C40F
Red (error) #DA5B4A

Would you mind applying the same palette to your PR and let us know your thoughts?

screenshot 2018-05-09 11 49 08

matthijskooijman:
Note that the web IDE does not currently distinguish warnings and debug output, though I believe to some degree it could (for warnings/debug output generated by the IDE itself or arduino-builder, not for external processes like the compiler).

lmihalkovic:
I did this in my local IDE. It is a trivial task and not much more work to extend the coloring to the entire lines (actually easier than to paint only a small part of the message). The part that is highlighted navigates to the relevant part of the source code.. personally, I find it useful (motivated me to chase all the warnings in my code) 

screen shot 2018-05-03 at 8 54 42 am

Issue #5366, opened 2016-09-13
install.sh only creates a shortcut, doesn't actually install the IDE


ThellraAK:
I was trying to get arduino working from the terminal when I realized 

trying to call arduino from terminal only resulted in a message saying it isn't currently installed
```
~/arduino-1.6.11⟫ arduino
The program 'arduino' is currently not installed. You can install it by typing:
sudo apt install arduino
```

When you do it with ./ it works (other than the problem with X stuff)
```
~/arduino-1.6.11⟫ ./arduino
Picked up JAVA_TOOL_OPTIONS: 
Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: java.awt.HeadlessException: 
No X11 DISPLAY variable was set, but this program performed an operation which requires it.
        at java.awt.SplashScreen.getSplashScreen(SplashScreen.java:117)
        at processing.app.Base.<clinit>(Base.java:92)`
```
Even when I ran the install script
```
~/arduino-1.6.11⟫ sudo ./install.sh
Adding desktop shortcut, menu item and file associations for Arduino IDE... done!
```
```
~/arduino-1.6.11⟫ arduino
The program 'arduino' is currently not installed. You can install it by typing:
sudo apt install arduino
```

I also did this without the sudo and got the same results.

The workaround I found is adding the ./ into the script provided by [this manpage](https://github.com/arduino/Arduino/blob/master/build/shared/manpage.adoc)
```bash
#!/bin/bash
Xvfb :1 -nolisten tcp -screen :1 1280x800x24 &
xvfb="$!"
DISPLAY=:1 ./arduino "$@"
kill -9 $xvfb`
```

facchinm:
Fixed by #7593

PR #7593, opened 2018-05-12
Symlink the arduino binary

2E0PGS:
install.sh should symlink arduino into `/usr/local/bin/arduino` so arduino is accessible via users PATH.

Related issues:

https://github.com/arduino/Arduino/issues/5780

https://github.com/arduino/Arduino/issues/5366

### DIFF
--- a/build/linux/dist/install.sh
+++ b/build/linux/dist/install.sh
@@ -68,7 +68,9 @@ xdg_install_f() {
 
   # Add symlink for arduino so it's in users path
   echo "" # Ensure password request message is on new line
-  sudo ln -s ${SCRIPT_PATH}/arduino /usr/local/bin/arduino
+  if ! ln -s ${SCRIPT_PATH}/arduino /usr/local/bin/arduino; then
+      echo "Adding symlink failed. Hope that's OK. If not then rerun as root with sudo."
+  fi
 
   # Clean up
   rm "${TMP_DIR}/${RESOURCE_NAME}.desktop"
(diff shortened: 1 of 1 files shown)


facchinm:
@2E0PGS this would make the "installation" dependant on `sudo`, which I would avoid at all costs.
Maybe checking the UID and then symlinking the executable in `/usr/local/bin/` if we have permissions to do it could be a good solution.
This way, launching `./install.sh` as a user wouldn't install the symlink, while `sudo ./install.sh` would.

2E0PGS:
./install.sh would prompt for a password once it hits the symlink part if ran as non root currently.


⁣Peter Stevenson (2E0PGS)
http://www.m3pgs.co.uk​

2E0PGS:
OK I suppose I assumed they are in the SUDOers group. Ref want to access it from CLI I think most who install software expect to find it in path. Thats the easy way to launch software besides using menu entries.

How about making the whole script require root level privileges and then I add code to move the folder into /opt/ and generate symlink to usr/bin. That to me seems like a more sensible trend of installers.

Could add flags for user level install and system wide installs. one requiring root and one not.

2E0PGS:
Yea that's a good idea.

2E0PGS:
I will add the changes ASAP.

2E0PGS:
Added the suggested changes.

2E0PGS:
Did anything like this ever get implemented? I am about to install Arduino and this pull never got merged so wonder if I need to manually symlink still...

2E0PGS:
I did indeed have to symlink it manually still or use my branch. 

I think this is fine for merge if the error message from `rm` or `ln` is ok in addition to the custom one. Otherwise I will need to modify this pull request so it only shows our custom error message if they don't run using sudo.

PR #7631, opened 2018-05-24
Create README in hardware folder

polymetric1:
The contents of the `hardware/` folder were moved, but it wasn't obvious
straight away from looking at the repository. This makes it immediately
obvious for anyone looking for something in this folder.

I made sure to adhere to the contributing guidelines. Since this isn't a
major change, I didn't create an issue first.

### DIFF
--- a/hardware/README.md
+++ b/hardware/README.md
@@ -7,3 +7,40 @@ The official Arduino cores can now be found here:
  - https://github.com/arduino/ArduinoCore-avr
  - https://github.com/arduino/ArduinoCore-sam
  - https://github.com/arduino/ArduinoCore-samd
+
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Good idea! Perhaps it would be good to also link to the sam and samd cores as well? They were never in this folder, but people that do not know that might still look for them here?

matthijskooijman:
I've taken the liberty of updating this PR with my suggested change, as well as add some git instructions on how to merge commits to the new repo (not sure if we already had something like that in place from the SAM migration?).

@polymetric1 note that I did a *force push* to your `patch-1` branch, so you'll have to be a bit careful if you want to pull from it.

Issue #6216, opened 2017-04-25
Saving an example sketch doesn't update the window title


seth10:
Say you open up an example sketch, modify it to suit your needs, and now want to save it. You'll see the "Sketch is read-only" message and be prompted to save it in another location. When you do, the tab name changes, but the window title remains unchanged.

### Platform
Arduino 1.8.2 (Windows Store 1.8.3.0) on Windows 10 Home

### Steps to Reproduce
1. Launch the Arduino IDE
2. File → Examples → 01.Basics → Blink
3. File → Save
4. OK
5. Save as "FlashyFlash"

### Expected outcome
- Tab title changes from "Blink" to "FlashyFlash"
- Window title changes from "Blink | Arduino 1.8.2" to "FlashyFlash | Arduino 1.8.2"

### Observed outcome
- Tab title changes from "Blink" to "FlashyFlash"
- Window title remains "Blink | Arduino 1.8.2"

seth10:
### Related behavior
- Creating a new sketch and saving it _does_ update the window title from "sketch_apr25a | Arduino 1.8.2" to "newSketch | Arduino 1.8.2"
- Clicking the down arrow to the right of the tab bar and selecting Rename does _not_ update the window title
- Using File → Save As... to save an existing sketch under a new name _does_ update the window title

seth10:
Component: IDE user interface
OS: Windows
Type: Bug
Priority: Low

PR #7674, opened 2018-06-07
Fixed the bug that doesn't updates the title when an example sketch is saved

JxJxJxMxNxS:
Added a line to update the title when a example sketch is saved, also changed the updateTitle method from private to public 

Resolves: #6216


cmaglie:
Thanks!

PR #7828, opened 2018-07-27
Fix Keyboard and Mouse Examples

SConaway:


SConaway:
I haven't tested this yet, but I will today(hopefully).

SConaway:
It still won't build sucessfully, but I'm probably doing something wrong when building from source.

SConaway:
NVM. I was building for wrong board. I'm going to update the comments at the top of each sketch.

facchinm:
Hi @SConaway, 
why are these changes needed? Arduino build system doesn't distinguish between "local" and "system" include (for the sake of simplicity). Are you using another build system? I'm closing the PR in the meantime, will reopen if you have a valid reason for this change :wink:

SConaway:
Didn't know, but thought we should try to be consistent.

per1234:
I also think consistency is a good thing. I'd like to see this as part of an official Arduino code style guide followed in all of Arduino's code.

Someone with experience understands that, in this particular case, there is not much difference between the quotes and angle brackets syntax (though angle brackets is most correct). But for a beginner this inconsistency is confusing: *"Why does one example sketch use `#include "SPI.h"` but this other one uses `#include <SPI.h>`?"* That's a good thing for them to learn eventually, but not like this.

However, I don't think the reasoning behind the change was well presented here. "Fix Keyboard and Mouse Examples" is very vague and makes me think some code bug was fixed. "Use correct #include syntax in examples" would be much more helpful when browsing through a long commit history. Certainly it should not be done over four commits but it is possible to squash to one commit while merging a PR.

SConaway:
@per1234 I agree. What happened was I fixed the `#includes`, then I fixed the comment blocks at the top of each sketch, then I fixed the `ArduinoISP` example. 

@facchinm Are you going to reopen this?

PR #7888, opened 2018-08-15
Fix "Increase Font Size" keybindings

cmaglie:
This commits adds "CTRL SHIFT =" as keybinding to increase font size.
The old shortcut "CTRL +" is still active, BTW it's not always possible to produce it if the keyboard layout doesn't have the + key on the base layer.
This add the possibility to compose "CTRL +" when the keyboard has the "+" available as "SHIFT =", that seems to be very common in many layouts.

Some reference charts:
http://kbd-intl.narod.ru/english/layouts
http://mentalfloss.com/article/52483/6-non-qwerty-keyboard-layouts
https://www.terena.org/activities/multiling/ml-mua/test/kbd-all.html

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1394,12 +1394,8 @@ public class Editor extends JFrame implements RunnerListener {
 
     menu.addSeparator();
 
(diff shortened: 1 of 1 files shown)


facchinm:
Wow, works perfectly! Merging right now

PR #6404, opened 2017-06-21
Moving a sketch to another folder should move all files

facchinm:
Fixes #6402

PR #7904, opened 2018-08-20
Avoid performing non trivial operations on files

facchinm:
Fixes https://github.com/arduino/Arduino/issues/7843 as per @cmaglie comment.

Translations should be updated removing the second part of the information text.

RATIONALE:
* The IDE used to ask the user if s/he wanted to move a file inside the folder with the same name (on double click, for example)
* This approach only works for single file sketches; for anything more complicated, the file would be moved but not the whole project (causing compile failures)
* #6404 tried to fix this by moving ALL the files inside a folder with the same name (works if you downloaded a project from github as ProjectName-master)

Problem: most of the times .ino files downloads from the web are being saved in the Download folder, so the move operation copied the whole Downloads content into that folder :flushed: 

@cmaglie suggested the IDE shouldn't be allowed to do this kind of operations; this patch removes the automatic move and requires the user to create the folder and move .ino file inside it manually

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1754,46 +1754,15 @@ public class Editor extends JFrame implements RunnerListener {
       } else {
         String properParent = fileName.substring(0, fileName.length() - 4);
 
-        Object[] options = {tr("OK"), tr("Cancel")};
+        Object[] options = {tr("OK")};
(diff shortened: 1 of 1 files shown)


facchinm:
Argh, I was a bit in a hurry and pasted the wrong link...
I'm also expanding the rationale, just to make explicit where and why the change should be applied.
Thanks for checking anyway :wink:

cmaglie:
Follow up on #7909

PR #7909, opened 2018-08-22
Avoid copying whole directory on Open in all cases

facchinm:
Alternative to #7904, tries to distinguish between

* click on a single ino file downloaded from the internet (old behaviour applies, directory is created and the file is moved in it)
* click on an ino file that for some reason is part of a project with the wrong name (eg. it was downloaded as zip from github, so the containing folder names becomes "projectName-master"). In this case if the directory is renamed after prompting the user)
2018-08-22-112348_1920x1080_scrot

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1793,6 +1793,7 @@ public class Editor extends JFrame implements RunnerListener {
 
       } else {
         String properParent = fileName.substring(0, fileName.length() - 4);
+        File properFolder;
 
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@facchinm are we still interested in moving ahead with these changes?

facchinm:
Yes, since the behaviour as of today is wrong :slightly_smiling_face: 
But maybe the right place to apply this would be the cli (if it enforces the `dirName/dirName.ino` convention)

PR #7922, opened 2018-08-25
Gracefully handle libraries having unparsable version info

PaulStoffregen:
If any library has malformed version info in its library.properties file, the Arduino IDE fails to launch.  If run from the command line, or arduino-debug.exe on Windows, the user sees a very unhelpful error telling then only "Invalid version found", but not *which* library is causing the problem.

This patch adds a more helpful message, and falls back to version "0.0.1" to allow the IDE to launch.

To test for this problem, simply install any library, then edit its library.properties.  I tested with DMXSerial2, changing it to "version=1.02.0".  The IDE is unable to start, and the user gets stuck without any reasonable way to discover which file is causing the problem.

### DIFF
--- a/arduino-core/src/processing/app/packages/UserLibrary.java
+++ b/arduino-core/src/processing/app/packages/UserLibrary.java
@@ -149,6 +149,10 @@ public class UserLibrary {
 
     String declaredVersion = properties.get("version").trim();
     Version version = VersionHelper.valueOf(declaredVersion);
+    if (version == null) {
+      System.err.println("Can't parse library version in " + propertiesFile.toString());
+      version = VersionHelper.valueOf("0.0.1"); // version must not be null
+    }
 
     UserLibrary res = new UserLibrary();
     res.installedFolder = libFolder;


PaulStoffregen:
This problem can also be triggered if the user alters their sketchbook location from File > Preferences.

When they restart the IDE for the change to take effect, the IDE is unable to start if the user unwittingly chose any folder containing (perhaps old) libraries with unparseable version info.  Without the ability to launch the IDE, the user can't return to the Preferences dialog to restore their old sketchbook location.

cmaglie:
Already fixed in #7920. I think that should be merged and another IDE release made ASAP.

PaulStoffregen:
Wow.  If 1.8.7 is to come quickly, is there any change of merging the AARCH64 stuff and put an official but "experimental" download on the website?

PaulStoffregen:
Closing this one, since #7920 is the intended fix.

PaulStoffregen:
@cmaglie - Still thinking about a release asap?  Want any help?  I'd really like to see 1.8.7 soon since so many users are suffering from the IDE not being able to start up.

PR #7927, opened 2018-08-25
Update appdata.xml - release 1.8.6

gvarisco:
This commit adds release 1.8.6 (2018-08-23) in appdata.xml

### DIFF
--- a/build/linux/dist/appdata.xml
+++ b/build/linux/dist/appdata.xml
@@ -36,6 +36,7 @@
   <content_rating type="oars-1.1" />
 
   <releases>
+    <release date="2018-08-23" version="1.8.6"/>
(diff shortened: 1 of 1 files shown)


cmaglie:
Maybe this should be generated in some way?
Is so important to put release dates here?

gvarisco:
Indeed, we should find an automated/half-automated way to generate it. 

FWIW, the <releases> section is needed in order for Flathub and other systems relying on appdata.xml to display a version number. Without it, the version shows up as "Stable", which isn't very informative to users. Check [this](https://flathub.org/apps/details/cc.arduino.arduinoide) out to get an example on how it looks like.

cmaglie:
May I already add the next release 1.8.7 and push in master (even if we release in a couple of days)?

PR #7935, opened 2018-08-27
Fix OSX "slow" menus

### LINKED ISSUE #7924: Lag when changing to File/Tools tab on macOS
OS: macOS 10.13.6
IDE: 1.8.6
When changing between tabs, File tab causes lag. 

When changing from other tabs to File tab, it will lag about a second. 
Changing to Tools tab also causes lag, but it's much shorter, about 0.1 sec. 
Other tabs Edit/Sketch/Help tabs work fine.

facchinm:
JRE 8u161 made `JMenu.insert()` MUCH slower on OSX.
This exposed a bug lurking there for years; in fact the menu entries were rebuilt every time since `JMenu.getComponent()` returns an empty list.
The correct function is `JMenu.getMenuComponents()`
In the meantime, also remove `sketchbookMenu` and `examplesMenu` deletion on board change; only their content needs to be updated, not the container itself.

Fixes #7924

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -508,7 +508,7 @@ public class Editor extends JFrame implements RunnerListener {
         if (!components.contains(sketchbookMenu)) {
           fileMenu.insert(sketchbookMenu, 3);
         }
-        if (!components.contains(sketchbookMenu)) {
+        if (!components.contains(examplesMenu)) {
           fileMenu.insert(examplesMenu, 4);
         }
         fileMenu.revalidate();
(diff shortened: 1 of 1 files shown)


DrDiodac:
Thanks! 
Tomorrow I will do testing and let you know what happens with Teensy 3.2

Send from iPhone

PaulStoffregen:
If you find any project that only happens with Teensy, but doesn't occur with official Arduino boards, please report it on the [PJRC Forum](https://forum.pjrc.com/forums/4-Suggestions-amp-Bug-Reports), not here.  I'll reply there and work on a fix.  This tracker is only for Arduino issues.

PR #7938, opened 2018-08-28
make "Increase Font Size" menu item a newJMenuItemShift() item

Vipas-ana:
This displays the shift symbol, in the Edit menu next to "Increase Font Size", showing us that we need to hold down shift to increase font size on mac os

do we need to do a OSUtils.isMacOS()?

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -1240,7 +1240,7 @@ public class Editor extends JFrame implements RunnerListener {
 
     menu.addSeparator();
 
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @Vipas-ana ,
the problem with this patch is that it won't work as expected on keyboards with '+' sign on the first level (like Italian, german and spanish layout).
Right now, you are prompted to  type "CRTL + Plus" , which is achievable on English keyboard via "CRTL + Shift + Equal" , so I'd leave it as is.
Feel free to comment if you think I misunderstood :wink:

PR #7957, opened 2018-08-31
Java now requires MacOSX 10.8 Mountain Lion or later

### LINKED ISSUE #7945: MacOS 10.7 (Lion) no longer works starting with Java 8u162
Arduino 1.8.6 will not launch on MacOS 10.7 Lion, because a newer Java JRE is used which fails to start.  The startup splash screen appears briefly, then the IDE crashes and give no indication why it failed to start.

Here is the Java info.
https://bugs.openjdk.java.net/browse/JDK-8201414

If a JRE 8u162 or later continues to be used, these lines Info.plist should be updated to specify 10.8 as the minimum MacOS version.

    <key>LSMinimumSystemVersion</key>
    <string>10.7</string>

This will give a message to users explaining their OSX is too old, rather than the IDE briefly appearing and then crashing.

The main website download page should probably be updated from "10.7 Lion or newer".

Or an older JRE can be used to keep compatibility with 10.7.

PaulStoffregen:
Fixes #7945.

Users with 10.7 will see message, instead of having the IDE appear to start and then crash for no apparent reason.

mac10 8

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -402,7 +402,7 @@
             mainclassname="processing.app.Base"
             copyright="Arduino LLC"
             applicationCategory="public.app-category.education"
-            minimumsystemversion="10.7"
+            minimumsystemversion="10.8"
             highresolutioncapable="true">
 
       <runtime dir="${MACOSX_BUNDLED_JVM}"/>


facchinm:
@ArduinoBot build this please

facchinm:
@ArduinoBot build this please

PR #7958, opened 2018-08-31
Improve first use experience for Upload before Port selected

### LINKED ISSUE #7943: Improve first-use usability with better default port name
Perhaps the default port name should be changed to a string which users can understand means they have not yet selected a port?

Today the default is COM1.  Perhaps this dates all the way back to Arduino's origin, when the board actually had a 9 pin serial connector and most PCs still had physical 9 pin serial ports?  Today almost no PCs have 9 pin serial ports exposed, but it's still very common to have an unused 10 pin header on the motherboard which actually implements COM1.

When a first-time user tries to upload to their Uno, Nano, or other AVR-based board, they see this rather unhelpful error:

**avrdude: ser_open(): can't open device "\\.\COM1": The system cannot find the file specified.**

Or...

**Using Port                    : COM1
Using Programmer              : wiring
Overriding Baud Rate          : 115200
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_ReceiveMessage(): timeout
avrdude: stk500v2_getsync(): timeout communicating with programmer**

Ideally the first-time user experience should be a helpful message which explains they need to click Tools > Ports to select the their board.  If the default name were something which could not possibly be a valid port on any OS, then the IDE could check for that setting before running AVRDUDE or other upload tools.

PaulStoffregen:
Fixes #7943

When running the IDE for the first time, this appears if Upload is attempted before selecting a Port.

capture

### DIFF
--- a/arduino-core/src/cc/arduino/UploaderUtils.java
+++ b/arduino-core/src/cc/arduino/UploaderUtils.java
@@ -55,11 +55,6 @@ public class UploaderUtils {
         return null;
       }
       boardPort = BaseNoGui.getDiscoveryManager().find(port);
-      //if (boardPort == null) {
-        // Is there ever a reason to attempt upload when
-        // the Port is not found by DiscoveryManager?
-        //return null;
-      //}
     }
 
     return new UploaderFactory().newUploader(target.getBoards().get(board), boardPort, noUploadPort);
--- a/arduino-core/src/cc/arduino/UploaderUtils.java
+++ b/arduino-core/src/cc/arduino/UploaderUtils.java
@@ -50,7 +50,16 @@ public class UploaderUtils {
 
     BoardPort boardPort = null;
     if (!noUploadPort) {
-      boardPort = BaseNoGui.getDiscoveryManager().find(PreferencesData.get("serial.port"));
+      String port = PreferencesData.get("serial.port");
+      if (port == null || port.isEmpty()) {
+        return null;
+      }
+      boardPort = BaseNoGui.getDiscoveryManager().find(port);
+      //if (boardPort == null) {
+        // Is there ever a reason to attempt upload when
+        // the Port is not found by DiscoveryManager?
+        //return null;
+      //}
     }
 
     return new UploaderFactory().newUploader(target.getBoards().get(board), boardPort, noUploadPort);
--- a/app/src/processing/app/SketchController.java
+++ b/app/src/processing/app/SketchController.java
@@ -709,6 +709,10 @@ public class SketchController {
 
     UploaderUtils uploaderInstance = new UploaderUtils();
     Uploader uploader = uploaderInstance.getUploaderByPreferences(false);
+    if (uploader == null) {
+      editor.statusError(tr("Please select a Port before Upload"));
+      return false;
+    }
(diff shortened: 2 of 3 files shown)


arduino-core/src/cc/arduino/UploaderUtils.java
```diff
@@ -50,7 +50,16 @@ public Uploader getUploaderByPreferences(boolean noUploadPort) {
...
-      boardPort = BaseNoGui.getDiscoveryManager().find(PreferencesData.get("serial.port"));
+      String port = PreferencesData.get("serial.port");
+      if (port == null || port.isEmpty()) {
+        return null;
+      }
+      boardPort = BaseNoGui.getDiscoveryManager().find(port);
+      //if (boardPort == null) {
+        // Is there ever a reason to attempt upload when
+        // the Port is not found by DiscoveryManager?
+        //return null;
```
matthijskooijman: Good question. It would be interesting to investigate in what cases (if at all) `find(port)` returns `null` (I suspect it does not return `null` when an invalid serial port is given, e.g. `COM1` on Linux, since that tries to run avrdude, so perhaps this is only done with network port stuff, if at all?). It is also interesting to investigate what `newUploader` below does with a null `boardPort`.

In any case, I think a `null` `boardPort` should not return `null` here, since that shows a "Please select a port" message, which is not appropriate when the user has selected a port, but it is not available or otherwise cannot be used.

Before merging, I think at least these commented lines should be removed, even if no more elaborate handling of this case is added.

PaulStoffregen: I tried with COM1 on Linux.  Indeed boardPort is null (verified by adding a println).  Then avrdude is run.  The result is this error:
sc

PaulStoffregen: But perhaps it's best to allow the uploader to run anyway?  Maybe some boards have uploaders that can auto-detect the board, even when given a port name that doesn't exist?

matthijskooijman: Oh, that is a bit unexpected. I would think that, since `boardPort` is `null`, the actual port name (`COM1`) is not passed into `newUploader`, so it could not pass it to `avrdude`. However, I guess the port name is read from the preferences separately when building the avrdude recipe, so the `boardPort` argument here serves a different purpose?

PaulStoffregen: Honestly, I didn't dig deeper into how it's really using boardPort.  That's the reason for the question!

But now that we're discussing this, I agree it's probably best to just allow the uploader to run even if boardPort is null.  I would worry about breaking 3rd party uploaders.

The point here was to just require first-time users to select *something* from the Ports menu.

PaulStoffregen: Commented stuff deleted.

Issue #7937, opened 2018-08-27
Corrupted library_index.json prevents IDE from starting


PaulStoffregen:
If library_index.json becomes corrupted, the IDE can't start up and throws a NullPointerException.

    Error parsing library.index:Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
     at [Source: (FileInputStream); line: 1, column: 2]
    java.lang.NullPointerException
            at cc.arduino.contributions.libraries.LibrariesIndexer.rescanLibraries(LibrariesIndexer.java:140)
            at cc.arduino.contributions.libraries.LibrariesIndexer.setLibrariesFolders(LibrariesIndexer.java:109)
            at processing.app.BaseNoGui.onBoardOrPortChange(BaseNoGui.java:715)
            at processing.app.Base.onBoardOrPortChange(Base.java:1313)
            at processing.app.Base$12.actionPerformed(Base.java:1540)
            at processing.app.Base.filterVisibilityOfSubsequentBoardMenus(Base.java:1579)
            at processing.app.Base.access$000(Base.java:83)
            at processing.app.Base$11.actionPerformed(Base.java:1512)
            at processing.app.Base.rebuildBoardsMenu(Base.java:1490)
            at processing.app.Base.<init>(Base.java:267)
            at processing.app.Base.main(Base.java:145)

To reproduce this problem, just edit ~/.arduino15/library_index.json (or similar location on Mac & Windows) and change the first character from '{' to '<'.  Restart the IDE will give the NullPointerException.

This problem came today for a user on the forum.

https://forum.arduino.cc/index.php?topic=565681

PaulStoffregen:
Quick followup, I got 2 similar NPE start failures if the change the permissions on the file and the directory.

chmod 000 ~/.arduino15/library_index.json

    Picked up JAVA_TOOL_OPTIONS: 
    java.lang.NullPointerException
	    at cc.arduino.contributions.packages.ContributionsIndexer.parseIndex(ContributionsIndexer.java:103)
	    at processing.app.BaseNoGui.initPackages(BaseNoGui.java:504)
	    at processing.app.Base.<init>(Base.java:262)
	    at processing.app.Base.main(Base.java:145)

chmod 000 ~/.arduino15/

    Picked up JAVA_TOOL_OPTIONS: 
    java.lang.NullPointerException
[...]

PR #7964, opened 2018-09-03
Fix crash due to index json parse errors and permission problems

cmaglie:
See #7937


PaulStoffregen:
Here's the library manager window after it has replaced library_index.json with a good copy.

librarymanager

The library list actually has been updated, but a user looking only at the GUI and not the file in the hidden folder would believe it didn't do anything.

cmaglie:
Yes, it has to re-open the window, I'm trying to fix this.

cmaglie:
Pushed other two commits, should fix it for good

Issue #8021, opened 2018-09-21
Feature Request: Don't close app when every sketch is closed


Jones-S:
Dear Arduino Devs,
This is a thing that bothered me since years. I finally make the effort to ask you to improve your software: I really dislike the behaviour that the application closes whenever I close all my sketches. I sometimes close them accidentally and after taht I have to wait quite a while to restart the app.
Would it be possible to adapt the behaviour like we all know it from other apps. 
chrome .e.g would not quit, just because I close all my windows...

This would be a great improvement and I know a lot of people who actually think the same.
Thank you for sharing your thoughts about this.

cheers

J

facchinm:
Hi @Jones-S ,
would you mind testing https://github.com/arduino/Arduino/pull/8023 when the build is ready? It's not yet bug free but it could be a beginning.

Jones-S:
Oh that was quick! I downloaded the mac build and opened the app. So far it does what I want it to. It stays open. 👏🏼 for that.
Unfortunately I'm currently abroad and I don't have any arduinos with me. anyway I found a first bug: When no sketch is open (but the app is) and I want to open up a recent item: `File` -> `Open recent`
It does not open the file. The command only works if there is already a sketch open.
Should I report this as a new issue?
Cheers

facchinm:
@Jones-S I'm closing the issue as duplicate of #936 (which contains more information).
In the meantime I should have fixed the remaining bugs you found so I'd love if you could test the new build as soon as it is available  on #8023 . Thanks!

PR #8023, opened 2018-09-21
[macOS] Modify behaviour when last editor gets closed

facchinm:
Fixes https://github.com/arduino/Arduino/issues/8021 and #936

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -49,6 +49,7 @@ import processing.app.debug.TargetBoard;
 import processing.app.debug.TargetPackage;
 import processing.app.debug.TargetPlatform;
 import processing.app.helpers.*;
+import processing.app.helpers.OSUtils;
 import processing.app.helpers.filefilters.OnlyDirs;
 import processing.app.helpers.filefilters.OnlyFilesWithExtension;
 import processing.app.javax.swing.filechooser.FileNameExtensionFilter;
@@ -931,15 +932,14 @@ public class Base {
     }
 
     if (editors.size() == 1) {
-
+      editor.setVisible(false);
+      editors.remove(editor);
       handleQuit();
-
     } else {
       // More than one editor window open,
       // proceed with closing the current window.
       editor.setVisible(false);
       editor.dispose();
-
       editors.remove(editor);
     }
     return true;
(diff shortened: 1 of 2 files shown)


app/src/processing/app/Base.java
```diff
@@ -921,7 +922,10 @@ public boolean handleClose(Editor editor) {
       storeSketches();
 
       // This will store the sketch count as zero
+      editor.setVisible(false);
+      //editor.dispose();
```
sandeepmistry: Let's remove the commented code :)

app/src/processing/app/macosx/ThinkDifferent.java
```diff
@@ -45,6 +47,20 @@
...
+    application.addAppEventListener(new AppReOpenedListener() {
+      @Override
+        public void appReOpened(AppReOpenedEvent aroe) {
+          try {
+            if (Base.INSTANCE.getEditors().size() == 0) {
+              Base.INSTANCE.handleNew();
+            }
+          } catch (Exception e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
sandeepmistry: Better error handling here?

gvarisco:
@sandeepmistry same here, despite I've noticed that if you have multiple tabs and decide to close the first one, the app "goes away" (as in: stays in background, do not kill itself, but tabs are gone if you click again on it) instead of showing the other tabs kept open.

app/src/processing/app/Base.java
facchinm: I left it there for a reason :) On my trials, if dispose() gets called the toolbar doesn't disappear but clicking on any element has no effect. I believe it's a broader bug, since the toolbar handlers are declared inside the first Editor instance. If we refactor the code dispose() should be called, so it's more like a reminder :)

app/src/processing/app/macosx/ThinkDifferent.java
facchinm: Absolutely (no idea about how to handle it anyway)

facchinm:
@gvarisco can you expand on that? The "new" behaviour should only kick in if a single Editor windows is open (multiple tabs on the same editor don't count they are part of an "atomic" sketch).
At reopen I asked @akash73 what is the expected behaviour for OSX users and we agreed to implement "new empty sketch" instead than "reopen last sketch". It can be changed of course :wink:

gvarisco:
Clarified with @facchinm - all good on my side. LGTM.

Issue #7942, opened 2018-08-28
Feature for a platform to add its discoverer to DiscoveryManager


PaulStoffregen:
The IDE should provide a configurable way for platforms to provide their own discoverers to the IDE's DiscoveryManager.

A recipe in platform.txt would allow a platform to specify a shell command to run for its discoverer.

Communication between the shell command and the IDE should use stdin/stdout with text in JSON format.  Two fields would be mandatory, "id" for the unique port identifier to be used in the upload recipe, and "label" for the text to be shown in the Ports menu.  Optional fields could provide information about the board, or FQBN, or partial QBN, depending on how much information about the hardware is known to the discoverer.

PaulStoffregen:
I started experimenting with parsing JSON as it arrives from a process's stdout stream.  Using Jackson's ObjectMapper with the BoardPort class seems like the most straightforward way.  The JSON would look like this when a board is connected:

    {
      "address": "/dev/ttyACM0",
      "online": true,
      "label": "/dev/ttyACM0 (Arduino Uno)"
    }

If the board is removed, the JSON would look like this:

    {
      "address": "/dev/ttyACM0",
      "online": false
    }

I didn't choose any of this... it's the natural result of using ObjectMapper and BoardPort.  Hopefully that's ok?

cmaglie:
(I was just looking at this)

The `BoardPort` class at the moment is:

```
public class BoardPort {

  private String address;
  private String protocol;
  private String boardName;
  private String vid;
  private String pid;
  private String iserial;
  private String label;
  private final PreferencesMap prefs;
  private boolean online;
```

but most of these fields are tied to a specific Discovery (for example `vid` and `pid` are specific for `SerialDiscovery`, instead `online`, `address`, `protocol` are for `NetworkDiscovery`. This can't work if we want to make it modular and the number of `xxxDiscovery` and relative field may potentially grow.

IMHO we should leave only "label", "id" and "type/protocol", all the other values should be part of the `prefs` map.

cmaglie:
> I started experimenting with parsing JSON as it arrives from a process's stdout stream

Did you have already implemented a `CommandLineDiscovery` or something like that? would you mind sharing it?

PaulStoffregen:
Sure, with the caveat this is early & unfinished, but it does work.

https://github.com/PaulStoffregen/Arduino-1.8.6-Teensyduino/blob/master/arduino-core/src/cc/arduino/packages/discoverers/TeensyDiscovery.java

[...]

PR #8038, opened 2018-09-28
Pluggable discovery: search in platform.txt (WIP)

cmaglie:
/cc @PaulStoffregen 
this is very WIP, but should be a good starting point.

This PR changes the `DiscoveryManager` so it search for custom discovery in the platform.txt as explained in https://github.com/arduino/Arduino/issues/7942#issuecomment-424634831, you can try it by editing an existing `platform.txt`.

The DiscoveryManager then runs a `PluggableDiscovery` that is a stub for now but, once implemented, it should start the custom discovery tool and handle all the communication with it. If you would like to try to implement this it would be great, I'll be 100% on IDE and CLI next week (so hopefully we should make good progress on this).


PaulStoffregen:
Thanks.  I will work on a PluggableDiscovery implementation over the weekend.

How to push commits to this pull request is probably well beyond my limited knowledge of git & github, but I can certainly work on the Java code and update the SerailDiscovery tool.

facchinm:
@ArduinoBot build this please

PaulStoffregen:
I started an implementation of PluggableDiscovery.java.  Here is the code so far:

https://github.com/PaulStoffregen/Arduino/blob/acb22b5d644ca44eca5e3a540c88d7fd68de8cfe/arduino-core/src/cc/arduino/packages/discoverers/PluggableDiscovery.java

Main question: Should I create a new class for the JSON parsing, rather than using BoardPort?  And add code to translate it to BoardPort for use by DiscoveryManager?  Or is the idea to change BoardPort to match the desired format?

Here's a quick screenshot with PluggableDiscovery.java using [SerialDiscovery_JSON](https://github.com/PaulStoffregen/SerialDiscovery_JSON):

screen

PaulStoffregen:
@cmaglie - Should I add another class to use as the JSON format?

arduino-core/src/cc/arduino/packages/discoverers/PluggableDiscovery.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+    write("START_SYNC\n");
+    pollingThread = null;
+  }
+
+  private void startPolling() {
+    // Discovery tools not supporting START_SYNC require a periodic
+    // LIST command.  A second thread is created to send these
+    // commands, while the run() thread above listens for the
+    // discovery tool output.
```
cmaglie: Not sure if this is the best way: we can call "LIST" just once directly in `listDiscoveredBoards()` (basically it will poll only when the Tools menu is triggered and not every 2.5 sec). On the other hand this may add some delay on opening the menu unless there is some internal cache in the discovery tool.

cmaglie:
Very nice!

> Should I add another class to use as the JSON format?

you mean to allow passing some more metadata?
yes, something like `PluggableDiscoveryMessage` should do I think.

arduino-core/src/cc/arduino/packages/BoardPort.java
```diff
@@ -47,6 +47,18 @@ public BoardPort() {
     this.prefs = new PreferencesMap();
   }
 
+  public BoardPort(BoardPort bp) {
+    prefs = new PreferencesMap();
+    // TODO: copy bp.prefs to prefs
```
cmaglie: `new PreferencesMap(bp.prefs)` should do the job

PaulStoffregen:
> you mean to allow passing some more metadata?
> yes, something like PluggableDiscoveryMessage should do I think.

Two issues.

1: Yes, metadata for the event type "add", "remove", "update" (or just "add" again with a label/address already used), "error" (currently only START_SYNC not supported).  But this could be also be done easily with just 1 extra variable in BoardPort.

2: Currently BoardPort is nothing like the JSON format you described.  Adding another class could give the JSON format you want without changes to BoardPort, but extra code & complexity.  Using only BoardPort is simpler, but BoardPort is widely used by the rest of the IDE, so many edits needed for any changes to BoardPort.

cmaglie:
At the moment BoardPort is defined as:

```java
  private String address;
  private String protocol;
  private String boardName;
  private String vid;
  private String pid;
  private String iserial;
  private String label;
  private final PreferencesMap prefs;
  private boolean online;
```

I'd like to change it and move "fixed" fields like `vid` and `pid` inside `prefs` (and change where they are used with the corresponding map get).
For the `online` field I'm unsure I need to see where it is used and if it can be removed.
In the end the BoardPort should be something like:

```java
  private String protocol;
  private String address;
  private String label;
  private final PreferencesMap prefs; // should be really called `props`, but let's keep it as is for now
  private final PreferencesMap identificationPrefs;
  // private boolean online; // Maybe?
```

If we reach this point we can change also the board detector so it can match the board model using the `identificationPrefs` and the `boards.txt` content.

Does the Teensy hid use some identifiers similar to `vid` and `pid`?

PaulStoffregen:
> I'd like to change it and move "fixed" fields like vid and pid inside prefs (and change where they are used with the corresponding map get).

I'll start working on this.  Do you want it to be on this pull request?

> If we reach this point we can change also the board detector so it can match the board model using the identificationPrefs and the boards.txt content.

I do not have a clear understanding how you want identificationPrefs to work.  My guess is you're imagining adding a separate class or package for board detection?  Maybe you're thinking identificationPrefs info would be the *input* to this board detection class?

So far, I have understood the board detection process happens within each discoverer instance.  Both SerialDiscovery and NetworkDiscovery have this detection process built in.  So I have imagined any board detection logic for PluggableDiscovery would occur within the external discovery tool and be communicated to the PluggableDiscovery instance as JSON fields.

I'm imagining identificationPrefs would be the *output* of board detection logic.  Rather than leaving the rest of the IDE to struggle with how VID, PID, HID descriptors, bluetooth info and other arbitrary info matches, all that info-to-board logic would be inside each discoverer.  Then identificationPrefs would have fields the rest of the IDE could readily use without guesswork, like the board name.

Before I do any work with identificationPrefs, I need to clearly understand how you want it to work.  Where the properties-to-identification logic will exist is the critical decision.

PaulStoffregen:
> In the end the BoardPort should be something like: ...

I added a commit to move closer to this goal.  Moving vid, pid, serial number into prefs is easy, because everything uses the getter & setting functions.  I've updated [SerialDiscovery_JSON](https://github.com/PaulStoffregen/SerialDiscovery_JSON).  Now it looks like this:

```javascript
{
  "eventType": "add",
  "address": "_/dev/ttyUSB0",
  "label": "/dev/ttyUSB0 (FT232R USB UART)",
  "boardName": "FT232R USB UART",
  "prefs": {
    "vendorId": "0403",
    "productId": "6001",
    "serialNumber": "A800I6T1"
  },
  "protocol": "Serial Device"
}
{
  "eventType": "add",
  "address": "_/dev/ttyACM0",
  "label": "/dev/ttyACM0 (USB Serial)",
  "boardName": "USB Serial",
  "prefs": {
    "vendorId": "16c0",
    "productId": "0483",
    "serialNumber": "3990740"
  },
  "protocol": "Serial Device"
}
```

PluggableDiscovery is no longer using the "online" boolean.  But it seems to be used by SerialBoardsLister.  I don't fully understand it, but the use appears to be related to making a port inaccessible during upload, and perhaps handling the case where the same board reappears as a different port shortly after the upload completes.  Pretty sure this can't be removed from BoardPort without breaking stuff.

Not sure what to do with boardName.  Need your opinion on how you want identificationPrefs to work and where you envision the properties-to-board logic implemented.  My hope for PluggableDiscovery would be to put that logic into the discovery tool and send the board info as JSON fields.  But if you're planning to add another package or some other layer to do that stuff, need to understand how you want it to work.

cmaglie:
> I do not have a clear understanding how you want identificationPrefs to work. My guess is you're imagining adding a separate class or package for board detection? Maybe you're thinking identificationPrefs info would be the input to this board detection class?

exactly, let me expand a bit on this one, the idea is really simple:

- each board should have the discovery information written in the boards.txt as generic properties, for example the Arduino Uno has:

```
uno.name=Arduino/Genuino Uno
uno.vid.0=0x2341
uno.pid.0=0x0043
uno.vid.1=0x2341
uno.pid.1=0x0001
uno.vid.2=0x2A03
uno.pid.2=0x0043
uno.vid.3=0x2341
uno.pid.3=0x0243
```

- each discovery will report the discovered port with all the identification information, for example (slightly changing the output of your Serial Discovery) a report may be:

```
{
  "eventType": "add",
  "address": "_/dev/ttyACM0",
  "label": "/dev/ttyACM0 (USB Serial)",
  "boardName": "USB Serial",
  "prefs": {
    "vendorId": "0x2341",
    "productId": "0x0043",
    "serialNumber": "3990740"
  },
  "identificationPrefs": {
    "vid": "0x2341",
    "pid": "0x0043"
  }
  "protocol": "Serial Device"
}
```

- the hypothetical `BoardPortMatcher` will scan all `boards.txt` searching for the board that match the `identificationPrefs`. In other words, looking at the example above, the matcher will search for a board that has the propertes `vid.x` and `pid.x` set respectively to `0x2341` and `0x0043` (in this case the Arduino Uno will be matched).

- this example has `vid` and `pid` but it really may be anything... `macaddressprefix`, `mdsntype`... `oranges` and `apple` :-) it just needs the discovery to provide these info and the matcher will do the rest.

> So far, I have understood the board detection process happens within each discoverer instance. Both SerialDiscovery and NetworkDiscovery have this detection process built in. So I have imagined any board detection logic for PluggableDiscovery would occur within the external discovery tool and be communicated to the PluggableDiscovery instance as JSON fields.

Doing this will not scale well, for example the discovery will not match 3rd party boards (for example the serial discovery will match only Arduino boards because they are the only one known to him), unless we find a way to pass the content of each installed `boards.txt` to the discovery tool, but this seems to complicate things IMHO.

PaulStoffregen:
Oh, ok, I believe I understand now.  Will try to add this later today.

PaulStoffregen:
I added identificationPrefs and a function which searches for a board which matches all the identificationPrefs fields.  Hopefully this is close to what you were thinking?

So far, it's only actually being used within PluggableDiscovery to set the board name.

facchinm:
@ArduinoBot build this please

arduino-core/src/cc/arduino/packages/discoverers/PluggableDiscovery.java
PaulStoffregen: We already do have the internal cache as the portList of discovered ports in PluggableDiscovery.  listDiscoveredBoards() just returns a copy of the list.  At least that part is easy.

To make this work, I believe DiscoverManager would need to give all discoverer instances an early notice that listDiscoveredBoards() may soon be called.  Maybe this could be done by detecting mouse hover over the Tools menu?  Or maybe mouse-over any part of the menus?

Early versions of Teensy's software waited in listDiscoveredBoards().  The result was usually ok, but some Windows & Mac users had a terrible experience with the GUI becoming unresponsive.  I believe it's best to always quickly return the list of previously discovered ports.

PaulStoffregen:
@cmaglie - Is there anything more you wanted to see done here?  Other than when to send "LIST", not sure what more I can do here?

Maybe getting close to being able to merge this?

cmaglie:
The algorithm is correct, but I'd like to move it away from the BoardPort class.
BTW I'm on it now, will report back soon.

cmaglie:
The `SerialDiscovery_JSON` repo is updated with the version you used to test?

PaulStoffregen:
Yes, SerialDiscovery_JSON is up to date.  The last change added identificationPrefs, so if you see that in the JSON output you can be sure you've got the latest.

cmaglie:
Ok I've slightly refactored some parts:
- completely removed fixed methods in BoardPort for VID/PID/etc. we can just get these values using get("vid") instead of getVid() or whatever.
- fixed the board detection algorithm, becuase it searched for `vid` and not `vid.N` (like `vid.0`, `vid.1`...)
- also fixed the display of the ports list, now the field `BoardPort.boardName` is used to populate the GUI instead of tweaking the label.

This changes allows your SerialDiscovery_JSON prototype to work and list the ports as the native SerialDiscovery java class.

Even if not strictly required for this PR, I've also merged the former SerialDiscovery class with the SerialBoardLister class, since the former was acting just like a proxy for the latter. This is another step toward a uniform disocvery plugin architecture.

On the discovery side, I'm quite satisfied with the current PR, now I'd like to explore how this extends to the uploading/programming part and the monitor/terminal. I've some ideas, will sketch some design draft later today.

PaulStoffregen:
Do you want to expand the scope of this PR to the serial monitor?  Or do that separately?

PaulStoffregen:
@cmaglie - Would you like me to do anything more on this?

PaulStoffregen:
@cmaglie - Can this be merged?  Even though more work will be needed later in the serial monitor, at least finalizing the discovery part can allow it to serve as a template for pluggable discover in the CLI.

I would like to be able to integrate Teensy with the CLI, but lack of pluggable discovery is a blocking issue.

cmaglie:
I've rebased this PR and I'm going to merge it, but I want to add a warning that informs this feature may very likely change in the future.

PaulStoffregen:
Ok.

If the JSON format is expected to change in the future, maybe the "START_SYNC" & "START" messages could communicate version if or some other way for the discovery tool to know which format is expected?

cmaglie:
We could change the command to `START_V2`.
BTW isn't better to just to do a breaking change and say that the first real supported version is, for example, IDE >= 1.8.9? What I'm trying to avoid is to avoid people start using this feature too much and having problems later if we change something... but maybe I'm worrying too much here :-)

PaulStoffregen:
Well, it'll still be better than having to fork & patch the IDE, which I what I've done for every non-beta IDE release over the last 9 years.

PaulStoffregen:
This didn't get merged for the 1.8.8 release today, did it?

cmaglie:
Nope, I'm really sorry Paul, I know that I told you otherwise, but there are some corner case that I still want to address, also the "discovery" alone is not enough, there are also the "upload" and the "port monitor" that should be made pluggable in some way. On the other hand, there were some nasty bugs fixed by 1.8.8 that were long overdue, so in the end we opted to release a good stable release of the IDE.

Anyway: we are committed to keep this PR updated and release the pluggable-discovery.

Some heads up on the work currently going on for this feature:

- I'm implementing the pluggable-discovery for the arduino-cli, it will be compatible with the implementation made here so we can make the feature available on other products like Create and the Create Agent. Also, at some point, I'd like to integrate the arduino-cli in the Arduino IDE making the latter just a frontend. My hope is to do this swap soon, but I know it will not be quick.

- the current serial and network discovery (integrated in the IDE) are being rewritten as pluggable discoveries ([here the serial one](https://github.com/arduino/serial-discovery)). This showed some of the limits of the current (IDE-integrated) discoveries and how they interact with the other parts of the IDE (for example the serial discovery is "paused" during upload, or the serial monitor is deactivated, etc) we should think a way to generalize those behaviors.

Paul, some time ago you offered to help on the pluggable-monitor, are you still available for this?

PaulStoffregen:
The upload already is plenty "pluggable" using the recipes in platform.txt.

Agreed, the serial monitor also need to be "pluggable".  But is that really necessary to expand the scope of this work?  Why not merge this, and then make the serial monitor stuff a separate project?

About 1 year ago, while visiting Portland, Oregon USA, @mbanzi specifically asked me to support installing Teensy through the boards manager.  I've tried to do that.  I'm preparing another custom installer right now for 1.8.8, not using the boards manager.  I do want to make a normal boards manager package, but this pluggable discovery and pluggable serial monitor is needed before I can use the boards manager.

PaulStoffregen:
On this...

> Paul, some time ago you offered to help on the pluggable-monitor, are you still available for this?

Yes, but at this moment I'm bringing up new hardware.  In a few months, I'll be able to contribute much more.

Right now, I might be able to help in small ways.  I do have quite a lot of native serial monitor code already written, with good USB connect / disconnect detection.  If I clean some of that up (remove the many Teensy-specific parts) and make a sample like the Linux udev discovery, maybe it could help for testing?

PaulStoffregen:
Maybe we should bring this over to the CLI?

From the commit history here on github, it's pretty clear you guys are putting nearly all your dev cycles into the CLI and not the IDE.....

PaulStoffregen:
Trying to work with Pluggable Discovery in just-released 1.8.9.  Quick question about how I'm supposed to write the pattern in platform.txt.

What I want to do is something like this:

    discovery.teensy.pattern="{runtime.hardware.path}/tools/teensy_ports" -J2

The problem is {runtime.hardware.path} doesn't get replaced with the pathname to the files I've installed.  I believe there are 2 issues.  First, discoverers are started very early, before the runtime.hardware.path pref exists.  Second, even if it did exist, this refers to the path for the currently selected board, which wouldn't let my discoverer start unless the user had selects the board and then quits & restarts the IDE.

I was able to make it work with this:

    discovery.teensy.pattern="{runtime.ide.path}/hardware/tools/teensy_ports" -J2

But this pattern will only work if my platform is installed inside the IDE, at that specific location.

I did write a patch for DiscoveryManager.java to create a special PreferencesMap with the platforms paths and then do a 2nd StringReplacer.replaceFromMapping() on the pattern.  That works great... but patching DiscoveryManager.java can't be the proper answer!

Hoping you can tell me how you intended for the pattern to be written for the path a platform-provided discoverer?

cmaglie:
The idea is to deliver the discovery as a board-manager tool and use `{runtime.tools.TOOLNAME.path}` or `{runtime.tools.TOOL_NAME-TOOL_VERSION.path}` (as explained [here](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.6.x-package_index.json-format-specification#tools-definitions) and [here](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#tools))

BTW we could allow also `{runtime.hardware.path}`, the problem of the selected board may be solved by setting the variable to the platform path that is currently scanned.

PaulStoffregen:
Ah, ok, now I understand.  The problem comes because I am just copying my platform stuff into the hardware folder, not using a JSON index file.

omzlo:
This is awesome. I've been following this new feature with great interest. I've built an IoT platform called NoCAN that identifies nodes with a special unique identifier (instead of a serial port or a ip address). After playing a bit with 1.8.9, I'm able to list all the nodes in the network in the Arduino IDE, with the Pluggable Discovery you have implemented. 

I have one question though: once the user selects a port, how to I fetch the port information in my `tools.....upload.pattern` definition. For example, how do I pass the "address" defined for the selected port to my uploader command?

PaulStoffregen:
Use {serial.port}.  So it should look something like this:

    tools.yourtoolname.upload.pattern="{runtime.hardware.path}/programname" "{serial.port}"

PaulStoffregen:
Unfortunately, there is still no pluggable serial monitor.  I have some code for Teensy which hard-codes this, using stdin/stdout pipes for the data, and stderr for status.  Thinking about converting it over to use these recipes, and probably JSON instead of my ad-hoc text messages.  @cmaglie - maybe that can serve as a starting point for a pluggable serial monitor in 1.8.10?

omzlo:
> Use {serial.port}. So it should look something like this:
> 
> ```
> tools.yourtoolname.upload.pattern="{runtime.hardware.path}/programname" "{build.path}/{build.project_name}" "{serial.port}"
> ```
> It's still called "serial.port" in the prefs, even though we're now able to do any arbitrary protocol and any "address".

@PaulStoffregen This is perfect thanks! I just tried it and it works in my proof of concept.

omzlo:
Hi. My POC works well with the START_SYNC mechanism, however it failed with the fallback LIST mechanism.
After looking at the code, I believe that line 146 in PluggableDiscovery.java  
```
        BoardPort port = mapJsonNodeToBoardPort(mapper, node);
```
shoud be changed to
```
        BoardPort port = mapJsonNodeToBoardPort(mapper, portNode);
```
This made things work in my setup.
I hope this helps.

facchinm:
@cmaglie what do you think about this :top: ?

omzlo:
I've been trying to release a package that takes advantage of the new discovery features of the Arduino IDE.
My tool is called "nocanc". I run into the following issue. 

The following fails in platform.txt:

```
discovery.nocanc.path={runtime.tools.nocanc.path}
discovery.nocanc.cmd=nocanc
discovery.nocanc.cmd.windows=nocanc.exe
discovery.nocanc.pattern="{path}/{cmd}" arduino-discovery
```

However, the following works (at least on Mac OSX):

```
discovery.nocanc.cmd=nocanc
discovery.nocanc.cmd.windows=nocanc.exe
discovery.nocanc.pattern="{runtime.tools.nocanc.path}/{cmd}" arduino-discovery
```

The reason seems to be that variable substitution using PreferencesData only happens on the pattern and not on the other `discovery.*` fields, as the code in DiscoveryManager suggests:

```java
  String pattern = discoveryPrefs.get("pattern");
  ...
  pattern = StringReplacer.replaceFromMapping(pattern, PreferencesData.getMap());`
```

I believe that this will cause further issues on Windows where the `cmd.windows` substitution won't happen as well. 
This could explain some problems previously reported by @PaulStoffregen above.

facchinm:
@omzlo the differentiation between windows and unix, as far as the filename is the same (without the extension), is not needed anymore (at least from IDE 1.6.something). 
However, I understand your point; the discoverer should try to resolve exactly as the normal platform.txt parser

PR #8128, opened 2018-10-23
Ability to override ssh upload port from mDNS info

BertvanHoekelen:
Adds the ability to specify the `ssh_upload_port` in the mDNS network discovery.

Most networks where I install a yún, don't allow traffic over port 22. So I have changed the default ssh port on the yún. But changing my `.ssh/config` file all the time when I add the yún into a different network costs me a lot of time. This change adds an extra TXT property `ssh_upload_port` so that port can be configured on the yún and that the IDE knows to use the different port. If the `ssh_upload_port` is not specified it will fall back to port `22`.


BertvanHoekelen:
Closing because lack of response.

Issue #7498, opened 2018-04-21
avrdude goes 100% cpu after unsuccessful uploading and closing arduino IDE


adamitsch:
When I close Arduino IDE after unsuccessful uploading (when it is trying to connect via serial port), avrdude process consumes 100% cpu.

facchinm:
Hi @adamitsch ,
which OS / IDE version are you running? Starting from 1.8.0 (if I remember correctly) all "zombies" upload tools get killed after two minutes of inactivity. The 100% cpu behaviour is only due to avrdude getting stuck in an infinite loop. 
This should be probably tracked under https://github.com/arduino/Arduino/issues/5928

hoonto:
I saw this problem this morning. Last night I had an unsuccessful upload to board due to serial port permissions. I fixed the serial port permissions with 

```
sudo usermod -a -G dialout mmullens
sudo chmod a+rw /dev/ttyUSB0
sudo chmod a+rw /dev/ttyACM0
```

My laptop runs a lot of stuff, so didn't notice the activity until this morning at which point I saw 100% cpu utilization:

```top - 10:01:11 up 3 days, 21:14,  1 user,  load average: 2.44, 2.41, 2.27
Tasks: 372 total,   4 running, 302 sleeping,   0 stopped,   1 zombie
%Cpu(s):  9.3 us, 18.4 sy,  0.0 ni, 72.3 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem : 65920704 total, 47128480 free,  9402004 used,  9390220 buff/cache
KiB Swap:        0 total,        0 free,        0 used. 55327280 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                                                                          
30554 mmullens  20   0   16972   5664   2220 R 100.0  0.0 779:09.58 avrdude                                                                                                                                          
30856 mmullens  20   0   16972   5696   2244 R 100.0  0.0 777:14.57 avrdude                                                                                                                                          ```

```
$ cat /etc/lsb-release
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=19
DISTRIB_CODENAME=tara
DISTRIB_DESCRIPTION="Linux Mint 19 Tara"
```

Arduino IDE 1.8.6

Hope that helps!

gibeautc:
seeing the same issue here.  Arduino IDE is not running anymore (hasnt been in a day or two) 

[...]

PR #8184, opened 2018-11-08
Kill active programmer if still alive after closing last IDE window

facchinm:
Fixes https://github.com/arduino/Arduino/issues/7498

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -26,6 +26,7 @@ import cc.arduino.Compiler;
 import cc.arduino.Constants;
 import cc.arduino.UpdatableBoardsLibsFakeURLsHandler;
 import cc.arduino.UploaderUtils;
+import cc.arduino.packages.Uploader;
 import cc.arduino.contributions.*;
 import cc.arduino.contributions.libraries.*;
 import cc.arduino.contributions.libraries.ui.LibraryManagerUI;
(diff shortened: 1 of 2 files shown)


app/src/processing/app/Base.java
```diff
@@ -925,45 +926,20 @@ public void actionPerformed(ActionEvent actionEvent) {
...
-        //ignore
-      }
-      rebuildRecentSketchesMenuItems();
 
-      // Save out the current prefs state
-      PreferencesData.save();
-
-      // Since this wasn't an actual Quit event, call System.exit()
-      System.exit(0);
+      handleQuit();
```
matthijskooijman: Isn't this something for a separate commit? Calling `handleQuit()` here seems unrelated to killing the programmer?

app/src/processing/app/Base.java
```diff
@@ -925,45 +926,20 @@ public void actionPerformed(ActionEvent actionEvent) {
...
-//        if (editor == editors[i]) {
-//          for (int j = i; j < editorCount-1; j++) {
-//            editors[j] = editors[j+1];
-//          }
-//          editorCount--;
-//          // Set to null so that garbage collection occurs
-//          editors[editorCount] = null;
-//        }
-//      }
+
```
matthijskooijman: Same here, this might be nicer in a separate commit.

facchinm:
Going for the separate commits :wink:

app/src/processing/app/Base.java
facchinm: Moved to own commit

app/src/processing/app/Base.java
facchinm: Moved to own commit

Issue #3495, opened 2015-07-10
Serial Port selected disconnects after upload.


ReeceM:
When I upload to my Sparkfun pro micro after selecting the port its on, after upload the selected port becomes deselected in the drop down and I can't upload again without going there and re-selecting the com port.

I am using a Hourly build when this happens : 
`Arduino 1.6.6 Hourly Build 2015/07/07 9:41`

But if I use ver. 1.6.0 the serial port stays connected after upload. 

Is this a localised problem or am I suffering from beta.

matthijskooijman:
The micro disconnects and reconnects USB when switching from sketch to bootloader and back, perhaps this causes the problem? I haven't tried reproducing this, though.

ReeceM:
I think the switching is causing the problem, because it switches to the port for uploading but never returns to the normal port. The IDE never reselects the port after an upload.

I think something that works with the Leonardo/micro upload style has been changed in the code, because it used to work correctly, i.e. it remembered the com port it is on after an upload.

facchinm:
Hi @ReeceM , are you on Linux? Your problem can be triggered by the serial monitor (which is now kept in a "zombie" state during upload) respawning too soon and thus locking the previous port, so the sketch will get another one. 
Can you post here your PC specs?

ReeceM:
Hi @facchinm, I am on Windows XP, never had this problem before.

facchinm:
Very strange, as Windows assigns different COM numbers to bootloader and sketch. Are you uploading with the serial monitor opened?

ReeceM:
The serial monitor is not opened. The problem is that the IDE is not re-selecting the port after upload, even though it is there, example:
***BEFORE***
![2015-07-13 12_04_19-blink _ arduino 1 6 6 hourly build 2015_07_10 12_34](https://cloud.githubusercontent.com/assets/2767904/8647335/597bd8fe-2957-11e5-8ada-e18c4ece8809.png)
***AFTER***
![2015-07-13 12_05_51-blink _ arduino 1 6 6 hourly build 2015_07_10 12_34](https://cloud.githubusercontent.com/assets/2767904/8647339/5dba1b2e-2957-11e5-809b-f02157c2352d.png)

facchinm:
Ok, thanks for the screenshot, we'll analyze this issue as soon as @ffissore is back!

ReeceM:
@facchinm  cool :smiley:

pablosun:
[...]

PR #8218, opened 2018-11-21
Refactor port selection after 1200bps touch upload

facchinm:
New behaviour:
if upload failed or we are uploading through a "Programming" port (that does not disappear), leave the user selected port selected.
if upload succeded and we are using 1200bps touch, wait for the first port that reappears, and if nothing reappears after the timeout select the bootloader port.

Fixes https://github.com/arduino/Arduino/issues/3495

### DIFF
--- a/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
+++ b/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
@@ -199,9 +199,6 @@ public class SerialUploader extends Uploader {
       BaseNoGui.getDiscoveryManager().getSerialDiscoverer().pausePolling(false);
     }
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I haven't looked at the code in detail, but can't we match the port based on the USB identifier (I mean the connection path / USB port, not the vidpid).

facchinm:
Matching on the usb identifier would be THE BEST; I tried that approach once with libusb hotplug but it didn't really work :slightly_smiling_face: Anyway, any help on this would be greatly appreciated.

PilotinControl25:
Ok so this build works under windows Vista by pressing the reset button and it uploads....any way of not doing that each time?

facchinm:
@PilotinControl25 which board are you targeting?

cmaglie:
The fix looks good, it should cover new cases without touching the existing one. I'm merging this one so it may get broader testing on the hourly build.

PR #8221, opened 2018-11-21
Adding support of "additional library directories" to IDE & compilation

garnould:
This commit adds support of "additional library directories" to IDE
(supporting File/Examples menu, Sketch/Include Library menu and
keywords highlighting), and compilation process (by sending additional
-libraries options to arduino-builder).

Those additional library directories (one or more) are setup by adding an
option to preferences.txt :  
additional_library_directories=/path/to/FreeIMU-Updates;[/mnt/ardether[;etc.]]

Use case : you need to use libs from a (different) git repository staying outside your arduino directory/sketchbook.  

Those additional libraries are included as Location.ADDITIONAL.

Modified priorities become Location.SKETCHBOOK > Location.ADDITIONAL > Location.CORE > Location.REFERENCED_CORE > Location.IDE_BUILTIN

### DIFF
--- a/arduino-core/src/cc/arduino/Compiler.java
+++ b/arduino-core/src/cc/arduino/Compiler.java
@@ -49,8 +49,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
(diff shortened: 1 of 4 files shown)


facchinm:
Hi @garnould ,
thanks for the submission! This looks very interesting indeed! 
I'd be fine with merging and documenting it without exposing in the preferences GUI, so you'd use it only if you really know what you are doing :slightly_smiling_face: 
@cmaglie any thought on this?

garnould:
Hi @facchinm ! Newbie question : you've added this PR to "1.8.8 milestone". Where could I read the meaning of this (in term of review, merge, etc.) ?

facchinm:
It simply means that it will be analyzed before the 1.8.8 release to decide if it's ok to merge or  if we need to wait a little bit :slightly_smiling_face:

cmaglie:
If we want to implement this we should port in the `arduino-cli` first.

PR #8349, opened 2019-01-02
Remove last editor window after saving its properties

### LINKED ISSUE #8337: IDE 1.8.8 does not remember last sketch nor window size.
I installed IDE 1.8.8 in linux mint (mate versions 18.1 and 19.1) to find that the IDE does not remember the last edited sketch when starting up.
Also it starts with a small window size no matter what the scaling settings are.
On my other pc's I am still using 1.8.5 and they are working ok. 
So I did some digging and found out that version 1.8.7 is still working fine here.
I also found that the issue remembering the last sketch is most likely caused by a reset to zero
of the last.sketch.count (in preferences.txt) at some point by the program.
I have no idea as to what could be causing the window sizing issue though.

facchinm:
Only OSX needs the "app" to stay open after handleQuit and to remove the editor windows when called.

Fixes #8337

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -932,9 +932,10 @@ public class Base {
     }
 
     if (editors.size() == 1) {
+      handleQuit();
+      // Everything called after handleQuit will only affect OSX
       editor.setVisible(false);
       editors.remove(editor);
-      handleQuit();
     } else {
       // More than one editor window open,
       // proceed with closing the current window.


vimes666:
Thanks for the swift response! 
I tested http://downloads.arduino.cc/javaide/pull_requests/arduino-PR-8349-BUILD-817-linux64.tar.xz just now and the problem seems to be solved :)
As for windows, I dont know, my latest working version is 98. Someone else will have to test that.

PR #8420, opened 2019-01-21
Use arduino-cli to manage libraries (WIP)

cmaglie:
This PR adds `arduino-cli` to the build of the Arduino IDE and use it to manage libraries (install/uninstall/list/search etc.).
This is still a work-in-progess,. Things to check:

* see if proxy setting are propagated correctly
* see if sketchbook directory is taken correctly (non-default sketchbook, portable mode, etc...)
* test cli of Arduino **IDE**
* see if Java tests are still passing (probably not)


cmaglie:
replaced by #8420

PR #8621, opened 2019-03-08
Update Eclipse project files

Pieter12345:
This PR adds missing build path entries to the Java Eclipse IDE specific `.classpath` files, effectively resolving all compile errors in the project. An outdated license file for jmdns was also updated in the process.
The changes in this PR only affect people who use Java Eclipse to build this project.


arduino-core/.classpath
```diff
@@ -11,7 +11,7 @@
 	<classpathentry kind="lib" path="lib/jssc-2.8.0-arduino2.jar"/>
 	<classpathentry kind="lib" path="lib/jsch-0.1.50.jar"/>
 	<classpathentry kind="lib" path="lib/commons-exec-1.1.jar"/>
-	<classpathentry kind="lib" path="../app/lib/commons-httpclient-3.1.jar"/>
+	<classpathentry kind="lib" path="/Arduino/app/lib/commons-httpclient-3.1.jar"/>
```
facchinm: I think that this line should be removed from the commit; not everyone clones this repo in a folder called `Arduino` , so the relative path is better suited for this.

Pieter12345: That is a fair point. Having `../app/lib/commons-httpclient-3.1.jar` does however result in:
`Project 'arduino-core' is missing required library: 'C:\Users\X\Documents\Java Eclipse Workspace\app\lib\commons-httpclient-3.1.jar'`
Which starts one directory below the desired directory.
Using `./app/lib/commons-httpclient-3.1.jar` or `app/lib/commons-httpclient-3.1.jar` results in: `Project 'arduino-core' is missing required library: 'app/lib/commons-httpclient-3.1.jar'`.

Since this might have to do with my setup, I will explain this setup here. I've cloned the repo into the `Arduino` directory in my workspace. Then, I clicked `new project` and named it `Arduino`, causing Eclipse to import the project. This does not import `arduino_core` yet, so I repeated this process linking to `Arduino/arduino_core` as workspace and `arduino_core` as project. This in combination with the added/updated dependencies in this PR results in two projects that can be compiled in Eclipse. What I am suspecting is that there might be a way to add `arduino-core` as a subproject with `Arduino` as parent instead, allowing relative paths to dependencies to go one directory higher (this is about an `arduino-core` dependency which is actually in the `Arduino` project that depends on `arduino-core`). Do you know whether this is true or not and if so, how to perform this setup in Eclipse?

Pieter12345: Apparentally the commons-httpclient-3.1 dependency was never used, so I removed it. Building with "ant dist" still succeeds.

Pieter12345:
@facchinm The requested change was made and the PR is up-to-date with master again. Is any additional action required to get this merged?

Pieter12345:
Force pushed to c00e50a to also remove references to that removed library from the class path in the launcher config.xml and config_debug.xml files.

Pieter12345:
These additional changes make the project more clean, but also allow running the Arduino IDE from Java Eclipse on Windows. To do so, create a run configuration for main class `processing.app.Base` with the following VM arguments:
```
-splash:"${workspace_loc:Arduino}/build/shared/lib/splash.png"
-Dsun.java2d.d3d=false
-Djna.nosys=true
-Djna.nounpack=true
-Djna.boot.library.name=jnidispatch-4.2.2-win32-x86
-Djna.boot.library.path="${workspace_loc:Arduino}/build/windows/work/lib"
-Djssc.library.path="${workspace_loc:Arduino}/build/windows/work/lib"
-DAPP_DIR="${workspace_loc:Arduino}/build/windows/work"
```

Pieter12345:
Force pushed to revert linking to my personal 32 bits JVM by name in the `.classpath` file.

.classpath
```diff
@@ -2,31 +2,45 @@
 <classpath>
 	<classpathentry excluding="processing/app/tools/format/|processing/app/tools/format/src/|processing/app/Trace.java|processing/app/RunnerClassLoader.java" kind="src" path="app/src"/>
 	<classpathentry kind="src" path="app/test"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry kind="lib" path="app/lib/apple.jar"/>
-	<classpathentry kind="lib" path="app/lib/ecj.jar"/>
-	<classpathentry kind="lib" path="app/test-lib/junit-4.11.jar"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
```
facchinm: This line bothers me a bit... Is it necessary to specify JavaSE-1.8 ?

Pieter12345: Ideally, I would like to specify Java 1.8 32 bits, but this does not seem to be possible. Not specifying a java version leads to using the user-defined workspace default, which can differ from what is used to build against by Ant and the Continuous Integration software. Since the Ant config contains `jvmRequired="1.8"`, I think it is a good idea to stick to that version. Note that later JDK versions can suffice for earlier version settings in Eclipse (e.g. JDK 1.8 can be used to build a project that's configured to build against 1.6).
I'd like to hear your thoughts / proposal on this.

Pieter12345:
Rebased this PR onto master.

Pieter12345:
Rebased this PR onto master to resolve merge conflicts.

Pieter12345:
In the current PR state, the following setup is possible in Eclipse by importing `Arduino` (my clone name), `Arduino/app` and `Arduino/arduino-core` into Eclipse:
afbeelding

Note that since parent project `Arduino` has the same source and binary directory as `Arduino/app`, it is not necessary to import `Arduino/app` as separate project. It is possible to remove the `Arduino` project setup so that it becomes just a container for the 2 child projects (or modules), I'd like to have your opinions on this @facchinm and @cmaglie.

facchinm:
Hi @Pieter12345 ,
I just tried the complete patchset; there's a problem with 6ed443e (in fact it breaks `ant build` since the ant projects are no more self contained). A trivial patch is to reference the other folder in the lib search path, like this
```patch
diff --git a/arduino-core/build.xml b/arduino-core/build.xml
index cef9fe404..4ddbaea94 100644
--- a/arduino-core/build.xml
+++ b/arduino-core/build.xml
@@ -5,6 +5,9 @@
     <fileset dir="lib">
       <include name="*.jar"/>
     </fileset>
+    <fileset dir="../app/lib">
+      <include name="*.jar"/>
+    </fileset>
     <pathelement path="${env.JAVA_HOME}/lib/tools.jar"/>
   </path>
diff --git a/app/build.xml b/app/build.xml
index cc38670ad..08b57ba31 100644
--- a/app/build.xml
+++ b/app/build.xml
@@ -2,6 +2,9 @@
 <project name="Arduino PDE" default="build">
 
   <path id="class.path">
+    <fileset dir="../arduino-core/lib">
+      <include name="*.jar"/>
+    </fileset>
     <fileset dir="lib">
       <include name="*.jar"/>
     </fileset>
```
but I'm not a big fan of this approach. Probably unifying the `lib` folder and have it at the top level would work (not very Java-ish). @cmaglie any hint?

Anyway, I also haven't been able to setup the eclipse project to run; it looks like APP_DIR variable is not being applied, so all files are searched under `app/lib/`.
Did you find something similar?

cmaglie:
> It is possible to remove the Arduino project setup so that it becomes just a container for the 2 child projects (or modules)

I've successfully setup eclipse without the "parent" Arduino project, so I'm positive to removing it.

> but I'm not a big fan of this approach. Probably unifying the lib folder and have it at the top level would work (not very Java-ish). @cmaglie any hint?

+1 to this too, but we should check launchers configuration for the "dist" target.

Pieter12345:
> 
> 
> I just tried the complete patchset; there's a problem with [6ed443e](https://github.com/arduino/Arduino/commit/6ed443e4d7756a51f55cc7176bb1f4e5d61016a8) (in fact it breaks `ant build` since the ant projects are no more self contained). A trivial patch is to reference the other folder in the lib search path, but I'm not a big fan of this approach. Probably unifying the `lib` folder and have it at the top level would work (not very Java-ish). @cmaglie any hint?


I've tried putting a `lib` directory in the base directory, but there appears to be no way to link there other than making this `lib` a separate project as well (the reason it is possible to link to `/arduino-core/lib/...` from `app` is that `arduino-core` is a known project in the workspace). Ideally, I would like to see Apache Ant download and link dependencies like most Maven or Gradle projects do (Maven or Gradle actually replace the need for also having dependencies in the `.classpath` file, making it impossible to have mismatching dependency versions in config files). I have never worked with Ant before, so support for this in Ant is a gray area for me though.
As for fixing this issue for this PR without fixing the above mentioned technical debt, the two easiest solutions I can come up with are the one you mentioned and introducing duplicate jars again (have the same jar in both projects if both projects use it). I would be in favor of the first option, since the second allows more space for errors and non-deterministic behavior (unknown which version of a dependency is picked when two versions are available by accident) in my opinion. Any thoughts on this in combination with better dependency management @facchinm / @cmaglie ?

> 
> Anyway, I also haven't been able to setup the eclipse project to run; it looks like APP_DIR variable is not being applied, so all files are searched under `app/lib/`.
> Did you find something similar?

Perhaps try to set your working directory (under `run configurations`) to the `work` directory?

Pieter12345:
Rebased onto master.

facchinm:
> Perhaps try to set your working directory (under run configurations) to the work directory?

This may be in fact what In was missing :slightly_smiling_face: 

About the maven/gradle dependency manager I'm totally positive about adding it. The problem right now is really technical debt since we don't have a proper Java programmer. Any pull request is well accepted as usual, but this time it could be a bit cumbersome to provide test builds since the build environment would change too.

Pieter12345:
> About the maven/gradle dependency manager I'm totally positive about adding it. The problem right now is really technical debt since we don't have a proper Java programmer. Any pull request is well accepted as usual, but this time it could be a bit cumbersome to provide test builds since the build environment would change too.

Maven would fully replace Apache Ant, which means that the entire Build.xml files would have to be converted to pom.xml Maven files in an entirely different format. Finding a translation for all functionality and then translating it will likely take quite some time, but would be good for the project (less software to install to start developing (only Maven, which comes with most IDE's) and no more IDE-dependent (.classpath / .project files for Eclipse) files).
This indeed does affect the build environment. I could not find much info on the build environment, but it looks like it is some task that builds the software every hour (whether it changed or not) and does never run tests. What build software is actually used and how is it configured (I did not see a CI setup file in the repository)?

facchinm:
The CI infrastructure is based on Jenkins and builds the IDE for every merged commit to compose the hourly build (crosscompiling with `ant dist` for every OS target). In parallel, some lower priority jobs build the IDE and run the tests on their native OS. 

If we replace `ant` we need a tool that:
- compiles jars :slightly_smiling_face: 
- compresses/uncompresses archives
- controls checksums
- can invoke other jars (like appbundler)
- invokes external binaries
- runs tests

They all look simple (maybe appbundler is not so easy but there are a couple of projects on github that look ok).
@cmaglie what do you think about that?

szb512:
revapi/testjars can compile jar files

Pieter12345:
>
> Need to fix ant build before going on

I've added a copy of the jars that are used in both `app` and `arduino-core` to `app/lib`, which should fix the Ant build. This might not be the ideal solution, but it should be sufficient for the scope of this PR. Changing the build system to download jars instead of including them as binaries would be the proper fix, but this falls out of scope for this PR.

As of this point, this PR:
- Fixes Eclipse project setup files (`.classpath` files and removal of the redundant parent container setup files).
- Cleans up dependencies (remove unused, update license file).
- Allows developers to run the Arduino IDE directly from Eclipse.

Can you validate that this change fixes the Ant build @facchinm? If so, is there anything left that prevents this PR from being merged?

szb512:
> 
> 
> > Need to fix ant build before going on
> 
> I've added a copy of the jars that are used in both `app` and `arduino-core` to `app/lib`, which should fix the Ant build. This might not be the ideal solution, but it should be sufficient for the scope of this PR. Changing the build system to download jars instead of including them as binaries would be the proper fix, but this falls out of scope for this PR.
> 
> As of this point, this PR:
> 
>     * Fixes Eclipse project setup files (`.classpath` files and removal of the redundant parent container setup files).
> 
>     * Cleans up dependencies (remove unused, update license file).
> 
>     * Allows developers to run the Arduino IDE directly from Eclipse.
> 
> 
> Can you validate that this change fixes the Ant build @facchinm? If so, is there anything left that prevents this PR from being merged?

@Pieter12345  I agree with you, but we can incorporate the proper fix with a few changes, but still, it would change the scope of this PR.

Issue #8238, opened 2018-11-27
Can't access buttons on Preferences window on screens with vertical resolution of 600 px (again)


rin67630:
On IDE 1.8.7 ARM  (Raspberry Pi)
The [Cancel] and [OK] buttons of the preference settings are still out of reach on screens with 600px heigth.

Please consider rearranging the 8 tick options on two rows of 4.
That should reduce the overall heigth enough to fit.

Thank you.

PaulStoffregen:
As I suggested on #6459, you really should post a screenshot of the problem, so developers can see the problem (without the trouble of using a Raspberry Pi *and* a special/tiny monitor).  Notice just below the text entry it says "Attach files by dragging & dropping, selecting them, or pasting from the clipboard."  That's how you put the screenshot into your message here.

rin67630:
Here the current screen on a 1024x600 screen:
![github 8238 1](https://user-images.githubusercontent.com/14197155/49097480-1a377800-f26d-11e8-9fe3-404f5c6a8d9d.jpg)
My proposal:
![github 8238 2](https://user-images.githubusercontent.com/14197155/49097501-26bbd080-f26d-11e8-8106-931a765cfebc.jpg)
Alternatively, bringing a drawer to the window would be more universal.

rin67630:
By the way: you don't need a Raspberry Pi and a special/tiny monitor to test: just set your screen resolution to 800x600 on your current PC and enjoy the trouble as well. 
Regards.
Laszlo

tigoe:
I don't have strong feelings about the column layout; I'm fine with the change if it helps others. Worth mentioning that the prefs panel is one of the more accessible parts of the IDE on OSX (turn on VoiceOver using command-F5 to navigate it using the screen reader), so I wouldn't want to mess that up.

I think if you're going to change the prefs panel, it's worth considering what else you might add there, perhaps from the preferences.txt or boards.txt, etc. What are the most commonly called for additions from preferences.txt?

cjdg:
Its happening again on 1.8.9

cjdg:
Maybe GUI should be splitted into tabs or adding a scrolling

facchinm:
@cjdg can you provide a screenshot? In 1.8.9 the Preference window is way  shorter than 600px

PR #8638, opened 2019-03-11
Preferences: arrange checkboxes into two columns

facchinm:
Should fix https://github.com/arduino/Arduino/issues/8238

I'm not a huge fan of this layout (it could be problematic for some translations) but it's one of the solutions.

@tigoe @mbanzi @cmaglie @sandeepmistry any thought?

Screenshot attached for  the lazies :wink: 
2019-03-11-173257_850x504_scrot


PaulStoffregen:
Is "Update sketch files to new extension on save" still needed in the UI, more than 7 years after adding the .ino in version 1.0?

Is "Aggressively cache compiled core" really needed as a user visible setting?

facchinm:
@PaulStoffregen I agree with both your concerns; when core caching was introduced I feared it could have side effects so I pushed to introduce the flag (I think it can be safely removed now since no bug related to that functionality was ever reported).
Same applies to pde -> ino , I'd keep it s default and hide in the UI.

tigoe:
I'm good with both of Paul's suggestions.

PR #8647, opened 2019-03-12
Properly scale all UI elements based on Theme.scale() setting

facchinm:
The code in helper class is UGLY but it looks functional at least.
Need some testing on various 4K display with different scaling setting (both native and via Preferences panel)


shepting:
These UI elements on the Mac (4k iMac) display are still blurry:


Mac UI Scale

lmihalkovic:
In my local branch i made them vector graphics that get rasterized and cached

PR #8660, opened 2019-03-15
Add serial monitor send/receive encoding options

### LINKED ISSUE #4452: Change serial monitor encoding
Hi!

It would be great if in the serial monitor had an option to change the encoding used.

In my case, my sketches are using UTF-8, so print messages use that encoding. By default, serial monitor is using ISO-8859 (probably the default one of Windows 7), so those print messages are not shown properly:

serial-monitor-encoding-problems

I've found other issues talking about this problem. Instead of detecting the encoding used, it might be easier if the user could select the charset he wants to use. The option (a selectable list with the most used character sets) may be put in the bottom right corner, left of the two existing selects (baud rate and line feed).

Thanks!!

Pieter12345:
- Add "send as <encoding>" dropdown menu.
- Add "receive as <encoding>" dropdown menu.

Sending and receiving can be done in any encoding specified in StandardCharsets with the additional option to send comma-separated bytes and receive newline-separated bytes directly.

This is related to arduino/arduino-ide#1728 and offers an easy implementation for issue arduino/arduino-ide#1727.

### DIFF
--- a/app/src/processing/app/AbstractTextMonitor.java
+++ b/app/src/processing/app/AbstractTextMonitor.java
@@ -40,6 +40,8 @@ public abstract class AbstractTextMonitor extends AbstractMonitor {
   protected JButton clearButton;
   protected JCheckBox autoscrollBox;
   protected JCheckBox addTimeStampBox;
+  protected JComboBox<String> sendEncoding;
+  protected JComboBox<String> receiveEncoding;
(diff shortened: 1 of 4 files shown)


Pieter12345:
Rebased this PR onto master.

Madygethub:
My arduino doesn't work, why?

PR #8674, opened 2019-03-18
Add serial monitor command history

### LINKED ISSUE #4891: [REQ] Ability to recall previously sent strings in Serial Monitor
Something other serial apps have that the IDE doesn't is the ability to recall previously sent strings in the  serial monitor. Data could be retained per Arduino session (not session of the serial monitor, so it would last code changes and uploads) and be invoked with arrow key up/down. This is especially useful if a user needs to repeatedly send the same string, yet without copying and pasting. 

Coolterm has good implementation of this and it would be great to see this implemented in Arduino IDE.

Pieter12345:
The behavior is as follows:
- Pressing the UP key will select older commands.
- Pressing the DOWN key will select newer commands, restoring the last unexecuted command if available.
- Pressing the ESC key will reset the input field to the latest unexecuted command and reset the traversal location. Pressing ESC while the latest unexecuted command is selected will clear the input field.

This fixes #4891, with the difference that the proposed solution has a command history per Arduino session and this implementation has a command history per serial monitor start.

### DIFF
--- a/app/src/processing/app/CommandHistory.java
+++ b/app/src/processing/app/CommandHistory.java
@@ -1,7 +1,7 @@
 package processing.app;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedList;
+import java.util.ListIterator;
 
 /**
  * Keeps track of command history in console-like applications.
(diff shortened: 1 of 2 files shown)


facchinm:
@ArduinoBot build this please

Pieter12345:
Rebased this PR onto master.

facchinm:
Manually rebased and merged, thanks!

PR #8686, opened 2019-03-20
Use user-defined tab settings in new sketch generation

Pieter12345:
When creating a new sketch, it is initialized with the BareMinimum example sketch. This example sketch uses 2-width whitespace indentation, which might differ from the user-defined tab settings. This PR makes the generated example sketch consistent with the user-defined tab settings.


Pieter12345:
@facchinm @cmaglie When making this change, I were left with two questions:
1. The `arduino-core/src/processing/app/helpers/FileUtils.java` class has a high overlap with the [org.apache.commons.io.FileUtils](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html) class. Is there a reason for not using this library?
2. The `arduino-core/src/processing/app/helpers/StringUtils.java` class has the same name as `org.apache.commons.lang3.StringUtils`. In this PR, I wrapped an Apache StringUtils method so that I didn't have to specify the full path in the `Base` path (to disambiguate from the other StringUtils). What is the desired way to handle this issue in this project?

facchinm:
Apache commons was introduced as a dependency very late, so that's the reason for code duplication. Using commons everywhere would be a good solution IMO. In parallel it would be awesome to make the whole codebase consistent again (I'm sure there are at least 3 methods to write a string to a file in use right now :slightly_smiling_face: )

Anyway, I really like this PR! We are currently having a small problem on the download server so I can't launch the ArduinoBot build on it but I'll do as soon as we fix it.

Pieter12345:
Rebased this PR onto master.

app/src/processing/app/Base.java
```diff
@@ -764,7 +764,20 @@ protected File createNewUntitled() throws IOException {
     if (!newbieFile.createNewFile()) {
       throw new IOException();
     }
-    FileUtils.copyFile(new File(getContentFile("examples"), "01.Basics" + File.separator + "BareMinimum" + File.separator + "BareMinimum.ino"), newbieFile);
+
+    // Initialize the pde file with the BareMinimum sketch.
```
pfeerick: We're still talking about `pde` files? That's _so_ last year! :laughing: :rofl:

PR #8688, opened 2019-03-21
Replace StringUtils + Partially cleanup FileUtils + Fix unit test

Pieter12345:
- Replace StringUtils by Apache's commons.lang3.StringUtils.
- Remove unused methods from FileUtils that do not have an obvious use case or have an easy replacement (for example through Files).
- Replace some calls to FileUtils with direct methods (functional code).

Having two StringUtils classes is confusing and not necessary since Apache's StringUtils provides all used functionality.

As for FileUtils; Apache commons.io contains a FileUtils class that can replace the copy / listFiles / delete methods, but some of these methods account for `SOURCE_CONTROL_FOLDERS`, which would make the current FileUtils implementation better if these `SOURCE_CONTROL_FOLDERS` actually exist in more than one use case.

The overall goal of this PR is to make the source code cleaner and more efficient without making functional changes.


arduino-core/src/processing/app/helpers/FileUtils.java
```diff
@@ -209,25 +146,6 @@ public static String readFileToString(File file, String encoding) throws IOExcep
...
   }
 
-  public static List<String> readFileToListOfStrings(File file) throws IOException {
-    List<String> strings = new LinkedList<>();
-    BufferedReader reader = null;
-    try {
-      reader = new BufferedReader(new FileReader(file));
-      String line;
-      while ((line = reader.readLine()) != null) {
-        line = line.replaceAll("\r", "").replaceAll("\n", "").replaceAll(" ", "");
```
cmaglie: Uhm... does `Files.readAllLines` handles all types of newlines (MAC  / Linux / WIN)?
This `replaceAll` chain is quite weird since it remove also the spaces(!), I guess it has been added for some reason, let me check if the git history can give some hints.

Pieter12345: readAllLines(Path, Charset) states to handle them all:
afbeelding
As for whitespaces, this method was only used to merge bootloader .hex (text) files with sketch .hex (text) files. This entire `MergeSketchWithBooloader.java` class does not seem to be used in anything other than a test. Is there a reason for keeping this code or could it be removed?

cmaglie: The original commit is https://github.com/arduino/Arduino/commit/88e8019419, so yes it has been added specifically for MergeSketchWithBootloader. AFAIK it's not used anymore, so it can be removed.

PR #8698, opened 2019-03-25
Add editor zoom "CTRL =" shortcut

Pieter12345:
Add "CTRL =" as additional shortcut to increase the editor font size. This shortcut should be added because the '+' and '=' characters are often on the same key on the keyboard and having to press SHIFT as well is not intuitive for all users (especially since many common applications support "CTRL =").


Pieter12345:
Added in https://github.com/arduino/Arduino/pull/8704.

PR #8704, opened 2019-03-26
Serial/network monitor and editor console dynamic font size hotkeys and improved updating

### LINKED ISSUE #8615: Font-size change with CTRL +/-/mouse wheel in Serial monitor and Console/messages pane
Dynamic font-size adjustment currently functions in the editing area.
CTRL +/- and CTRL Mousewheel work beautifully for this purpose (and are obviously the standard shortcuts for it).

HOWEVER, if someone could implement this in at least two other areas it would be extremely useful to many people:
1. The Console Window (the place where you see compile/upload messages), and
2. The Serial Monitor

I am aware of features being settable, so before anyone informs me of this, I am specifically referring to the ability to do it dynamically, as is allowed in the Editor.  :)

Perhaps needless to say: This is especially useful for the visually-impaired, who have to constantly balance how much material they can see (as much text as they can) with the size needed to see it (which changes with the time of day, position, rubbing their eyes, etc.)

Pieter12345:
- Add CTRL +/- and CTRL scroll shortcuts to increase/decrease the serial/network monitor and editor console text size in the same way the editor text can currently be adjusted.
- Make the serial/network monitor text size update when the editor font size is changed.
- Make the editor console text size update when the editor font size is changed. Before this, the size was already updated, but not applied to existing text in the editor. This lead to the following bug:
afbeelding

These changes combined cause the editor, editor console, serial monitor and network monitor to share their font in such a way that the font size can be updated using CTRL +/-/scroll in these elements and the new font size immediately applies to all these elements on change.

Fixes #8615.
Outdates https://github.com/arduino/Arduino/pull/8698 (this is added to this PR).

### DIFF
--- a/app/src/processing/app/EditorConsole.java
+++ b/app/src/processing/app/EditorConsole.java
@@ -26,6 +26,9 @@ import cc.arduino.ConsoleOutputStream;
 import javax.swing.*;
 import javax.swing.text.*;
 import java.awt.*;
+import java.awt.event.KeyAdapter;
+import java.awt.event.KeyEvent;
+import java.awt.event.MouseWheelEvent;
 import java.io.PrintStream;
 
 import static processing.app.Theme.scale;
@@ -61,7 +64,7 @@ public class EditorConsole extends JScrollPane {
   private SimpleAttributeSet stdOutStyle;
   private SimpleAttributeSet stdErrStyle;
 
-  public EditorConsole() {
+  public EditorConsole(final Base base) {
     document = new DefaultStyledDocument();
 
     consoleTextPane = new JTextPane(document);
@@ -110,6 +113,40 @@ public class EditorConsole extends JScrollPane {
     setMinimumSize(new Dimension(100, (height * lines)));
 
     EditorConsole.init(stdOutStyle, System.out, stdErrStyle, System.err);
+
+    // Add "CTRL scroll" hotkey for font size adjustment.
+    consoleTextPane.addMouseWheelListener((MouseWheelEvent e) -> {
+      if (e.isControlDown()) {
+        if (e.getWheelRotation() < 0) {
+          base.handleFontSizeChange(1);
+        } else {
+          base.handleFontSizeChange(-1);
+        }
+      } else {
+        e.getComponent().getParent().dispatchEvent(e);
+      }
+    });
+
+    // Add "CTRL (SHIFT) =/+" and "CTRL -" hotkeys for font size adjustment.
+    consoleTextPane.addKeyListener(new KeyAdapter() {
+      @Override
+      public void keyPressed(KeyEvent e) {
+        if (e.getModifiersEx() == KeyEvent.CTRL_DOWN_MASK
+            || e.getModifiersEx() == (KeyEvent.CTRL_DOWN_MASK | KeyEvent.SHIFT_DOWN_MASK)) {
+          switch (e.getKeyCode()) {
+            case KeyEvent.VK_PLUS:
+            case KeyEvent.VK_EQUALS:
+              base.handleFontSizeChange(1);
+              break;
+            case KeyEvent.VK_MINUS:
+              if (!e.isShiftDown()) {
+                base.handleFontSizeChange(-1);
+              }
+              break;
+          }
+        }
+      }
+    });
   }
 
   public void applyPreferences() {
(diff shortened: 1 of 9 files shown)


facchinm:
@ArduinoBot build this please

app/src/processing/app/EditorConsole.java
```diff
@@ -110,18 +113,86 @@ public EditorConsole() {
...
+      if (e.isControlDown()) {
+        if (e.getWheelRotation() < 0) {
+          base.handleFontSizeChange(1);
+        } else {
+          base.handleFontSizeChange(-1);
+        }
+      } else {
+        e.getComponent().getParent().dispatchEvent(e);
+      }
+    });
```
cmaglie: I see this MouseWheelEvent repeated around, may this be factored out?

app/src/processing/app/EditorConsole.java
```diff
@@ -110,18 +113,86 @@ public EditorConsole() {
...
+              break;
+            case KeyEvent.VK_MINUS:
+              if (!e.isShiftDown()) {
+                base.handleFontSizeChange(-1);
+              }
+              break;
+          }
+        }
+      }
+    });
```
cmaglie: Same for this KeyAdapter, maybe it can be written just once inside Base?

app/src/processing/app/EditorConsole.java
```diff
@@ -110,18 +113,86 @@ public EditorConsole() {
...
+          document.insertString(start, text, newAttrs);
+        } catch (BadLocationException e) {
+          // Should only happen when text is async removed (through clear()).
+          // Accept this case, but throw an error when text could mess up.
+          if (document.getLength() != 0) {
+            throw new Error(e);
+          }
+        }
+      }
+    }
```
cmaglie: May you explain why you do this loop backwards? There is no other way to do it all at once (I guess no because the text may have multiple attributes like colors etc.)?

Pieter12345: I've tried looping forward, but since the text is taken out and re-inserted, getting the next offset will bug out. I am uncertain why this happens, but looping from back to front ensures that re-inserted parts will never be visited twice or affect next parts in the loop.

app/src/processing/app/EditorConsole.java
Pieter12345: This crossed my mind already, I will look into this.

Pieter12345: Done in 2d593c4.

app/src/processing/app/EditorConsole.java
Pieter12345: Done in 2d593c4.

app/src/processing/app/EditorConsole.java
Pieter12345: And indeed, the text has to be handled per document element due to it containing info and error text, which has a different color that should be maintained.

cmaglie:
Now I have a failing test here:

```
    [javac] Compiling 41 source files to <http://jenkins.arduino.cc:8080/job/arduino-linux64/ws/app/test-bin>
    [javac] <http://jenkins.arduino.cc:8080/job/arduino-linux64/ws/app/test/processing/app/UpdateTextAreaActionTest.java>:17: error: constructor AbstractTextMonitor in class AbstractTextMonitor cannot be applied to given types;
    [javac]       super(boardPort);
    [javac]       ^
    [javac]   required: Base,BoardPort
    [javac]   found: BoardPort
    [javac]   reason: actual and formal argument lists differ in length
    [javac] 1 error
```

I'm wondering if passing Base in the constructor is the correct thing to do.

Pieter12345:
@cmaglie Ah I see, for some reason I didn't have `app/test` added as a source directory. Usually, I would use Mockito to create a mock object for the `Base` class and pass this in the test. However, this project does not have Mockito as a dependency. How do you think about adding Mockito as a dependency to support JUnit tests and increase testability? Another option would be to allow a null `Base` in the constructor and do not add these listeners in this case.

cmaglie:
Either options looks good to me (Mockito and checking for `null`-ness of Base).

I was wondering about a third option: now `AbstractTextMonitor` depends on Base because it has to call this method `base.addEditorFontResizeListeners(...)`, what about inverting the dependency, for example we could add a `addMouseWheelListener(..)` method to `AbstractTextMonitor` and call it from the outside (where the monitor is created) to set the wheel listener.
Also, now that I'm thinking more about it, `AbstractTextMonitor` is a JFrame that is in turn a `Component`, it should already have this method.

PR #8776, opened 2019-04-12
Support .tar.xz for ArchiveExtractor to make download more fast

vowstar:
Now ```tar.xz``` format is widely used, and the official arduino
IDE download URL also shows that arduino uses ```tar.xz``` format.
(https://downloads.arduino.cc/arduino-1.8.9-linux64.tar.xz).

As we all know, the tar.xz format has the optimal size compared to tar,
tar.gz, tar.bz2, and zip.
(https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/)

Therefore, it is very unreasonable not to support tar.xz.

Supporting this format can save almost half of the bandwidth resources
and download time when compressing the gcc toolchain,
making users more comfortable.

Screenshot from 2019-04-12 12-00-59

### DIFF
--- a/arduino-core/src/cc/arduino/utils/ArchiveExtractor.java
+++ b/arduino-core/src/cc/arduino/utils/ArchiveExtractor.java
@@ -36,6 +36,7 @@ import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.zip.ZipArchiveInputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
 import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
+import org.apache.commons.compress.compressors.xz.XZCompressorInputStream;
 import org.apache.commons.compress.utils.IOUtils;
 import processing.app.I18n;
 import processing.app.Platform;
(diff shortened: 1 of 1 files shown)


vowstar:
@cmaglie

facchinm:
I think that adding tar.xz support can be a good idea; the only problem is that it introduces a minimum IDE version for the cores using it, and we are currently missing a flag to tell the user about this.
Adding an entry like `minimum_ide_version` in the json format won't solve the issue anyway since the older IDEs couldn't interpret it anyway, resulting in strange and unexplainable errors).
@cmaglie any idea? Should we add support on the cli maybe and use the switch as a turning point?

sandeepmistry:
Is there any way we could have a fallback URL?

vowstar:
A possible suggestion:
Let the server return a specific json according to the arduino version, only return the tar.xz with the version number greater than a certain value, and the rest return tar.bz2/gz or zip. Users using the latest version of arduino can add a suffix to the URL to let the server recognize arduino version.

facchinm:
The fallback/alternative url could work; I'm a bit scared in replacing json entries on the fly but it could work if handled correctly. Anyway, if we choose this way, we should apply all the relevant patches to https://github.com/arduino/arduino-cli first of all

cmaglie:
IMHO we should just add support for .xz **without** changing any existing json.
We can't swtich to .xz on the main Arduino package index because this will cause a massive amount of support requests, it's not worth the gain.

vowstar:
Is there any update?

vowstar:
@cmaglie @facchinm @sandeepmistry Hello, is there any plan to merge this branch?

PR #8781, opened 2019-04-12
Linux build: remove symlinks only if they exist

MrMebelMan:
The message `"Removing symlink failed. Hope that's OK. If not then rerun as root with sudo."` appears even when there's no symlink (for example, when installing for the first time). I don't think that a user should be adviced to run it as root in this case.
* check if symlink exists before trying to remove it

### DIFF
--- a/build/linux/dist/install.sh
+++ b/build/linux/dist/install.sh
@@ -152,8 +152,8 @@ xdg_uninstall_f() {
 
   # Remove symlink for arduino
   echo "" # Ensure password request message is on new line
-  if ! rm /usr/local/bin/arduino; then
-      echo "Removing symlink failed. Hope that's OK. If not then rerun as root with sudo."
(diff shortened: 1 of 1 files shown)


MrMebelMan:
Since the user will probably still need superuser privilages later, maybe it's also a good idea to use
```bash
ln -s foo bar || sudo ln -s foo bar
```
instead of just `ln -s foo bar`  when creating a symlink?

So it will ask the user for sudo password only if needed.

facchinm:
@MrMebelMan thanks for the PR! I'm merging it right now; about the second part, it would complicate the script a bit but could be a good idea for systems lacking `sudo` (not a very wide problem probably :wink: )

MrMebelMan:
Thanks :pray:

luisfigo27:
I had this error, how to delete the file? 
image

image

Issue #8778, opened 2019-04-12
Arduino not built for ppc64el (OpenPOWER) systems


madscientist159:
There are no prebuilt packages available for ppc64el (OpenPOWER) systems, despite the software working on those platforms (after some significant manual work to pull in dependencies using the correct architecture).

OpenPOWER systems are being deployed more widely, especially in education, and having an easy way to install a newer IDE version than the 1.5 version in Debian Experimental would be much appreciated.

madscientist159:
Note that the Arduino build process pulls in a lot of precompiled files from Arduino servers, so we will need assistance in getting those files updated.  We can provide free long term access to ppc64el build hardware if needed in support of this update.

In the interim, here's the first patch that would enable support for the ppc64el architecture.  It will fail immediately due to the Artistic Style binary package on the Arduino servers not containing the ppc64el shared library, however.

[arduino-initial-patch.txt](https://github.com/arduino/Arduino/files/3073080/arduino-initial-patch.txt)

PR #8782, opened 2019-04-13
Enable build on ppc64le systems

madscientist159:
This fixes arduino/Arduino#8778

The other pull requests listed will need to be handled first, and the new ppc64le binaries uploaded, but the IDE works fine on a POWER9 Blackbird desktop using locally built binaries and this patch.


madscientist159:
Pushed correct version, tested with Mega board.

madscientist159:
Can we get this merged?

madscientist159:
Can we get this merged please?

PR #8808, opened 2019-04-21
Bump MULTIPLE dependencies - constantly updating

szb512:
This will bump 3 dependencies to the latest version.


facchinm:
Hi @szb512 ,
since we don't require maven or any other kind of dependency manager, all jars binaries should also be updated in the project. Would you mind adding them too? Thanks!

szb512:
Done. Added jar files.

szb512:
Since there has been no response in the past year, closing because of staleness.

PR #9093, opened 2019-07-23
Fix log path when you run `ant run`

mattiabertorello:
When the IDE was started with `ant run` the log folder was wrong due to a premature initialization of the logging environment before set the log directory path.

### DIFF
--- a/app/src/log4j2.xml
+++ b/app/src/log4j2.xml
@@ -8,8 +8,8 @@
     </Console>
 
     <!-- Rolling File Appender -->
-    <RollingFile name="RollingFile" fileName="${sys:log4j.saveDirectory}/logs/application.log"
-                 filePattern="${sys:log4j.saveDirectory}/logs/application-%d{MM-dd-yyyy}-%i.log.gz"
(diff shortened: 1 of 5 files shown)


cmaglie:
I've cherry-picked the first two commits but still not working for me, it still creates the `${sys:log4j.dir}`.

mattiabertorello:
The last two commits fix the problem this is the line that must delete it https://github.com/arduino/Arduino/pull/9093/commits/9dd4c735a8ca9bac8a4aec9f01dd552c927ca9d0#diff-6ccded341b7bbb710383be5a14f6529aL63

To address better this problem you need to refactor the setting folder path generation in a way that can be used in the custom configuration of log4j https://logging.apache.org/log4j/2.x/manual/customconfig.html

cmaglie:
Ah now I see, that's the reason why you also extracted BoardCloudResolver! it wasn't clear at a first sight they looked more like two unrelated changes, this is the kind of things that it's worth explaining it in the commit message.

mattiabertorello:
Yes you are right, I should put in commit message

PR #9096, opened 2019-07-24
Use contributions.trust.all and download the signature in any case

mattiabertorello:
This PR will remove the signature check if `allow_insecure_packages ` or `contributions.trust.all` is true but the signature will be downloaded in any case


arduino-core/src/cc/arduino/contributions/DownloadableContributionsDownloader.java
```diff
@@ -215,6 +214,12 @@ public boolean checkSignature(MultiStepProgress progress, URL signatureUrl, Prog
       // Download signature
       download(signatureUrl, packageIndexSignatureTemp, progress, statusText, progressListener, true);
 
+      if (skipVerification) {
+        log.info("Allowing insecure packages because allow_insecure_packages is set to true in preferences.txt" +
+          " but the signature was download");
+        return true;
```
cmaglie: Returning `true` here will not update the `.sig` file if the signature is good.
IMHO we should just do everything as normal here (and do not consider the skipVerification preference).

arduino-core/src/cc/arduino/contributions/DownloadableContributionsDownloader.java
```diff
@@ -199,11 +200,9 @@ public boolean verifyDomain(URL url) {
...
 
     final boolean allowInsecurePackages =
-      PreferencesData.getBoolean("allow_insecure_packages", false);
-    if (allowInsecurePackages) {
-      log.info("Allow insecure packages is true the signature will be skip and return always verified");
-      return true;
-    }
+      PreferencesData.getBoolean(Constants.ALLOW_INSECURE_PACKAGES, false);
+    final boolean trustAll = PreferencesData.getBoolean(Constants.PREF_CONTRIBUTIONS_TRUST_ALL);
+    final boolean skipVerification = allowInsecurePackages || trustAll;
```
cmaglie: What about adding a method `PreferencesData.areInsecurePackagesAllowed()` and call that method every time we need to check these preferences? something along the line of:

```
public static boolean areInsecurePackagesAllowed() {
  if (getBoolean("allow_insecure_packages", false)) {
    log.info("allow_insecure_packages preference is true, the signature check will be ignored.");
    return true;
  }
  if (getBoolean("contributions.trust.all", false)) {
    log.info("contributions.trust.all preference is true, the signature check will be ignored.");
    return true;
  }
  return false;
}
```
We could remove also the two `Constants.ALLOW_INSECURE_PACKAGES` and `Constants.PREF_CONTRIBUTIONS_TRUST_ALL`.

mattiabertorello: Done https://github.com/arduino/Arduino/pull/9096/commits/e0374c115a5e842e4f2d2f4438676b4679ed15e5

PR #9158, opened 2019-08-20
Make update boards and libraries startup dialog accessible - 

joew46167:
Please see [design at Google developers](https://groups.google.com/a/arduino.cc/d/msgid/developers/3ae1d74f-ea86-46e2-a650-6a2896a5d3a0%40arduino.cc?utm_medium=email&utm_source=footer) for design description
### All Submissions:

### New Feature Submissions:

1. [x] Does your submission pass tests?

I'll update the first check box after test have completed

### Changes to Core Features:

I ran tests on Windows with NVDA and JAWS

### DIFF
--- a/app/src/cc/arduino/view/preferences/Preferences.java
+++ b/app/src/cc/arduino/view/preferences/Preferences.java
@@ -831,8 +831,6 @@ public class Preferences extends javax.swing.JDialog {
 
     PreferencesData.setBoolean("update.check", checkUpdatesBox.isSelected());
 
-    PreferencesData.setBoolean("editor.save_on_verify", saveVerifyUploadBox.isSelected());
-
     PreferencesData.setBoolean("ide.accessible", accessibleIDEBox.isSelected());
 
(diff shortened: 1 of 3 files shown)


app/src/cc/arduino/view/preferences/Preferences.java
```diff
@@ -826,7 +831,7 @@ private void savePreferencesData() {
 
     PreferencesData.setBoolean("update.check", checkUpdatesBox.isSelected());
 
-    PreferencesData.setBoolean("editor.save_on_verify", saveVerifyUploadBox.isSelected());
```
facchinm: I think this line should be preserved; it's probably a copy-paste error

joew46167: I removed that line because it was identical to the line 2 above it, and it didn't seem to need to be there twice.

facchinm: Ah great, I didn't catch it. Would you mind creating a separate commit the removes that line so the history is cleaner?

PR #9164, opened 2019-08-21
Add Buttons in Library and Boards managers

joew46167:
Using the new "use accessibility features" settings option, when the option is selected then instead of more info and online help links in the description there are now buttons for these features that do the same thing as the links.

As a side note, when the option is not selected, the links will now include translated strings for the links.  This made coding the buttons easier, and imo makes the links better.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Tested in Windows with NVDA.

### DIFF
--- a/app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
+++ b/app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
@@ -57,11 +57,14 @@ import cc.arduino.contributions.packages.ContributedHelp;
 import cc.arduino.contributions.packages.ContributedPlatform;
 import cc.arduino.contributions.ui.InstallerTableCell;
 import processing.app.Base;
+import processing.app.PreferencesData;
 import processing.app.Theme;
 
 @SuppressWarnings("serial")
 public class ContributedPlatformTableCellJPanel extends JPanel {
 
+  final JButton moreInfoButton;
+  final JButton onlineHelpButton;
   final JButton installButton;
   final JButton removeButton;
   final Component removeButtonPlaceholder;
@@ -72,6 +75,8 @@ public class ContributedPlatformTableCellJPanel extends JPanel {
   final JPanel buttonsPanel;
   final JPanel inactiveButtonsPanel;
   final JLabel statusLabel;
+  private final String moreInfoLbl = tr("More Info");
+  private final String onlineHelpLbl = tr("Online Help");
 
   public ContributedPlatformTableCellJPanel() {
     super();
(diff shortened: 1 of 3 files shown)


joew46167:
Some of the commits came from rebasing after my previous pull request was merged in.

joew46167:
Oh, I forgot to separate out another commit.

When these managers open the focus is set to the filter edit box, but the name isn't read.  So I added a description for that edit box.

joew46167:
Failed checks are all:

Use explicit scoping instead of the default package private level

for each of the lines where I declared a new button.  I debated between adding the word private to my buttons, or leaving them as they are without the word private, or adding the word private to all of the buttons to maintain code consistency.  I chose to just leave it as it is.

Please advise if I should change

facchinm:
> Please advise if I should change

Not necessary to change it, thanks

app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
```diff
@@ -216,16 +251,17 @@ void update(JTable parentTable, Object value, boolean isSelected,
     } else if (selected.getParentPackage().getHelp() != null) {
       help = selected.getParentPackage().getHelp();
     }
+
     if (help != null) {
       String url = help.getOnline();
       if (url != null && !url.isEmpty()) {
-        desc += " " + format("<a href=\"{0}\">Online help</a><br/>", url);
```
facchinm: The `<br/>` went missing in the refactor (it could be a good idea to include it in both the strings to keep things cleaner)

app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
```diff
@@ -149,6 +165,25 @@ public ContributedPlatformTableCellJPanel() {
...
+
+    if (accessibleIDE) {
+      button.setVisible(true);
+      button.addActionListener(e -> {
+        Base.openURL(url);
+      });
+    }
+    else {
+      // if not accessible IDE, keep link the same EXCEPT that now the link text is translated!
+      retString += " " + format("<a href=\"{0}\">{1}</a>", url, label);
```
facchinm: I'd change this into `retString += format("<a href=\"{0}\">{1}</a><br/>", url, label);` to target the other comment

app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
joew46167: Good call.  Great thing to have a good code review!

Addback back for both libraries.

app/src/cc/arduino/contributions/packages/ui/ContributedPlatformTableCellJPanel.java
joew46167: This is what I did in code change for other comment

facchinm: Not sure if the extra space at the beginning is still needed though... Better leave it there and eventually tackle it in a refactor to uniform the `setButtonOrLink` methods

joew46167: I don't believe that space is - or ever was - needed. That space is not use in the library manager version.

I've been looking for a utils class to move that method to, but I haven't found a home for it.

Do you want it changed? sounds like no.

PR #9168, opened 2019-08-23
Tools Boards menu scroll updates for accessibility

joew46167:
When a user uses the keyboard to navigate the menu, when you arrow up or down onto the scroll arrows, the scroll just starts scrolling the menu, scrolling a new board every 150 ms.  The board names don’t get read by the screen reader as they scroll by.  So the screen reader user has no idea what’s going on.  They are just on a menu item that says “menu item” at the screen reader.  If they hit the same arrow key again, they get to the Boards Manager… menu item, either by scrolling up or by scrolling down and wrapping around.  The wrapping around isn’t an issue – the user at least now knows where they are again because it says more than “menu item”
 
A big issue is that if they have a small screen and a very large number of board installed, it is possible that the board that they are looking for would scroll onto the visible part of the menu and then back off the visible part as the scroll continues.  After a bit of wondering what’s going on, because “screen item” made no sense, they may arrow back the other way to try to figure out what is going on.  Going the other way they might notice that the boards had changed, and they would think that that is odd.   And name of the board they want would never be read to them – it scrolled into and out of the visible area.  They know their board is installed, but they just can’t get to it.
 
Ideally, the menu wouldn’t scroll continuously.  Instead when they get to the end of the visible area and hit arrow again, only the next item would appear.   And I could easily make that happed by just removing the scroll – then the screen reader just keeps going through the list, but they end up scrolling off of the screen.  But for the low vision user that’s no good – they use both the screen reader and what they can see on the screen. Or if the person who is blind asks for sighted person’s help, the sighted person would want to see what’s going on.
 
So the problems isn’t that it goes back to the head of the list.  The problem is that the scroll doesn’t tell the user that it’s scrolling.  And as it scrolls it doesn’t tell the user WHAT is scrolling.  So this is intended to scroll the list one item at a time AND read that item name.  While still maintaining usability for low vision and sighted users.  
 
I hope that made sense.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/app/src/processing/app/tools/MenuScroller.java
+++ b/app/src/processing/app/tools/MenuScroller.java
@@ -280,9 +280,6 @@ public class MenuScroller {
       scrollCount = autoSizeScrollCount;
     }
 
-//    if (PreferencesData.getBoolean("ide.accessible")) {
-//      interval = 1000;
-//    }
     if (scrollCount <= 0 || interval <= 0) {
       throw new IllegalArgumentException("scrollCount and interval must be greater than 0");
     }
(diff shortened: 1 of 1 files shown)


joew46167:
Test in windows with NVDA and JAWS

Issue #8858, opened 2019-05-07
Board / Library Manager] split ridiculus long menus into sub-menus


datenheim:
The menu list of the board-manager is very (too?) long and scrolling in this menu is not convenient and seems to have (had) many bugs.

Indeed the Boards already are grouped by "Arduino AVR-Boards" etc.
These groups could/should simply be transformed to a submenu.

Thus boards are simpler and quicker to reach and scroll'n'search is reduced largely.

Same may happen with the library manager. There also a grouping is existing and could be used.
Libraries not fitting into the three groups could get an own submenu.

I have only ESP32, ESP8266 and Micro/Mini/MegaCore Boardmanagers and even on FHD+ screen the menu is 3x longer as the screen. Scrolling really takes some time and simply the wrong solution to this problem.

Related to all versions up to 1.8.9 on Windows and afaik also on Linux.

per1234:
>Scrolling really takes some time and simply the wrong solution to this problem.

Although not what you're requesting, this issue will be somewhat alleviated via https://github.com/arduino/Arduino/pull/8607

datenheim:
Yes, the feature mentioned in #8607 presenting recently used boards (libraries) more prominently will fit nicely with my request. There could well be a first submenu showing the recently used objects.

matthijskooijman:
For libraries, I implemented something similar in #1986, which divides puts libraries into submenus based on the current heading distinctions, as well as allowing them to be organized into subfolders under "libraries" (and these subfolders also cause submenus). The pull request was recently closed though, and was already quite outdated, but might still serve as inspiration in the future.

per1234:
The **Tools > Board** component of this request was fixed by https://github.com/arduino/Arduino/pull/9238 (thanks Matthijs!). The **Sketch > Include Library** component remains.

PR #9238, opened 2019-09-19
Divide board menu into submenus

matthijskooijman:
When multiple platforms (board packages) are installed, the boards menu quickly becomes very long and hard to use. This PR changes the menu to have a single submenu for each installed platform, under which the boards for that platform are listed. This makes the list a lot faster to navigate:

Screenshot from 2019-09-19 17-00-33

When there is just a single platform installed (e.g. on the default install), submenus do not really add anything, so the board menu still lists the boards directly as before (the only change is that the platform name is no longer shown at the top of the menu):

Screenshot from 2019-09-19 16-56-58

I think there has been some discussion about this in the past (see #8858, #1986 is also somewhat related). This also relates to keeping a recently-used-boards list (PR at #8607), either that or this PR might need some refactoring if both will be merged (but I needed this feature myself, so I thought I might as well share it here as well).

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1481,24 +1481,21 @@ public class Base {
     ButtonGroup boardsButtonGroup = new ButtonGroup();
     Map<String, ButtonGroup> buttonGroupsMap = new HashMap<>();
 
+    List<JMenu> platformMenus = new ArrayList<JMenu>();
+
     // Cycle through all packages
-    boolean first = true;
     for (TargetPackage targetPackage : BaseNoGui.packages.values()) {
       // For every package cycle through all platform
       for (TargetPlatform targetPlatform : targetPackage.platforms()) {
 
-        // Add a separator from the previous platform
-        if (!first)
-          boardMenu.add(new JSeparator());
-        first = false;
-
         // Add a title for each platform
         String platformLabel = targetPlatform.getPreferences().get("name");
-        if (platformLabel != null && !targetPlatform.getBoards().isEmpty()) {
-          JMenuItem menuLabel = new JMenuItem(tr(platformLabel));
-          menuLabel.setEnabled(false);
-          boardMenu.add(menuLabel);
-        }
+        if (platformLabel == null)
+          platformLabel = targetPackage.getId() + "-" + targetPlatform.getId();
+
+        JMenu platformBoardsMenu = new JMenu(tr(platformLabel));
+        MenuScroller.setScrollerFor(platformBoardsMenu);
+        platformMenus.add(platformBoardsMenu);
 
         // Cycle through all boards of this platform
         for (TargetBoard board : targetPlatform.getBoards().values()) {
@@ -1507,14 +1504,27 @@ public class Base {
           JMenuItem item = createBoardMenusAndCustomMenus(boardsCustomMenus, menuItemsToClickAfterStartup,
                   buttonGroupsMap,
                   board, targetPlatform, targetPackage);
-          boardMenu.add(item);
+          platformBoardsMenu.add(item);
           boardsButtonGroup.add(item);
         }
       }
     }
 
+    JMenuItem firstBoardItem = null;
+    for (JMenu platformMenu : platformMenus) {
+      if (firstBoardItem == null && platformMenu.getItemCount() > 0)
+        firstBoardItem = platformMenu.getItem(0);
+      boardMenu.add(platformMenu);
+    }
+
+    if (firstBoardItem == null) {
+      throw new IllegalStateException("No available boards");
+    }
+
+    // If there is no current board yet (first startup, or selected
+    // board no longer defined), select first available board.
     if (menuItemsToClickAfterStartup.isEmpty()) {
-      menuItemsToClickAfterStartup.add(selectFirstEnabledMenuItem(boardMenu));
+      menuItemsToClickAfterStartup.add(firstBoardItem);
     }
 
     for (JMenuItem menuItemToClick : menuItemsToClickAfterStartup) {
@@ -1666,16 +1676,6 @@ public class Base {
       return firstVisible;
     }
 
-    throw new IllegalStateException("Menu has no enabled items");
-  }
-
-  private static JMenuItem selectFirstEnabledMenuItem(JMenu menu) {
-    for (int i = 1; i < menu.getItemCount(); i++) {
-      JMenuItem item = menu.getItem(i);
-      if (item != null && item.isEnabled()) {
-        return item;
-      }
-    }
     throw new IllegalStateException("Menu has no enabled items");
   }
 
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1493,7 +1493,7 @@ public class Base {
         if (platformLabel == null)
           platformLabel = targetPackage.getId() + "-" + targetPlatform.getId();
 
-        JMenu platformBoardsMenu = new JMenu(tr(platformLabel));
+        JMenu platformBoardsMenu = new JMenu(platformLabel);
         MenuScroller.setScrollerFor(platformBoardsMenu);
         platformMenus.add(platformBoardsMenu);
 
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1510,6 +1510,8 @@ public class Base {
       }
     }
 
+    platformMenus.sort((x,y) -> x.getText().compareToIgnoreCase(y.getText()));
+
     JMenuItem firstBoardItem = null;
     if (platformMenus.size() == 1) {
       // When just one platform exists, add the board items directly,
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1511,10 +1511,21 @@ public class Base {
     }
 
     JMenuItem firstBoardItem = null;
-    for (JMenu platformMenu : platformMenus) {
-      if (firstBoardItem == null && platformMenu.getItemCount() > 0)
-        firstBoardItem = platformMenu.getItem(0);
-      boardMenu.add(platformMenu);
+    if (platformMenus.size() == 1) {
+      // When just one platform exists, add the board items directly,
+      // rather than using a submenu
+      for (Component boardItem : platformMenus.get(0).getMenuComponents()) {
+        boardMenu.add(boardItem);
+        if (firstBoardItem == null)
+          firstBoardItem = (JMenuItem)boardItem;
+      }
+    } else {
+      // For multiple platforms, use submenus
+      for (JMenu platformMenu : platformMenus) {
+        if (firstBoardItem == null && platformMenu.getItemCount() > 0)
+          firstBoardItem = platformMenu.getItem(0);
+        boardMenu.add(platformMenu);
+      }
     }
 
     if (firstBoardItem == null) {
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -999,9 +999,9 @@ public class Base {
     // kill uploader (if still alive)
     UploaderUtils uploaderInstance = new UploaderUtils();
     Uploader uploader = uploaderInstance.getUploaderByPreferences(false);
-    if (uploader != null && uploader.programmerPid != null && uploader.programmerPid.isAlive()) {
+    if (uploader != null && Uploader.programmerPid != null && Uploader.programmerPid.isAlive()) {
         // kill the stuck programmer
-        uploader.programmerPid.destroyForcibly();
+        Uploader.programmerPid.destroyForcibly();
     }
 
     if (handleQuitEach()) {
@@ -1444,8 +1444,9 @@ public class Base {
         String filterText = "";
         String dropdownItem = "";
         if (actionevent instanceof Event) {
-          filterText = ((Event) actionevent).getPayload().get("filterText").toString();
-          dropdownItem = ((Event) actionevent).getPayload().get("dropdownItem").toString();
+          Event e = ((Event) actionevent);
+          filterText = e.getPayload().get("filterText").toString();
+          dropdownItem = e.getPayload().get("dropdownItem").toString();
         }
         try {
           openBoardsManager(filterText, dropdownItem);
@@ -1481,7 +1482,7 @@ public class Base {
     ButtonGroup boardsButtonGroup = new ButtonGroup();
     Map<String, ButtonGroup> buttonGroupsMap = new HashMap<>();
 
-    List<JMenu> platformMenus = new ArrayList<JMenu>();
+    List<JMenu> platformMenus = new ArrayList<>();
 
     // Cycle through all packages
     for (TargetPackage targetPackage : BaseNoGui.packages.values()) {
@@ -1602,7 +1603,7 @@ public class Base {
           };
           List<TargetBoard> boards = (List<TargetBoard>) subAction.getValue("board");
           if (boards == null) {
-            boards = new ArrayList<TargetBoard>();
+            boards = new ArrayList<>();
           }
           boards.add(board);
           subAction.putValue("board", boards);
@@ -2003,6 +2004,7 @@ public class Base {
               Base.this.handleFontSizeChange(-1);
             }
             break;
+          default:
           }
         }
       }
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -761,6 +761,20 @@ public class Editor extends JFrame implements RunnerListener {
     toolsMenu.add(item);
 
     toolsMenu.addMenuListener(new StubMenuListener() {
+      public JMenuItem getSelectedItemRecursive(JMenu menu) {
+        int count = menu.getItemCount();
+        for (int i=0; i < count; i++) {
+          JMenuItem item = menu.getItem(i);
+
+          if ((item instanceof JMenu))
+            item = getSelectedItemRecursive((JMenu)item);
+
+          if (item != null && item.isSelected())
+            return item;
+        }
+        return null;
+      }
+
       public void menuSelected(MenuEvent e) {
         //System.out.println("Tools menu selected.");
         populatePortMenu();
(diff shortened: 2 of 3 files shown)


facchinm:
I tested the patch and have one major concern; when you select a board, it doesn't appear anymore after "Board" label, just on the bottom right.

matthijskooijman:
> I tested the patch and have one major concern; when you select a board, it doesn't appear anymore after "Board" label, just on the bottom right.

Good point. It took me a while to find where this was done (ended up grepping for `": \"` to find it...), it is a bit obscure code (IMHO it might be worth refactoring at some point to be a bit more explicit and less magical, but some refactoring might be good in all this menu code...).

In any case, I fixed the code to support nested menus now, I added a new commit for that before the existing commits (to make sure it is never broken in the history. The two existing commits are unchanged).

There is a small behavioral change (see commit message: e5551bf2), removing a check that seems unneeded. @PaulStoffregen, you originally added this code. Do you happen to remember why this `if` is needed (i.e. when `item.getText()` could be null): https://github.com/arduino/Arduino/blob/6116a8e9206124329e0c8b58e828eb5b83b6055a/app/src/processing/app/Editor.java#L780-L781

datenheim:
> white_check_mark Build completed.
> Please test this code using one of the following:
> 
> arrow_down http://downloads.arduino.cc/javaide/pull_requests/arduino-PR-9238-BUILD-892-windows.zip

Tested on Windows 7-64 Bit, works like a charm. Hope it makes it to the release soon :)

This in fact solves half of my feature request #8858.

Thank you.

ladyada:
sending good vibes to this PR - we have to QA against so many packages it takes 15 seconds to scroll thru them all :D this is gonna be awesome!

datenheim:
... and get even more useful when combined with the recently used boards #8607 which made it to the current beta already. Both features might be a good add to the insert library menu also.

datenheim:
> 
> 
> I tested the patch and have one major concern; when you select a board, it doesn't appear anymore after "Board" label, just on the bottom right.

Yes, that is true, and it is a no-go. But - the split menu itself ist great :)

datenheim:
> I tested the patch and have one major concern; when you select a board, it doesn't appear anymore after "Board" label, just on the bottom right.

@facchinm Indeed correct, this is a no-go. Overlooked that because I was so happy about the split board menu :(

facchinm:
@matthijskooijman did you forget to push the fix commit? :slightly_smiling_face:

matthijskooijman:
> @matthijskooijman did you forget to push the fix commit? slightly_smiling_face

I was sure I pushed it, turns out I accidentally pushed to this repository rather than my own fork, w00ps. Anyway, fixed now, the extra fix commit is now added to this PR :-)

facchinm:
@tigoe @Giorgio-Olivero what do you think about this? Do you think it will improve the usability even for people with 2/3 cores installed?

tigoe:
I like the concept, there are a couple of implementation issues I noticed:

When using the menus with VoiceOver on MacOS, there's a weird behavior that happens when you back out of a menu. Normally with VO, you hold the ctrl-option keys and navigate menus with the arrows. The main menu drops entirely rather than going back up one sub-menu.  I think it'd be pretty common to browse the menus up and down a bit looking for your board, so making the arrow keys do the same traversing of the sub-menus that you'd do with a mouse would be appropriate, I think.

It brings up another issue I've seen with students and the Boards menu: most casual users and beginners don't know the architectures, they product lines. So they'd expect a Nano Every, for example, to be next to a Nano 33 IoT or BLE. Same with the Adafruit Feathers, the ESPs, etc. Sometimes that happens by virtue of being the same processor architecture, but often not. It would be handy to be able to sort the Boards submenus by product family or alphabetically, not just by architecture.

tigoe:
Here's a test on that VoiceOver issue:

* MacOS Mojave, turn on VO using command-F5
* press ctrl-option-M to turn on voiceOver
* press right arrow 5 times to get to tools menu
* press down arrow until you're on Boards
* press right arrow to open submenu
* press down arrow to get to the Arduino SAMD boards
* press right arrow to scroll through them
* press left arrow to go up a submenu

At this last step, the entire menu closes on some cases without selecting a board. Doesn't seem to be all cases, though, and I can't identify what the connection between them is yet.

tigoe:
@facchinm I also wanted to check your question of how it would be for users with only 2-3 cores. So I removed my Arduino15 folder and restarted. The default install only has the AVR boards, so the menu looks just like it does currently. When you add a core, it groups them by submenu appropriately, and seems convenient enough. So I'd say it works okay for both beginners and power users, once the glitch with voiceOver is solved.  

I'd still add the "sort by..." option, but that should properly be a separate issue.

matthijskooijman:
> When using the menus with VoiceOver on MacOS, 

Is this specific to using VO? AFAICS, VO means that OSX will read out selected items to you while navigating by keyboard. But you can navigate the menus already without VO, I think? Does it then also "break"? Or does VO actually change the way keyboard navigation works (I'm not a Mac user, so I'm not quite familiar with this).

Does it work properly with other menus inside the IDE? I'm wondering what would be different, since we essentially just create some nested JMenu/JMenuItem instances and let java handle the rest, so I wonder if this might just be the way Java handles this and whether there is anything we can do to influence how this works...

> It brings up another issue I've seen with students and the Boards menu: most casual users and beginners don't know the architectures, they product lines. So they'd expect a Nano Every, for example, to be next to a Nano 33 IoT or BLE. Same with the Adafruit Feathers, the ESPs, etc. Sometimes that happens by virtue of being the same processor architecture, but often not. It would be handy to be able to sort the Boards submenus by product family or alphabetically, not just by architecture.

With "Sort", do you then mean changing the order of items? Or also changing the grouping in submenus? Grouping of boards seems to have been tried at some point in https://github.com/arduino/Arduino/pull/120, however it seems that the custom menus have been used instead to solve that particular usecase (e.g. grouping Pro Mini 8Mhz and Pro Mini 16Mhz together).

To really make finding boards easy, one could imagine having a standalone board browser window, where you could switch between different orderings and groupings and filter by name, etc. but that is not so suitable for inside a drop-down menu. And as you said, I think this is really a separate thing that might need to be discussed (and implemented) separately later.

tigoe:
The submenu and arrow issue seems to be the same when navigating with arrows without voice over on. On Windows, you could probably see the same with a screen reader turned on there. NVDA or JAWS would be good tests of it, but I don't have a Windows machine at home at the moment.

matthijskooijman:
> The  submenu and arrow issue seems to be the same when navigating with  arrows without voice over on. On Windows, you could probably see the  same with a screen reader turned on there.

Huh? You're saying you do *not* need Voice Over to reproduce this problem, so why would one need a screenreader on Windows? Or did I misunderstand?

> NVDA or JAWS would be good  tests of it, but I don't have a Windows machine at home at the moment.

I don't have Windows at hand either, nor a screenreader.

I did test on Linux (Gnome 3), without a screenreader, and there keyboard navigation works as expected (press alt-T to activate the tools menu, navigate into the subsubmenu, press left and only the innermost menu closes).

tigoe:
I think it should be tested on the other OSes, and with screen readers, before merging. The fact that I got unexpected behavior means others might too.

matthijskooijman:
I just tested the build on Windows 10 (without screenreader, I have no admin access here to install anything) and the arrow-key-behaviour is also ok.

matthijskooijman:
@tigoe, Another question: Is this unwanted arrow-key behaviour specific to the (extra nested) boards menu? Or did it occur already with the unmodified releases builds of Arduino? In the latter case, this might be a separate issue (though this PR does make the issue more pronounced because of the additional layer of nested menus, of course)?

tigoe:
That's good. Windows 10 has a built-in screen reader called Narrator. Give
that a try.

T

matthijskooijman:
> That's good. Windows 10 has a built-in screen reader called Narrator. Give
that a try.

Ah, thanks. Just tried that, and could not really get Narrator to speak out anything from the Arduino interface, it seems. Also, unlike your voice-over thing, I think there is no hotkey to activate Narrator to navigate through the menubar. I could not reproduce your problem with Narrator running.

However, I did notice this:
 - If you open up the tools menu
 - Then move the mouse over the "Board" item
 - Then close the tools menu, and reopen it using alt+T
 - Then the mouse is still over the board item, so the board submenu opens, but no item is selected yet
 - If you now press right-arrow, instead of entering the submenu, it navigates from Tools to Help in the top menu bar. Pressing down or up rather than right does enter the menu as expected (and then pressing left leaves it as expected).

Maybe this is also what you've been seeing? Or perhaps the mouse position at least explains the erraticness of what you've been seeing? This behaviour happens here on Windows 10 both with or without Narrator, both with the 1.8.9 release and this PR build. It does not happen with menus in Chrome or regedit, so I suspect this might be specific to the Java implementation of menus.

tigoe:
That's weird. It might be what I was seeing. When I'm back in proximity of
a laptop on Monday I'll give it a shot and see, thanks.

T

matthijskooijman:
I've just did another force-push to remove an unrelated commit about String examples I accidentally added, and to add a fixup commit to slightly simplify and optimize some code and fix an unused variable warning pointed out by Codacy at the same time.

The fixup commit should be squashed before merging this commit.

tigoe:
I ran your latest build this morning on MacOS and tested with VoiceOver again. Couldn't replicate the problem I was hitting last week. Either your cleanup fixed it or I was imagining things in my earlier test, but either way we can close that thread.

The Adafruit Circuit Playground Express showed up in the Arduino SAMD submenu, whereas the rest of the Adafruit boards showed in their own Adafruit SAMD submenu. I was a little confused by that, but it's not a show stopper for me.

ladyada:
@tigoe that's correct, Adafruit Circuit Playground Express should appear under **Arduino SAMD** And Adafruit Circuit Playground will be under **Arduino AVR**

bpeschier:
> Here's a test on that VoiceOver issue:
> 
> * MacOS Mojave, turn on VO using command-F5
> * press ctrl-option-M to turn on voiceOver
> * press right arrow 5 times to get to tools menu
> * press down arrow until you're on Boards
> * press right arrow to open submenu
> * press down arrow to get to the Arduino SAMD boards
> * press right arrow to scroll through them
> * press left arrow to go up a submenu
> 
> At this last step, the entire menu closes on some cases without selecting a board. Doesn't seem to be all cases, though, and I can't identify what the connection between them is yet.

Hi, I tested this with the version you are referencing here and the later version and could not get it replicated with either. Only the submenu gets closed, which is the expected behaviour.

tigoe:
Good, thanks all for the confirmations. From there, I will defer to @facchinm.

matthijskooijman:
@tigoe, @bpeschier, thanks for reporting back on the voice-over issue. Seems that is not actually an issue then after all.

As far as I'm concerned, that resolves all open issues with this PR and it could be merged. I went ahead and rebased on top of master and squashed my fixup commit already, making it ready to merge (to see the history just before squashing, see f6f5a0eb9a7fc7c28f573b634dd6f95fc8edbcbf).

matthijskooijman:
I've just pushed another update to this PR. I found a small problem that occurred when no (valid) board was selected during startup, which I fixed in e94d676cb741a84ede41229c9a6b832dfb6d1aa9. I also rebased to master.

@facchinm, is there anything left before merging this?

datenheim:
Tested on win10 with this list of extra board managers:

https://adafruit.github.io/arduino-board-index/package_adafruit_index.json
http://downloads.arduino.cc/packages/package_mkr1000_index.json
https://raw.githubusercontent.com/ioteamit/smarteverything-core/master/package_arrow_index.json
https://mcudude.github.io/MegaCore/package_MCUdude_MegaCore_index.json
https://mcudude.github.io/MicroCore/package_MCUdude_MicroCore_index.json
https://mcudude.github.io/MightyCore/package_MCUdude_MightyCore_index.json
https://mcudude.github.io/MiniCore/package_MCUdude_MiniCore_index.json
http://arduino.esp8266.com/stable/package_esp8266com_index.json
https://dl.espressif.com/dl/package_esp32_index.json

Interestingly there appear two submenus "Arduino ESP 32" with overlapping entries, see screen shots:
Anmerkung 2020-01-31 085606
Anmerkung 2020-01-31 085509
I do not understand why that happens.


Furthermore it may be useful to sort the submenus alphabetically. At least I would find things faster that way....
... anyway still a major improvement right now :)

matthijskooijman:
> Interestingly there appear two submenus "ESP32 Arduino" with some overlapping entries (board that appear in both submenus), see screen shots:

I just tried installing the esp32 core through the board manager using the url you showed, which only shows up once in the boards menu. Could it be that you have installed the core twice, maybe once through the board manager and once using a manual download into your Sketchbook/hardware folder?

Regardless, I think this is not something caused by this PR, this duplication probably already happened before (but was not so clear because you could never see all of them in a single view anyway).

> Furthermore it may be useful to sort the submenus alphabetically. At least I would find things faster that way....

Good suggestion. This does mean that all platforms will be sorted individually (right now, they are grouped by packages, e.g. Arduino AVR and Arduino SAMD are both platforms aka architectures within the Arduino packages), but if the naming is consistent, they will probably still end up adjacent.

matthijskooijman:
I just rebased on top of master, and added alphabetical sorting of board submenus. For that, I had to somewhat refactor the generation of these board submenus, so just amended the existing commits for that.

datenheim:
> > Interestingly there appear two submenus "ESP32 Arduino" with some overlapping entries (board that appear in both submenus), see screen shots:
> 
> I just tried installing the esp32 core through the board manager using the url you showed, which only shows up once in the boards menu. Could it be that you have installed the core twice, maybe once through the board manager and once using a manual download into your Sketchbook/hardware folder?

Today I discovered that the boardmanager link:
[https://raw.githubusercontent.com/ioteamit/smarteverything-core/master/package_arrow_index.json](https://raw.githubusercontent.com/ioteamit/smarteverything-core/master/package_arrow_index.json)
when present under preferences/boardmanagers somehow creates the doubled 

With sorting I meant the submenus themselfes AS WELL as the entries in the submenus....
grafik
Thanks for working on this, by the way :)

matthijskooijman:
> Today I discovered that the boardmanager link:
ioteamit/smarteverything-core:package_arrow_index.json@master (raw)
when present under preferences/boardmanagers somehow creates the doubled

Hm, I cannot reproduce this (adding just the esp32 and smarteverything urls and installing the esp32 core). Could you check if this issue also occurs with a normal IDE version (i.e. without this PR applied)? If it does, please open a new issue for it, since then it only clouds discussion this PR. If you do, maybe you could create a zip of your ~/.arduino15 directory (or the similar directory on Windows in your userdir somewhere) and provide (might be a big zip, though).

> With sorting I meant the submenus themselfes AS WELL as the entries in the submenus....

I intentionally did not sort the boards within the menu, since this allows the maintainer of the board package to choose their own ordering (e.g. putting a few commonly used boards at the top).

datenheim:
> > With sorting I meant the submenus themselfes AS WELL as the entries in the submenus....
> 
> I intentionally did not sort the boards within the menu, since this allows the maintainer of the board package to choose their own ordering (e.g. putting a few commonly used boards at the top).

I Understand. Whatsoever the maintainer can not know how the community users are actually using the boards.... :)

Knowing it is complicated to implement, this could be overcome with a "recently used" submenu or something like that. But Idon't  dare to ask for that.

matthijskooijman:
> Knowing it is complicated to implement, this could be overcome with a "recently used" submenu or something like that. But Idon't dare to ask for that.

Actually, #8607 already implements that (will probable need some work if this PR is merged, though).

datenheim:
Eagerly waiting for a merge :heart_eyes:

matthijskooijman:
@cmaglie, seeing you're merging some PRs today, are you considering this one as well? As far as I'm concerned, it's ready to be merged, I've been running with this PR applied every since I created it.

Once this is merged, this opens up the way to rebasing #8607 and getting that into shape for merging.

cmaglie:
Rebased to master and added a commit: https://github.com/arduino/Arduino/pull/9238/commits/747ce0188b1ac4ba3e28891e2a5d91cf6a0666f8
```
Use un-translated labels to sort boards submenus

Otherwise it may happen some weird sorting when untraslated and
translated labels are sorted together:

    Arduino megaAVR Boards
    Arduino nRF52 Board
    ESP32 Arduino
    ESP8266 Modules
    Schede Arduino AVR   <-- the localized string falls to the bottom

all the Arduino boards should be packed together, so it's better to not
use localization at all while sorting. After this patch it will look like

    Schede Arduino AVR
    Arduino megaAVR Boards
    Arduino nRF52 Board
    ESP32 Arduino
    ESP8266 Modules
```

matthijskooijman:
Hm, is it not actually unexpected if the sorting does *not* follow the displayed strings? This might end up with grouping related platforms based on their untranslated names, but this might be unclear to a user.

I was going to say that sorting on translated strings should work fine if all platforms are consistently translated, but I think there is currently no way for third-party cores to supply any translations at all, so the only translations available are those which are supplied with the IDE (which might even be limited to the AVR core). Would it not make sense to simply not translate the platform names at all, so at least all are treated equally?

cmaglie:
The sorting should be based on the vendor/architecture and not on the word "Boards". What happens is that in english we have:

`Arduino AVR Boards`

since the word "Boards" is at the end of the string it basically doesn't influence the sorting but when we go to italian the word "Schede" is placed at the beginning of the string:

`Schede Arduino AVR`

So sorting in english happens to be correct but it may not be the same in other languages.
Another thing to consider is that some platforms doesn't have the word "Boards" at all in their name, so for example `ESP32 Arduino` will stay the same in every language and it will be presented before `Schede Arduino AVR` in italian. IMHO this is even *more* unexpected...

> but I think there is currently no way for third-party cores to supply any translations at all

**This** is a good point, I agree, better to remove the translation here.

app/src/processing/app/Base.java
```diff
@@ -1481,19 +1481,22 @@ public void actionPerformed(ActionEvent actionevent) {
...
-    List<JMenu> platformMenus = new ArrayList<JMenu>();
+    class PlatformJMenu extends JMenu {
+      public PlatformJMenu(TargetPlatform pl) {
+        String text = pl.getPreferences().get("name");
+        if (text == null) {
+          text = pl.getContainerPackage().getId() + "-" + pl.getId();
+        }
+        setText(text);
+      }
+    }
```
matthijskooijman: Now that untranslated strings are no longer needed, all this code seems no longer needed? This commit could just be a matter of removing `tr()`, no point in creating the `PlatformJMenu` class anymore I think?

cmaglie: done

PR #9262, opened 2019-09-27
Add titles to rows in the manager tables

joew46167:
When arrowing through the Library Manager table and the Board Manager table while using a screen reader, the rows are not read.  

By adding a TitleBorder the rows of the table get read.  

This currently does not work in MacOS - JDK 9+ is required for it to work on MacOS.  As such, I hope to soon write a replace to these managers for when accessibility features setting is turned on so that the current tables are avoided.  However this may take some time.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


facchinm:
Hi Joe, while I love the patch, visually it makes things a bit different from what it used to be
2019-09-30-101155_861x359_scrot
I kinda like the new look but maybe just exposing under `isAccessible()` flag could be better right now. What do you think? @tigoe

tigoe:
The bigger font is nice for those who are low vision, but the repeat of the text on two lines seems awkward. Is there a nightly build I can run it on to hear what it's like in a screen reader?

facchinm:
@ArduinoBot build this please

facchinm:
The build should be coming shortly :wink:

tigoe:
Thanks @facchinm Unfortunately the PR has the same problem as before, as Joe mentioned. It doesn't work on MacOS yet, so I can't test until I'm back on a PC. It behaves the same as previously: freezes up at the filter box.

joew46167:
I had at one point taken the title out of the actual cell and only left it in the TitleBorder.  But then it didn't read well in the screen reader, so I put it back into the cell.
Maybe taking it out of the cell, but putting it into the accessible text is the way to go?  I kinda dislike this idea because I'm not a fan of the accessible text not matching the on screen text.  But it might be a good way to go anyway?

    On Monday, September 30, 2019, 04:15:27 AM EDT, Martino Facchin <notifications@github.com> wrote:  
 
 
Hi Joe, while I love the patch, visually it makes things a bit different from what it used to be

I kinda like the new look but maybe just exposing under isAccessible() flag could be better right now. What do you think? @tigoe

tigoe:
If it's a choice between where to put the title, I actually prefer the larger type of the title in the bounding box, personally. I'd leave it there and cut the title that's just below it. But I recognize that there may be factors I'm not considering.

joew46167:
I have made a new push to my repository.

I remove the name from the cell, leaving it only in the title.  This is pretty good, but not perfect.

As I enter the cell it will read the title, then the contents of the cell string.
However, if I tab to the More Info button, then tab back to the cell string, it doesn't read the board name, only the string.  I can make it read the board name when doing this, but as you tab into the cell it reads the name twice, which is awkward.

Even at awkward in one use case, I still think this is the better way to go.

joew46167:
Is there an update on this?  More that I need to do?

facchinm:
I need to retest it in the next few days, sorry for the radio silence but we are a bit overwhelmed in this period :slightly_smiling_face:

facchinm:
Looks perfect now, merging. Thanks again!

PR #9272, opened 2019-10-01
Cleanup board/library manager color setting and somewhat support dark themes

matthijskooijman:
I'm running with a dark system theme, which defaults interfaces to have white/light grey on black/dark grey coloring. This works pretty well in the Arduino IDE, except for the library and board managers.

As an example, here's how the board manager looks using the default Gnome light theme (adwaita):

Default theme

However, when switching to a dark theme, adwaita-dark (can be done using the gnome-tweaks command), the background colors change but the foreground color stays at the (now) default light color:

Dark theme

While investigating, I also found that there is a lot of color-setting code that is not actually effective (because the colors are set on the wrong component, or are overwritten by other code). The first commit removes this useless code. See the commit message for a detailed analysis of how the code paths run.

The second commit does a bit of cleanup and refactoring needed for the last commit, which also sets the foreground color to black to match the background color. This results in a usable interface again:

Dark theme fixed

With the default light theme, the interface is unchanged. All of the above also applies to the library manager, which is fixed in the same way.

Note that this essentially forces a light theme for the board and library manager, but leaves the dropdowns and buttons unchanged (since it is probably tricky to completely undo the dark them for those).

An alternative solution would be only use the default colors, which would actually preserve the dark theme colors in these managers as well (rather than forcing black-on-white/grey as now). There are default colors for selected and non-selected table cells that could be used, but these are different from the current colors. Additionally, the current odd/even alternating colors are then also no longer available.

For completeness, here's how that would look in a light theme:

image

And in a dark theme:

image

### DIFF
--- a/app/src/cc/arduino/contributions/libraries/ui/ContributedLibraryTableCellEditor.java
+++ b/app/src/cc/arduino/contributions/libraries/ui/ContributedLibraryTableCellEditor.java
@@ -125,6 +125,7 @@ public class ContributedLibraryTableCellEditor extends InstallerTableCell {
     editorCell.versionToInstallChooser
         .setVisible(!mayInstalled.isPresent() && notInstalled.size() > 1);
 
+    editorCell.setForeground(Color.BLACK);
     editorCell.setBackground(new Color(218, 227, 227)); // #dae3e3
     return editorCell;
   }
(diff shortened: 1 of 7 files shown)


feikname:
If you plan on going with the second version I recommend switching the links color to something brighter (maybe brighter blue or orange?) in the dark theme so it doesn't hurt the eyes as much.

matthijskooijman:
> If you plan on going with the second version I recommend switching the links color to something brighter (maybe brighter blue or orange?) in the dark theme so it doesn't hurt the eyes as much.

Well, this is exactly the main reasons to *not* use the second version, since it is tricky to know what theme is being used (light and dark could maybe be autodetected with some smartness, but there might be completely different themes too). In this example I just used the default (theme) colors for background and normal text, but it seems links use a hardcoded color, or perhaps a default link color that is not changed with the theme. Figuring out a good color depending on the used theme is probably tricky...

PR #9361, opened 2019-10-27
Improve the blinking logic

lapo-luchini:
Blink on exact interval, avoid accumulating delays in case we miss the right millisec.
Also avoids a subtraction in the inner loop (and could be slightly more clear to understand?).

### All Submissions:


cmaglie:
The proposed change breaks the example (as explained in the previous comment). Closing as invalid.

PR #9487, opened 2019-12-02
 Changing URLs from HTTP to HTTPS

luigigubello:
I have changed the URLs from `http` to `https`. In particular the page `latest.txt` can be reached on both port 80 and port 443.


gvarisco:
LGTM

PR #9644, opened 2020-01-20
Do not clear serial monitor if the port is disabled for upload

### LINKED ISSUE #9234: Serial Monitor clears if port disappears in release 1.8.10
I have just updated from 1.8.9 to 1.8.10, and am having a problem with the serial monitor.
I am working on a system using motor shields and motors. To assist with debugging I have serial.print statements to output to the monitor. When things go wrong I either need to remove the motor power, or disconnect the usb to stop the hardware getting damaged. 
When I do this in 1.8.10 the displayed information disappears, meaning I can't track what has happened. This did not happen in 1.8.9.
Using Arduino Mega 2560
on Win 10 computer

cmaglie:
Fix #9234 

Used the patch provided by @PaulStoffregen

### DIFF
--- a/app/src/processing/app/AbstractMonitor.java
+++ b/app/src/processing/app/AbstractMonitor.java
@@ -1,7 +1,6 @@
 package processing.app;
 
 import cc.arduino.packages.BoardPort;
-import cc.arduino.packages.DiscoveryManager;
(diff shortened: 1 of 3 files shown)


cmaglie:
@arduinobot build this please

PR #9649, opened 2020-01-21
Improved sanity checks on archiveFileName field in package_index.json

cmaglie:
/cc @luigigubello


luigigubello:
For me it's ok, you can merge it

PR #9695, opened 2020-01-31
Update CommandLineTest.testCommandLineVersion()

tlk:
This makes the CommandLineTest.testCommandLineVersion work with the latest code.

Looks like the test was broken by the System.out.println statement in Base.java https://github.com/arduino/Arduino/commit/824567d763a29aaf439a271414c08d17ad521299

### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests]

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -207,9 +207,6 @@ public class Base {
     BaseNoGui.getPlatform().init();
 
     BaseNoGui.initPortableFolder();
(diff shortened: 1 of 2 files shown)


cmaglie:
We should probably do the println only when not running in command line mode: it doesn't make sense to have that printed when launched with `--version` I expect only the version in the output.

May you change the PR accordingly?

tlk:
Good point, will have a look at that. Thanks!

PR #9699, opened 2020-01-31
Offline unit tests for GzippedJsonDownloaderTest and JsonDownloaderTest

tlk:
Running `cd app; ant test` on the master branch results in test errors. I noticed these errors after applying #9698 but they were also present before, only less noticeable:

```
[junit] Running cc.arduino.contributions.GzippedJsonDownloaderTest
[junit] Problem getting data folder: Error getting the Arduino data folder.
[junit] java.lang.NullPointerException
[junit]     at processing.app.BaseNoGui.getSettingsFolder(BaseNoGui.java:300)
[junit]     at cc.arduino.utils.network.FileDownloaderCache.<clinit>(FileDownloaderCache.java:82)
[junit]     at cc.arduino.utils.network.FileDownloader.downloadFile(FileDownloader.java:167)
[junit]     at cc.arduino.utils.network.FileDownloader.download(FileDownloader.java:129)
[junit]     at cc.arduino.contributions.DownloadableContributionsDownloader.download(DownloadableContributionsDownloader.java:147)
[junit]     at cc.arduino.contributions.DownloadableContributionsDownloader.download(DownloadableContributionsDownloader.java:130)
[junit]     at cc.arduino.contributions.JsonDownloader.download(JsonDownloader.java:49)
[junit]     at cc.arduino.contributions.GZippedJsonDownloader.download(GZippedJsonDownloader.java:63)
[junit]     at cc.arduino.contributions.GzippedJsonDownloaderTest.testJsonDownload(GzippedJsonDownloaderTest.java:43)
[junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[junit]     at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[junit]     at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[junit]     at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
[junit]     at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
[junit]     at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
[...]

PR #9727, opened 2020-02-06
Fixed failures in download tests

cmaglie:
This PR should fix 2 failing test:

- `JsonDownloaderTest.testJsonDownload`
- `GzippedJsonDownloaderTest.testJsonDownload`

Supersedes #9699

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -1,6 +1,8 @@
 name: Java CI
 
-on: [push]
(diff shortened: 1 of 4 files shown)


cmaglie:
@tlk it seems that the gh-action is not triggered on pull-requests, I guess we should change the config file and change:

```
on: [push]
```

to

```
on:
  push:
  pull_request:
```

PR #9809, opened 2020-02-26
Fix serial monitor output cut short with multibyte (UTF-8) data

### LINKED ISSUE #9808: Serial monitor output cut short with multibyte (UTF-8) data
I noticed an issue with the serial monitor, where some data might get stuck and not be displayed in the monitor as expected.

To reproduce, run the following sketch (probably works on any native USB Arduino, I tried on an STM32 board which I had at hand):

```
void setup() {
  Serial.begin(9600);
  for (uint8_t i = 0; i < 12; ++i) {
    Serial.print("0123456789");
  }
  Serial.print("°0123456789");
  delay(10000);
  Serial.print("X");
}

void loop() { }
```

This prints just over 128 bytes of data, with a multi-byte character in the first 128 bytes. Then it waits a bit, and prints one more byte.

When running this, the serial monitor will first print:

```
012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789°0123456
```
Then, after 10 seconds, it adds the missing part plus the X:

```
789X
```

For this to occur, the Arduino must send its data quite fast (hence the native USB requirement). I reproduced this once on a Uno at 2Mbps, but could not reliably do that.

The problem is in the code that converts data from bytes to chars: https://github.com/arduino/Arduino/blob/b40f54af06f95d28a07a7f4fed67812614a4a433/arduino-core/src/processing/app/Serial.java#L177-L195

This code reads an arbitrary amount of data using `port.readBytes` into `buf`. It then puts as much bytes as possible into `inFromSerial`, which is then converted to chars which are stored into `outToMessage`, continuing until `outToMessage` is full, which is then passed on to the serial monitor, and the entire thing is continued until `buf` is fully read.

However, in the case simulated with this sketch:
 - `buf` contains 132 bytes (12 x `01234567890` + `°0123456789`, not that `°` is 2 bytes).
 - On the first inner loop, 128 bytes are read into `inFromSerial`, which are decoded into *127* characters in `outToMessage`.
 - On the second inner loop, the remaining 4 bytes are read into `inFromSerial`, but just one is decoded into `outToMessage` because then `outToMessage` is full.
 - These first 128 characters are sent to the serial monitor, as shown above. The last three bytes/characters (`789`) are still in `inFromSerial`.
 - The outer loop checks `next < buf.length`, which is false (since everything has been read from `buf`, so processing stops.
 - When later the `X` is sent, that is copied into `inFromSerial` as well, where the `789` are still waiting, and everything together is decoded and displayed.

To fix this, processing should probably just continue until `inFromSerial` is empty (or actually until no bytes can be read anymore, since there might be a partial UTF-8 character left in `inFromSerial` that must be left there until the rest is received).

matthijskooijman:
This fixes #9808, detailed description in the commit messages.

This seems like a class that could use some automated testing, but I did not see any existing tests (none at all in arduino-core, only in app), nor did I find any meaningful documentation about how to run tests (I only found "ant test", which runs *all* tests and does not show a useful summary to tell if any tests failed AFAICS). Hence, no tests.

### DIFF
--- a/app/test/processing/app/SerialTest.java
+++ b/app/test/processing/app/SerialTest.java
@@ -1,0 +1,58 @@
+/*
+ * This file is part of Arduino.
+ *
+ * Copyright 2020 Arduino LLC (http://www.arduino.cc/)
+ *
+ * Arduino is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As a special exception, you may use this file as part of a free software
+ * library without restriction.  Specifically, if other files instantiate
+ * templates or use macros or inline functions from this file, or you compile
+ * this file and link it with other files to produce an executable, this
+ * file does not by itself cause the resulting executable to be covered by
+ * the GNU General Public License.  This exception does not however
+ * invalidate any other reasons why the executable file might be covered by
+ * the GNU General Public License.
+ */
+
+package processing.app;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class SerialTest {
+  class NullSerial extends Serial {
+    public NullSerial() throws SerialException {
+      super("none", 0, 'n', 0, 0, false, false);
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Also, I realize that this is a performance-critical bit of code (there has been quite some testing in the past to ensure that high-bandwidth serial streams from native USB boards are properly handled), but I think none of these changes are significant in that sense (the process is still the same, except that in some cases, there might be one extra loop to process any leftover data).

Also note that the `decode` method returns some status about why it has terminated (underflow of input, overflow of output), which *might* be able to detect the "there is only a partial multi-byte character left in the input, so no point in decoding that now"-case earlier (with my code, this is detected by decoding it and noticing that no output is produced, which requires one extra loop iteration). However, the [documentation](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html#decode(java.nio.ByteBuffer,%20java.nio.CharBuffer,%20boolean)) was not clear enough for me to really see how to detect this case, so I just went for the easy route (especially since this is also quite the corner case, I think).

cmaglie:
I've added the missing unit test, the patch LGTM, I'll merge as soon as the CI pass.

matthijskooijman:
Test looks good, thanks!

PR #9827, opened 2020-03-02
Dynamic splash image

alranel:
This PR allows to display dynamic splash images according to an URL returned by the server polled for version check. Such behavior can be used to inform users about news in the Arduino world.

To trigger this feature, one more line can be added to the `https://www.arduino.cc/latest.txt` file:

```
10812
splashImage=https://www.arduino.cc/foo-bar.png
```

(This does not break version check by current versions of the IDE, as they only read the first line.)

The `splashImage` URL, if any, is stored in the application preferences and used at the next run (version check is performed _after_ the splash screen disappears).

As of now, no caching mechanism is in place: the remote image is loaded from remote at launch time. Loading is pretty quick, so users see the default image for a bit, and then it gets replaced by the new one.

I also tried to implement a click handler in order to open a URL when user clicks on the splash screen, but it seems that this is not possible with `java.awt.SplashScreen` and we would need to reimplement the splash screen as a `JPanel`.

**Note:** this is just a proof-of-concept. We might want to add signature verification. Do not merge for now :)

Screenshot 2020-03-02 at 18 01 58

### DIFF
--- a/app/src/cc/arduino/view/SplashScreenHelper.java
+++ b/app/src/cc/arduino/view/SplashScreenHelper.java
@@ -33,6 +33,8 @@ package cc.arduino.view;
 
 import java.awt.*;
 import java.awt.geom.Rectangle2D;
+import java.io.IOException;
+import java.net.URL;
 import java.util.Map;
 
 import processing.app.Theme;
@@ -78,6 +80,10 @@ public class SplashScreenHelper {
     drawText(text);
 
     ensureTextIsDiplayed();
+  }
+
+  public void setImageURL(URL imageURL) throws NullPointerException, IOException, IllegalStateException {
+    splash.setImageURL(imageURL);
   }
 
   private void ensureTextIsDiplayed() {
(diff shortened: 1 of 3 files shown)


matthijskooijman:
I like the idea, but some reservations:
 - Users might not be amused, or at least be surprised, when the IDE does significant amounts of data transfer on startup (downloading a few bytes for a version number is negligable, but an image file might not be). Personally, I'm not too worried about this, but I think some users might be. This probably needs at least an opt-out setting somewhere.
 - Will this work gracefully on a broken or slow connection? I think your latest commit adds error handling, but it might be good to doublecheck this.

cmaglie:
The IDE already does a non-neglibile amount of transfer to check for updates on libraries and plaforms, library_index.json.gz is already 1.2M (and growing).

```
$ curl http://downloads.arduino.cc/libraries/library_index.json.gz -s |wc
   4364   24364 1200885
```

cmaglie:
Follow up on #9935

PR #9875, opened 2020-03-14
Set foreground color on board/library titles in managers

matthijskooijman:
In commit 93581b03d (Set foreground color in library/board manager), the
foreground color was set in addition to the background color, to make
sure that the library and board manager would remain readable even with
a non-standard color scheme (e.g. a dark theme).

When that commit was created, this worked properly. However, between
creating that commit and merging it as part of #9272, the title
rendering was changed from being part of the description (which had its
color set up properly) to being part of the title border (which used
default colors) in #9262.

This commit fixes this again by applying the foreground color also to
the TitledBorder component.

In ContributedPlatformTableCellJPanel this also moves the creation of
the TitledBorder into the constructor, and for
ContributedLibraryTableCellJPanel upwards in the constructor (where it
is run unconditionally), so the property can be final.

Before:

image

After:

image


facchinm:
Merged, thanks Matthjis :slightly_smiling_face:

Issue #7846, opened 2018-07-30
Arduino 1.8.5 does not work after install


ixb11170:
The application does not start after install on RPI 3b running Raspbian Jessie.

I followed the instructions on https://www.arduino.cc/en/Guide/Linux, downloaded the ARM version, unpacked, built and installed all successfully.

I got the warning:

    Adding desktop shortcut, menu item and file associations for Arduino IDE...
    touch: cannot touch '/root/.local/share/applications/mimeapps.list': No such file or directory
    /usr/bin/xdg-mime: 803: /usr/bin/xdg-mime: cannot create /root/.local/share/applications/mimeapps.list.new: Directory nonexistent
    done!

Found this message in a few other issues posts but can't resolve the problem.

I tried to start the application from the Desktop, as well as from the command line and neither worked. Let me know if any more details would help.

Cheers.

cmaglie:
Please:
- un-tar the arduino archive in a folder
- run `arduino` in that folder from command line
- copy the output here

dsyleixa:
for me the installation worked this way:
wget http://downloads.arduino.cc/arduino-1.8.5-linuxarm.tar.xz
tar -x -f arduino-1.8.5-linuxarm.tar.xz

no automatic arduino link creation, instead localize the arduino program in the arduino directory by leafpad and then create a link on the desktop

AndersonLovesYou:
Hello, I had a same issues with a Raspberry Pi 4, Quad core 64 bit ARM cortex-A72 cpu.

I followed this tutorial exactly.
https://magpi.raspberrypi.org/articles/program-arduino-uno-raspberry-pi

using sudo apt-get install arduino will install a very outdated version of Arduino on a computer.

Executing the command will result in following error messages.

root@raspberrypi:/opt/arduino-1.8.13# sudo ./install.sh
Adding desktop shortcut, menu item and file associations for Arduino IDE...

touch: cannot touch '/root/.config/mimeapps.list': No such file or directory
/usr/bin/xdg-mime: 848: /usr/bin/xdg-mime: cannot create /root/.config/mimeapps.list.new: Directory nonexistent

 done!

Arduino won't run if you click on the GUI.

Running ARDUINO will result in an error from a source code.

root@raspberrypi:/opt/arduino-1.8.13# **./arduino**
./arduino: line 35: /opt/arduino-1.8.13/java/bin/java: cannot execute binary file: Exec format error


I am down to help if there are any questions but I am a NOOB.

[...]

PR #9908, opened 2020-03-23
adds to install.sh

shawayush:
This change solves two of the issues

- issue #7846 

There was a file missing at the root of the System, all I did was create this new file, and let the install.sh write the command. I hope this solves these two issues, if any changes to be made, please feel free to contact


matthijskooijman:
For future reference: Instead of closing a PR and opening a new one, you can also force push to the same branch to replace the commits in a PR. This is often preferred, since it better keeps history of commits and comments. No problem now, just something to keep in mind for the future.

Issue #9946, opened 2020-03-28
Serial Plotter Plotting Bug


rithviknishad:
When plotting using Serial.println and then pausing the Serial.println (just like a one time trigger), and then resetting the arduino to perform it again, the old serial plotted point is not erased and remains static in the plot window.

![Annotation 2020-03-28 082557](https://user-images.githubusercontent.com/25143503/77813323-c7263d00-70cd-11ea-912f-bc3a5c9fdeae.png)

rithviknishad:
Scnrshot after multiple reset
![Annotation 2020-03-28 084645](https://user-images.githubusercontent.com/25143503/77813772-3a7d7e00-70d1-11ea-91c9-0cea09c0fc6e.png)


What I did to arrive at this error was: (for pausing the SerialPlotter when a SONAR sensor has been activated and it's Echo and Trig Pins are monitored by another arduino)

![image](https://user-images.githubusercontent.com/25143503/77813837-c68fa580-70d1-11ea-9817-5c4f4c4e593f.png)

nikolradulov:
I don't know if this is of any help but the java code for plotter ca be modified such that a clear command can be introduced. In the **~/Arduino/app/src/processing/app/** directory modify the **SerialPlotter.java**  file and add
 `if ( line.equals("<CLEAR>"))
	  {
		  graphs.clear(); // Clear the graphs
		  xCount = 0; // Reset the count of data points
	  }`
in the `public void message` function right after the line trim occurs. In this way you should be able to manually send a clear message and make sure that the graph is cleared even after the resets.
I have not tested this idea on the recent distrib of the IDE,as i just took it from an old project I developed with my friends back in 2017. Hope it helps.

madsdyd:
This is not a bug, but by design. The Arduino GUI, which the Serial Plotter is part of, have no way to know if you reset your board, e.g. by pressing a reset button.

@nikolradulov has made a remote command patch in #9953 but this has some overlap with #7461. It seems someone need to decide upon a roadmap or similar for Serial Plotter.

PR #9953, opened 2020-03-28
Serial Plotter Plotting Bug #9946

nikolaradulov:
Changes tested on Linux and Windows.
Introduced  <CLEAR> as a command to delete all current graph points in and reset data count to 0, in order to try and solve bug [#9946](https://github.com/arduino/Arduino/issues/9946). No other changes to program were made. Linux and Windows users shouldn't be affected by this change. I have no guarantee this command solves the problem mentioned as I can't reproduce the conditions.


madsdyd:
This has some overlap with #7461

PR #9954, opened 2020-03-28
Handle CR without NL printed in EditorConsole

matthijskooijman:
When uploading a sketch using dfu-util (e.g. used by various STM32 cores), progress bar lines would be printed. However, each update would be printed on a new line, rather than overwriting the same line as intended. Even more, the console would not scroll when subsequent lines were added, which made it quite annoying to see the upload progress.

It turns out dfu-util uses lone `\r` to overwrite the previous line. However, `EditorConsole` would simply replace those with `\n` instead. But because `ConsoleOutputStream` only scrolled on `\n`, these extra lines would fall out of view on the bottom of the console.

I originally tried updating the scroll behaviour to include these lines, but then thought it would be even nicer to just properly handle `\r` characters instead. That took a bit more time than I had anticipated, but the result is fairly clean code, I think :-)

With this PR applied, the dfu-util now looks like this:

Peek 2020-03-28 15-14

Without this PR, it was like this (note that most of the action happens *outside* of view below, until the upload is complete):

Peek 2020-03-28 17-13

As an additional test for partial line overwriting, I added this bit of shell script to a script that was used during upload:

```
echo -ne 'ABCDEF\r'
sleep 3;
echo -ne 'DEF\r'
sleep 3;
echo -ne 'X\n'
sleep 3;
echo -ne 'ZZZ\n'
```
Which works as expected:

Peek 2020-03-28 15-13

Again, this would be a good candidate for some testcases. @cmaglie, do you have some notes you could share about how to run the testsuite in a meaningful way (especially just run a single testcase)?

### DIFF
--- a/app/src/processing/app/EditorConsole.java
+++ b/app/src/processing/app/EditorConsole.java
@@ -186,7 +186,7 @@ public class EditorConsole extends JScrollPane {
   }
 
   public String getText() {
-    return consoleTextPane.getText().trim();
+    return consoleTextPane.getText();
   }
 
 }
--- a/app/src/processing/app/EditorConsole.java
+++ b/app/src/processing/app/EditorConsole.java
@@ -27,6 +27,8 @@ import javax.swing.*;
 import javax.swing.text.*;
 import java.awt.*;
 import java.io.PrintStream;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static processing.app.Theme.scale;
 
@@ -37,6 +39,8 @@ public class EditorConsole extends JScrollPane {
 
   private static ConsoleOutputStream out;
   private static ConsoleOutputStream err;
+  private static int startOfLine = 0;
+  private static int insertPosition = 0;
 
   public static synchronized void setCurrentEditorConsole(EditorConsole console) {
     if (out == null) {
@@ -161,6 +165,8 @@ public class EditorConsole extends JScrollPane {
   public void clear() {
     try {
       document.remove(0, document.getLength());
+      startOfLine = 0;
+      insertPosition = 0;
     } catch (BadLocationException e) {
       // ignore the error otherwise this will cause an infinite loop
       // maybe not a good idea in the long run?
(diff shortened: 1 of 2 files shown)


facchinm:
LOVE. THIS. PR

app/src/processing/app/EditorConsole.java
```diff
@@ -179,10 +185,48 @@ public boolean isEmpty() {
...
+    // start of the line to allow overwriting the most recent line (e.g.
+    // for a progress bar). Any CR or NL that are immediately followed
+    // by another NL are bunched together for efficiency, since these
+    // can just be inserted into the document directly and still be
+    // correct.
+    //
+    // This regex is written so it will necessarily match any string
+    // completely if applied repeatedly. This is important because any
+    // part not matched would be silently dropped.
+    Matcher m = Pattern.compile("([^\r\n]*)([\r\n]*\n)?(\r+)?").matcher(str);
```
vdeconinck, line 202: Knowing this will be called for each and every line printed to the console, I wonder if it wouldn't make sense to move the time consuming compiled pattern to a static variable at the top of the file:
`private static Pattern newLinePattern = Pattern.compile("([^\r\n]*)([\r\n]*\n)?(\r+)?");`
Then here in the insertString method, just call 
`Matcher m = newLinePattern.matcher(str);`
Maybe premature optimization is the root of all evil, but repeatedly compiling the same Pattern can be quite expensive...

cmaglie:
Nice! Definitely I want to see this feature implemented!

> As an additional test for partial line overwriting, I added this bit of shell script to a script that was used during upload:

I'd like to see also a test when multiple lines are sent together like:

```
echo -ne 'ABCDEF\rDEF\rX\nZZZ'
sleep 3;
echo -ne 'ZZZ\rYYY\n'
```

> Again, this would be a good candidate for some testcases. @cmaglie, do you have some notes you could share about how to run the testsuite in a meaningful way (especially just run a single testcase)?

You can't run a single test from `ant` with the current `build.xml` AFAIK, I usually run them one by one from eclipse.

app/src/processing/app/EditorConsole.java
matthijskooijman: Good idea, I'll see about that soon.

matthijskooijman:
> I'd like to see also a test when multiple lines are sent together like:

Good point. I might have already tested this, but let's do this regardless.

> You can't run a single test from ant with the current build.xml AFAIK, I usually run them one by one from eclipse.

Ah, I don't have eclipse set up anymore, so that's not an easy way out here :-)

app/src/processing/app/EditorConsole.java
matthijskooijman: Done, just pushed this.

matthijskooijman:
Ok, just pushed an update. I rebased on top of master and added a fixup commit with some improvements and a testcase. Took me a bit longer than expected after I started on this two days ago, but I produced a few more PRs with test-suite fixes and repo cleanup along the way, so that's nice :-p

As far as I'm concerned, I think this is now ready to merge (after squashing the fixup commit, of course).

matthijskooijman:
Just amended some more style fixes into the fixup commit and prepended some more commits to `EditorConsole` to make the testcases work (I had those already, but accidentally split them into the wrong pull request :-p). So *now* it is ready for merging :-)

facchinm:
Rebased and merged :tada:

PR #9959, opened 2020-03-28
Adds to isntall.sh

shawayush:
Resolves issue #7846

### DIFF
--- a/build/linux/dist/install.sh
+++ b/build/linux/dist/install.sh
@@ -205,6 +205,14 @@ simple_uninstall_f() {
 # Update desktop file and mime databases (if possible)
 updatedbs_f() {
 
+  if [ if "${HOME}/root/.config/mimeapps.list"]
+    then
+      continue
+  else
+    touch mimeapps.list
+    chmod +x mimeapps.list
+  fi
+  
   if [ -d "${HOME}/.local/share/applications" ]; then
     if command -v update-desktop-database > /dev/null; then
       update-desktop-database "${HOME}/.local/share/applications"


build/linux/dist/install.sh
```diff
@@ -205,6 +205,14 @@ simple_uninstall_f() {
 # Update desktop file and mime databases (if possible)
 updatedbs_f() {
 
+  if [ if "${HOME}/root/.config/mimeapps.list"]
+    then
```
per1234: Please be consistent with the formatting convention used in the rest of the file (`if [ ... ]; then`)

build/linux/dist/install.sh
```diff
@@ -205,6 +205,14 @@ simple_uninstall_f() {
...
+      continue
+  else
```
per1234: These lines are unnecessary if you invert your logic using the NOT Boolean operator `!` (`if [ ! ...`)

build/linux/dist/install.sh
```diff
@@ -205,6 +205,14 @@ simple_uninstall_f() {
 # Update desktop file and mime databases (if possible)
 updatedbs_f() {
 
+  if [ if "${HOME}/root/.config/mimeapps.list"]
```
per1234: You have multiple syntax errors here. This line causes the error:
```
./install.sh: 208: [: missing ]
```
That is caused by the missing space before the closing bracket. Once that error is fixed, we get a new error:
```
./install.sh: 208: [: if: unexpected operator
```
Did you maybe mean to write [`-f`](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html) instead of `if`?

shawayush: yes, sorry about that, I will change it now.

shawayush:
Hi! I have updated the changes

build/linux/dist/install.sh
```diff
@@ -205,6 +205,11 @@ simple_uninstall_f() {
 # Update desktop file and mime databases (if possible)
 updatedbs_f() {
 
+  if [ ! -f "${HOME}/root/.config/mimeapps.list" ]; then
```
cmaglie, line 208: - `$[HOME}/root/.config/mimeapps.list` folder doesn't look correct to me
- `chmod +x` is wrong, this file is not supposed be executable

Finally, why you create an empty mimeapps.list? how is this supposed to fix #7846?
From what I read in #7846 there is `xdg-mime` that cannot create `/root/.local/share/applications/mimeapps.list.new` becuase the parent directory is nonexistent.

Are you able to reproduce the error reported in the issue #7846 ? If yes, have you tested this fix before submitting?

shawayush: @cmaglie 
yes, I was able to reproduce the error reported, and yes and I did the test before  I submitted my PR, whatever PR I have given for this organization during the GSOC period, have been tested and reproduced the same, so that my PR may be merged efficiently without any problems. I am still working on one of the features in Arduino-cli. Though it might look wrong to you, I have tried my best to tackle that issue I encounter with it and came up with this reasonable solution, which apparently solves the issue.

shawayush: ar
and after adding this patch to the code, this is what is written in that particular file, `mimeapps.list`

PR #10007, opened 2020-04-07
Boards Menu: add hint if core lives in sketchbook

facchinm:
### All Submissions:

For core development the recommended methods is to clone the core repo in `$sketchbook/hardware`. Before https://github.com/arduino/Arduino/commit/9b48e8d0475d11ea2bb2c883a00ad629207d6d09 these cores were listed after all Board Manager cores; now they are interleaved and quite difficult to spot immediately.
This patch adds the `(sketchbook)` label to cores living in sketchbook, using `getAbsolutePath` so symlinked cores can be spotted too).

### DIFF
--- a/arduino-core/src/processing/app/debug/LegacyTargetPlatform.java
+++ b/arduino-core/src/processing/app/debug/LegacyTargetPlatform.java
@@ -245,4 +245,9 @@ public class LegacyTargetPlatform implements TargetPlatform {
       res += "  " + boardId + " = " + boards.get(boardId) + "\n";
     return res + "}";
   }
+
+  @Override
+  public boolean isInSketchbook() {
+	return getFolder().getAbsolutePath().startsWith(BaseNoGui.getDefaultSketchbookFolder().getAbsolutePath());
+  }
 }
--- a/arduino-core/src/processing/app/debug/LegacyTargetPlatform.java
+++ b/arduino-core/src/processing/app/debug/LegacyTargetPlatform.java
@@ -248,6 +248,6 @@ public class LegacyTargetPlatform implements TargetPlatform {
 
   @Override
   public boolean isInSketchbook() {
-	return getFolder().getAbsolutePath().startsWith(BaseNoGui.getDefaultSketchbookFolder().getAbsolutePath());
+	return getFolder().getAbsolutePath().startsWith(BaseNoGui.getSketchbookHardwareFolder().getAbsolutePath());
   }
 }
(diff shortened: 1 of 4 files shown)


matthijskooijman:
Sounds like a good change. I've been modifying the platform.txt of these cores to change the label here, doing this automatically would be even better.

But I'm wondering if `(sketchbook)` is clear to users? I suspect a lot of users do not actually know the term sketchbook. People developing cores likely will, but there might also be people that have just followed some tutorial to install a core into their sketchbook and have no idea.

One alternative that springs to mind would be `(manually installed)` (as opposed to installed through board manager), but I'm not sure if that's really better (does not help finding the core files if needed).

Or how about `(from sketchbook)` (or `(in sketchbook)` or `(installed in sketchbook`))? That might make it clearer that this refers to the location of the core?

facchinm:
`(in sketchbook)` sounds good, I'm amending the commit :wink:

matthijskooijman:
Hm, one more caveat: This now actually checks whether the absolute path is inside the sketchbook, but in my setup all my git checkouts live elsewhere and are symlinked into the sketchbook, so none of them show the `(sketchbook)` addition...

Would it make sense to, instead of looking at the path, keep a flag (or maybe a "source" field of "install dir/package dir/sketchbook") for each platform that is set when the platforms are loaded?

facchinm:
Mine too are symlinked but I can see the label (they didn't show if I replace `getAbsolutePath` with `getCanonicalPath` in the subfolder check ). One alternative idea is to show the path (or a substring) if the sources are not in Arduino15 folder, which is unlikely to contain symlinks

matthijskooijman:
> Mine too are symlinked but I can see the label (they didn't show if I replace getAbsolutePath with getCanonicalPath in the subfolder check ).

Oh, good point. Let me doublecheck my setup and get back to you :-)

arduino-core/src/processing/app/debug/LegacyTargetPlatform.java
```diff
@@ -245,4 +245,9 @@ public String toString() {
       res += "  " + boardId + " = " + boards.get(boardId) + "\n";
     return res + "}";
   }
+
+  @Override
+  public boolean isInSketchbook() {
+	return getFolder().getAbsolutePath().startsWith(BaseNoGui.getDefaultSketchbookFolder().getAbsolutePath());
```
matthijskooijman, line 251: You're using `getDefaultSketchbookFolder()` here, not the actual folder (which I customized through the preferences). I would even suggest using `getSketchbookHardwareFolder()` instead, which is more specific. I just tried this, which works as expected on my system.

```suggestion
	return getFolder().getAbsolutePath().startsWith(BaseNoGui.getSketchbookHardwareFolder().getAbsolutePath());
```

facchinm: :+1:

PR #10033, opened 2020-04-11
Check for nulls when using programmer

tedder:
Try to test and gracefully fail rather than throwing a NullPointerException.

See #10032 for an example of this exception being thrown.

### All Submissions:

### Changes to Core Features:

* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes

I have **not** compiled or run tests or linted this. I'm hoping this is a change that someone can shepherd into an upcoming release.

### DIFF
--- a/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
+++ b/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
@@ -280,11 +280,17 @@ public class SerialUploader extends Uploader {
   private boolean uploadUsingProgrammer(String buildPath, String className) throws Exception {
 
     TargetPlatform targetPlatform = BaseNoGui.getTargetPlatform();
+    if (targetPlatform == null) {
+      throw new RunnerException(tr("Gould not load platform for programmer. Ensure programmer selection is correct for the board."), false);
+    }
     String programmer = PreferencesData.get("programmer");
     if (programmer.contains(":")) {
       String[] split = programmer.split(":", 2);
       targetPlatform = BaseNoGui.getCurrentTargetPlatformFromPackage(split[0]);
       programmer = split[1];
+    }
+    if (programmer == null) {
+      throw new RunnerException(tr("Gould not load programmer. Ensure programmer selection is correct for the board."), false);
     }
 
     PreferencesMap prefs = PreferencesData.getMap();


arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
```diff
@@ -280,12 +280,18 @@ private String waitForUploadPort(String uploadPort, List<String> before, boolean
   private boolean uploadUsingProgrammer(String buildPath, String className) throws Exception {
 
     TargetPlatform targetPlatform = BaseNoGui.getTargetPlatform();
+    if (targetPlatform == null) {
+      throw new RunnerException(tr("Gould not load platform for programmer. Ensure programmer selection is correct for the board."), false);
+    }
     String programmer = PreferencesData.get("programmer");
     if (programmer.contains(":")) {
       String[] split = programmer.split(":", 2);
       targetPlatform = BaseNoGui.getCurrentTargetPlatformFromPackage(split[0]);
```
matthijskooijman, line 289: I think this could again return null when an invalid package is specified in the "programmer". Or does this throw an exception instead?

arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
```diff
@@ -280,12 +280,18 @@ private String waitForUploadPort(String uploadPort, List<String> before, boolean
   private boolean uploadUsingProgrammer(String buildPath, String className) throws Exception {
 
     TargetPlatform targetPlatform = BaseNoGui.getTargetPlatform();
+    if (targetPlatform == null) {
+      throw new RunnerException(tr("Gould not load platform for programmer. Ensure programmer selection is correct for the board."), false);
```
matthijskooijman, line 284: Do you have any idea *why* `targetPlatform` might end up null? Is this when an invalid board is selected? I think this is still unrelated to having a programmer selected?

tedder: [here's where GetTargetPlatform returns null](https://github.com/arduino/Arduino/blob/15133a072044a822990ab77944434b3a1e134cf6/arduino-core/src/processing/app/BaseNoGui.java#L414). I think it might be when a programmer that isn't connected is selected.

arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
tedder: As in the other comment. Trying to be defensive AND give a unique error message for future users who ask for support.

arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
matthijskooijman: I don't think that this is related to the selected programmer, and I'm pretty sure that the IDE doesn't actually check for a connected programmer at all (it does not know how, often programmers have specific USB interfaces that the IDE does not know about).

In this case, I think the "targetPlatform" is based on the target package and platform preferences, which, along with the board preference, together indicate the selected board. So I guess the getTargetPlatform can return null when no board is selected (which might not be possible, since there is always a default board selected I think), or when the currently selected board is no longer available (because the platform was uninstalled, or the board was removed from it in an update).

So I guess the error message should talk about the board selection, rather than the programmer selection?

Have you been able to specifically reproduce this null, to verify that it is handled correctly with your PR?

arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
matthijskooijman: I'm a bit confused. I think this can return `null` *without* that resulting in a proper error message (but still a NullPointerException). Maybe you misunderstood my meaning?

arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java
tedder: I haven't set up a build environment. I was hoping this would be a start for a core maintainer to take on. If you don't think it's helpful, that's fine.

matthijskooijman: AH, I see you already said you didn't compile this, sorry for missing that. I think this is helpful, but the error messages probably need a bit more tweaking to be a bit more appropriate :-)

As for a maintainer taking this on, with all the load that is already there, I would not expect that to happen quickly, but maybe some other contributor can work on top of this (I won't have time for it, probably).

cmaglie:
The best fix for this is changing `BaseNoGui.getTargetPlatform()` to return an `Optional<TargetPlatform>` instead of a `TargetPlatform`.

This change will trigger compile errors on every place where `getTargetPlatform` is used: at this point all the errors should be fixed by porting the old code to the `Optional` API and, at the same, audit the code and see if the `Optinal.empty()` (previously `null`) is handled correctly on every place.

PR #10166, opened 2020-05-07
Repo cleanup

matthijskooijman:
While looking for a place to document how to run tests, I noticed there were some files and directories in the repository root that do not really need to be there. This PR (re)moves them, resulting in a cleaner repository.

Also, I decided to [document the test process on the wiki](https://github.com/arduino/Arduino/wiki/Testing-Arduino), so I added a note to the README that points there (which is a bit off-topic for this PR, but since I also wanted to add some other links to the README, I included that commit in here as well).

Also, maybe we should remove the `hardware` folder too, but I did not do that yet. It currently contains:
 - A README documenting where the AVR and SAM repos went and how to port over pullrequests. The former is now also documented in the main README (with this PR), the latter is probably no longer relevant by now?
 - A `package_index_bundled.json` file, which describes the bundled AVR core. I wonder if that should not really be tracked as part of the ArduinoCore-avr repo (but that might be a fundamental discussion about the relation between these json files and platform.txt). Alternatively, this file could maybe be moved into `build/` somewhere, though I did really try to come up with a good place.


matthijskooijman:
Force pushed some fixes to the README formatting, pointed out by the codacy check.

matthijskooijman:
Force pushed some more fixes to the README formatting, pointed out by the codacy check.

PR #10169, opened 2020-05-07
Improve running tests through ant

matthijskooijman:
This PR improves running tests through ant, by allowing individual tests to be ran, improving the generated output and automatically (re)building the project before running tests.

I originally also planned to add documentation about this in this PR, but I ended up [creating a page on the wiki](https://github.com/arduino/Arduino/wiki/Testing-Arduino) instead (since the building docs also live there).

That page still has minimal docs (additional docs on running tests from Eclipse or other IDEs welcome, I have no experience with that). When this PR is merged, the CLI-section of that page can be replaced by the following.

### Commandline using ant

From the commandline, `ant` can be used for running tests, just like building. For this, go into the `build` directory and run the `test` target:

    build$ ant test

This runs the entire testsuite. To run individual tests, you can use e.g.:

    build$ ant test -Dsingle-test-class=processing.app.CommandLineTest
    build$ ant test -Dsingle-test-class=processing.app.CommandLineTest -Dsingle-test-methods=testCommandLineBuildWithRelativePath,testCommandLinePreferencesSave

This runs all tests from a single test class, or selected test methods (comma separated) from that class respectively.

A class is selected by using its qualified java class name (to find it, look at the `.java` file and join together the `package` name (e.g. `processing.app`) and the `class` name (e.g. `CommandLineTest`). It is not currently possible to select more than one test class to run.

#### Test results
When running all tests on the commandline, only a minimal test summary is printed for each test. When running just a single test, full test output (failures with backtraces, but also full stdout and stderr output) is shown directly.

In either case, a browsable HTML report is generated. This report contains full details for each test run (including stdout/stderr output, link in the lower right). The path to that report is printed after completing the tests.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -24,6 +24,11 @@ jobs:
       run: sudo apt-get install -y xvfb
     - name: Run tests
       working-directory: ./app
-      run: xvfb-run --auto-servernum --server-args "-screen 0 1024x768x24" ant test
+      run: xvfb-run --auto-servernum --server-args "-screen 0 1024x768x24" ant test -Drunning-from-github-action=1
+    - name: Publish results
+      uses: actions/upload-artifact@v1
+      with:
+        name: html-results
+        path: app/test-bin/results/html/
     - name: Cleanup xvfb
       uses: bcomnes/cleanup-xvfb@v1
--- a/app/build.xml
+++ b/app/build.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<project name="Arduino PDE" default="build">
+<project name="Arduino PDE" xmlns:if="ant:if" xmlns:unless="ant:unless" default="build">
 
   <path id="class.path">
     <fileset dir="lib">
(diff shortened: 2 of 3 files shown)


matthijskooijman:
One more disclaimer: I am not quite familiar with ant, so there are probably better ways to approach this. Regardless, it works well as it is now :-)

matthijskooijman:
I just realized that with this commit, when running all tests, does not produce any detailed output at all. The idea was that you can just look at the HTML report for details, but I just now realized that this is not really convenient when running from CI checks (if a test fails, you cannot easily see what failed).

So I just amended the first commit to generate full output for failing tests only (which will probably be easier to read, before this commit it would print full stdout/stderr, but not the failures actually.

I also just realized the HTML report could maybe be added as an output asset of the workflow, so let me see if that would work (not sure if you could easily view HTML though).

matthijskooijman:
Ok, that *seems* to work, I added one more commit that makes the github action publish the HTML report as an asset. It shows up as a zip.

I initially tried to add the one-page HTML version, but does not seem to include the full command output for all tests, which I think can sometimes be relevant. So I now included the multi-page version, but *also* the single-page version since that is a lot easier to open from the zip (you can just open that file, without having to extract all files).

I'm not entirely sure it works, since the build seems to fail now, because https://github.com/arduino-libraries/WiFi101-FirmwareUpdater-Plugin/releases/download/v0.10.10/WiFi101-Updater-ArduinoIDE-Plugin-0.10.10.zip is failing to download somehow? Seems to be unrelated, so I'll just leave this for now and maybe retrigger a build tomorrow :-)

cmaglie:
@matthijskooijman please push on another branch on your fork if you want to see the gh-actions outcome, otherwise this gets very spammy :-)

cmaglie:
Too bad that the frame-less version does not contains the verbose test output (this makes the .zip artifact from github much less handy).

matthijskooijman:
>Too bad that the frame-less version does not contains the verbose test output (this makes the .zip artifact from github much less handy).

Yeah, that's indeed annoying. I just doublechecked the XSL source that generates the report (`org/apache/tools/ant/taskdefs/optional/junit/xsl/junit-noframes.xsl` inside `/usr/share/java/ant-junit.jar`), which indeed does not reference the `system-out` property (that contains stdout) at all, so it is really not included.

If we care enough, we *could* copy the XSL file entirely (setting [`styledir` on the `report` element](https://ant.apache.org/manual/Tasks/junitreport.html) in `build.xml`) and modify it to include the stdout/stderr as well, but I don't think I care enough to learn XSL (or that we really want to carry a copy of this XSL around).

Issue #10168, opened 2020-05-07
Some unittests fail: Automated clicking mis their target


matthijskooijman:
I've been working on making the test suite a bit more usable (Will submit PRs for that next). With some fixes, most of the tests now work fine on my system, *except* for some of the tests that spawn the GUI and try to click things.

What seems to happen, is that the coordinates for something to be clicked get messed up. E.g. instead of opening a menu, it clicks somewhere in the bar above the menu (not straight above, a bit too far to the left too), failing to open the menu and showing e.g.:

```
    [junit] Clicking on menu item <javax.swing.JMenu[name='menuEdit', text='Edit', selected=false, enabled=true, visible=true, showing=true]> never showed a pop-up menu
    [junit] org.fest.swing.exception.ActionFailedException: Clicking on menu item <javax.swing.JMenu[name='menuEdit', text='Edit', selected=false, enabled=true, visible=true, showing=true]> never showed a pop-up menu
    [junit]     at org.fest.swing.exception.ActionFailedException.actionFailure(ActionFailedException.java:33)
    [junit]     at org.fest.swing.driver.JMenuItemDriver.ensurePopupIsShowing(JMenuItemDriver.java:137)
    [junit]     at org.fest.swing.driver.JMenuItemDriver.click(JMenuItemDriver.java:73)
    [junit]     at org.fest.swing.driver.JMenuItemDriver.activateParentIfIsAMenu(JMenuItemDriver.java:97)
    [junit]     at org.fest.swing.driver.JMenuItemDriver.show(JMenuItemDriver.java:79)
    [junit]     at org.fest.swing.driver.JMenuItemDriver.click(JMenuItemDriver.java:71)
    [junit]     at org.fest.swing.fixture.JMenuItemFixture.click(JMenuItemFixture.java:84)
    [junit]     at processing.app.ReduceIndentWith1CharOnLastLineTest.shouldJustCloseTheDialog(ReduceIndentWith1CharOnLastLineTest.java:51)
    [junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    [junit]     at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    [junit]     at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    [junit] 
[...]

PR #10170, opened 2020-05-07
Prevent testsuite from reading user settings and data

matthijskooijman:
While trying to add a testcase for #9954, I noticed that some tests were failing on my system. I also noticed that the tests were reading my regular preferences, `~/.arduino15` directory and sketchbook. This resulted in test failures, since that caused some "Invalid library" messages to be printed and I had "external editor" set making the IDE editor read-only. But also, this is generally a bad idea, since it makes tests unreliable and might even end up messing with a user's settings and data.

Most of this PR are small fixes and refactors, working up to the "Tests: Do not read system user's data" commit that fixes this particular problem. There are also a few commits with small unrelated fixes (which were enabled by these refactorings).

With this PR applied, the testsuite ran correctly, except for the issues caused by #10168.

### DIFF
--- a/app/test/processing/app/AbstractGUITest.java
+++ b/app/test/processing/app/AbstractGUITest.java
@@ -41,25 +41,21 @@ import processing.app.helpers.FileUtils;
 import javax.swing.*;
 import java.util.Random;
 
-public abstract class AbstractGUITest {
+public abstract class AbstractGUITest extends AbstractWithPreferencesTest {
 
   protected ArduinoFrameFixture window;
 
   @Before
   public void startUpTheIDE() throws Exception {
+    // This relies on AbstractWithPreferencesTest to set up the
+    // non-gui-specific stuff.
+
     System.setProperty("mrj.version", "whynot"); //makes sense only on osx. See https://github.com/alexruiz/fest-swing-1.x/issues/2#issuecomment-86532042
-    Runtime.getRuntime().addShutdownHook(new Thread(DeleteFilesOnShutdown.INSTANCE));
 
     FailOnThreadViolationRepaintManager.install();
 
-    BaseNoGui.initPlatform();
-    BaseNoGui.getPlatform().init();
-    PreferencesData.init(null);
     JPopupMenu.setDefaultLightWeightPopupEnabled(false);
-    Theme.init();
     BaseNoGui.getPlatform().setLookAndFeel();
-    Base.untitledFolder = FileUtils.createTempFolder("untitled" + new Random().nextInt(Integer.MAX_VALUE), ".tmp");
-    DeleteFilesOnShutdown.add(Base.untitledFolder);
 
     window = GuiActionRunner.execute(new GuiQuery<ArduinoFrameFixture>() {
       @Override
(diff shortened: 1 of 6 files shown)


matthijskooijman:
Just pushed some more style fixes to satisfy the checks :-)

matthijskooijman:
I just force pushed this again, I had accidentally pushed some commits in here that were intended for #9954 :-)

matthijskooijman:
@cmaglie, it seems you've merged this (and other) pull requests by rebasing and without a merge commit. Looking at the resulting git history, that is clean, but also loses information about which commit were grouped into a pull request and the number of that pull request. Maybe it would be better to always use normal merge commits for pull requests instead (unless you squash-and-merge, since than you just get one commit with the PR number added, of course)?

cmaglie:
It's not so simple unfortunately, doing always the merge commit (as we did in the past when the squash or rebase merge "button" wasn't available) lead to something like:

image

that IMHO fails in the goal of grouping commits based on PRs.
The "easy" alternative is to squash commits as we do in the `arduino-cli`, doing so produes one commit in the master repo with the PR number written in it, but:
* the inner history of the PR will be available is lost and will be available on github only
* it will make bisection difficult for bigger PR (the counter argument is that you should not do big PR, this is a good argument, but sometimes this is impossible as we discovered in the arduino-cli repo).
* the fact that the commits get squashed lead to making commits with little or no explanation written in it (since they gets squashed... who cares...)

The "hard" alternative is to ask the author to rebase the PR *just before* merging. This will be the best because:
* the history will be linear (a sequence of merges)
* the PR number will be written in the merge commit
* the PR commits will be grouped together
* the full history is preserved and it's bisectable
The cons is that this is feasible only if the OP is practical with git and he's reactive. We alwats struggle to ask authors to make clean PRs, so it will be a real pain to also wait the OP to make a rebase before merging.

BTW, if the PR is already rebased we can do the merge commit, I agree, I'll keep an eye on this for the next merges.

matthijskooijman:
Well, in a way the grouping is still preserved, but indeed a lot harder to see. If only github had a rebase-and-then-create-merge-commit option... I was going to contact github to suggest it, but a little bit of googling showed that people have already done this: https://github.com/isaacs/github/issues/1143 (also gitlab and bitbucket implement this option, as a nice global option to apply to all merges).

> The con is that this is feasible only if the OP is practical with git and he's reactive. We always struggle to ask authors to make clean PRs, so it will be a real pain to also wait the OP to make a rebase before merging.

Yeah, that is indeed a problem (and even when it is rebased, then it might be out of date again before you get around to merging it...).

Often, you can also do the rebase yourself though (if the author did not uncheck the "allow pushes" checkbox), but that will require manual work (though you often would do a checkout anyway for testing, I guess).

Looking further, I found [this thread](https://github.com/isaacs/github/issues/88#issuecomment-286639281) about a request to add a "rebase" button to github (that you can press before merging, and then even wait for CI to complete to check the rebase did not break anything). Among the suggestions there, is this action, which can be used directly to trigger rebases using a `/rebase` comment: https://github.com/cirrus-actions/rebase (but see [this issue](https://github.com/cirrus-actions/rebase/issues/49)) or this app that automatically rebases on *every* push to master (but I suspect we have too many open PRs to make that feasible): https://github.com/tibdex/autorebase

PR #10196, opened 2020-05-12
Bugfixes on tests and github actions runner for Windows

cmaglie:
This PR makes automatic testing on Windows working on github-actions.
It requires a bunch of bugfix/improvements:

* Fixed an illegal operation in ant (even if it targets MasOSX the error is triggered on the Windows version of `ant` used on gh actions)
* Disabled CRLF mangling on some files (git adds OS specific line endings when checking out, this is not good for tests that requires exact matching files)
* String encoding by default is not UTF-8 on Windows. This affects SerialTest
* CommandLineTests requires full stdin/out draining otherwise the Process.wait() command will wait forever...
* msvc DLL are now placed with the bundled `java` executable, this way they are found even if `arduino_debug.exe` is launched from another folder
* replaced FEST with assertJ (this makes finally the GUI tests passing on github actions, BTW I see random failures that goes away by re-running the tests)


app/test/processing/app/SerialTest.java
```diff
@@ -51,8 +51,16 @@ protected void message(char[] chars, int length) {
...
+        52, 53, 54, 55, 56, 57, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 48, 49,
+        50, 51, 52, 53, 54, 55, 56, 57, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57,
+        48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 48, 49, 50, 51, 52, 53, 54, 55,
+        56, 57, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 48, 49, 50, 51, 52, 53,
+        54, 55, 56, 57, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 48, 49, 50, 51,
+        52, 53, 54, 55, 56, 57, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, //
+        -62, -80, // UTF8 char
+        48, 49, 50, 51, 52, 53, 54, 55, 56, 57 };
+    s.processSerialEvent(testData);
+    assertEquals(new String(testData, "UTF-8"), s.output);
```
matthijskooijman, line 64: Hm, this is a lot less readable. Can you not fix this by passing a charset to `getBytes()` explicitly? https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#getBytes(java.nio.charset.Charset)

app/.classpath
```diff
@@ -45,14 +45,16 @@
...
 	<classpathentry kind="lib" path="lib/jtouchbar-1.0.0.jar"/>
 	<classpathentry kind="lib" path="lib/commons-lang3-3.8.1.jar"/>
 	<classpathentry kind="lib" path="lib/jssc-2.8.0-arduino4.jar"/>
+	<classpathentry kind="lib" path="test-lib/assertj-swing-junit-3.9.2.jar"/>
+	<classpathentry kind="lib" path="test-lib/assertj-core-1.7.1.jar"/>
+	<classpathentry kind="lib" path="test-lib/assertj-swing-1.2.0.jar"/>
+	<classpathentry kind="lib" path="test-lib/assertj-swing-junit-1.2.0.jar"/>
+	<classpathentry kind="lib" path="test-lib/assertj-swing-junit-4.5-1.2.0.jar"/>
+	<classpathentry kind="lib" path="test-lib/fest-util-1.2.5.jar"/>
+	<classpathentry kind="lib" path="test-lib/junit-4.5.jar"/>
```
matthijskooijman, line 58: This seems to downgrade junit from 4.11 to 4.5. Is that intentional?

.github/workflows/unit-test.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       working-directory: ./build
       run: xvfb-run --auto-servernum --server-args "-screen 0 1024x768x24" ant test -Drunning-from-github-action=1
     - name: Publish results
+      if: always()
```
matthijskooijman, line 25: What does this do exactly? Does `if` default to something like "all previous steps were successful" and this overrides to always run? Maybe this could use a comment in that case?

app/test/processing/app/CommandLineTest.java
```diff
@@ -94,8 +105,8 @@ public Process runArduino(boolean output, boolean success, File wd, String[] ext
 
     Process pr = rt.exec(args.toArray(new String[0]), null, wd);
     if (output) {
-      IOUtils.copy(pr.getInputStream(), System.out);
-      IOUtils.copy(pr.getErrorStream(), System.out);
+      consume(pr.getInputStream());
+      consume(pr.getErrorStream());
     }
     pr.waitFor();
```
matthijskooijman, line 112: Should this also wait for the IO threads to be completed? I *think* this now has a race condition, where `pr.waitFor()` (and thus the testcase) completes before all the output has been consumed, which could lead to output being truncated or even be intermixed into the next testcase?

build/build.xml
```diff
@@ -1161,6 +1155,15 @@
       </condition>
     </fail>
 
+    <copy todir="windows/work">
+      <fileset dir="windows/dist" includes="*.dll" />
+    </copy>
```
matthijskooijman, line 1160: If these files are in the bin directory, are they still needed here? I haven't checked, but won't they be shipped twice in the resulting distribution? Or are they just duplicated in the build dir?

app/test/processing/app/CommandLineTest.java
matthijskooijman: And with the last commit, where these threads also run to capture output, I think this could even result in test failure.

build/build.xml
cmaglie: Yes, they will be shipped twice.

It's quite weird how Windows and `java` decides how to look for native libraries... in theory what you say is correct but I'm quite afraid of removing the two dlls from where they stayed for many years... so in the end I opted to keep in both places and consume a few kilobytes on the distribution files.

app/test/processing/app/CommandLineTest.java
cmaglie: You're right, even if in practice this never happened, we need to join the two threads.
Will add a check for this.

.github/workflows/unit-test.yml
cmaglie: If the previous step fails (i.e. the Unit Test step) then this step (i.e. the Publishing Step) is not executed (for example see https://github.com/cmaglie/Arduino/runs/664274907?check_suite_focus=true)
Before this patch the CI published the results only if all test passes... that's not very informative :-)

app/.classpath
cmaglie: It's a forced dependency from `assertj-swing-junit-4.5-1.2.0.jar`, unfortunately I didn't found an `assertj-swing-junit-4.11-1.2.0.jar`.

Maybe it's ok to build with `juint-4.11`, but since it's not a critical dependency I've just downgraded it.

app/test/processing/app/SerialTest.java
cmaglie: I'll fix this in an incoming PR.

PR #10204, opened 2020-05-14
[editor] Allow formatting only the selection

ricardojlrufino:
### All Submissions:

### DIFF
--- a/app/src/cc/arduino/packages/formatter/AStyle.java
+++ b/app/src/cc/arduino/packages/formatter/AStyle.java
@@ -76,7 +76,16 @@ public class AStyle implements Tool {
 
   @Override
   public void run() {
-    String originalText = editor.getCurrentTab().getText();
+    
+    String originalText = editor.getCurrentTab().getSelectedText();
+    boolean selection = true;
+    
+    // If no selection use all file
+    if(originalText == null || originalText.isEmpty()) {
+      originalText = editor.getCurrentTab().getText();
+      selection = false;
+    }
+    
     String formattedText = aStyleInterface.AStyleMain(originalText, formatterConfiguration);
 
     if (formattedText.equals(originalText)) {
@@ -84,7 +93,10 @@ public class AStyle implements Tool {
       return;
     }
 
-    editor.getCurrentTab().setText(formattedText);
+    if(selection)
+      editor.getCurrentTab().setSelectedText(formattedText);
+    else
+      editor.getCurrentTab().setText(formattedText);
 
     // mark as finished
     editor.statusNotice(tr("Auto Format finished."));


app/src/cc/arduino/packages/formatter/AStyle.java
```diff
@@ -95,4 +107,4 @@ public String getMenuTitle() {
     return tr("Auto Format");
   }
 
-}
+}
```
matthijskooijman, line 110: This removes a trailing newline at the end it seems, which is probably something your editor does automatically. Ideally, this change would not be included (by changing your editor settings to not make such whitespace changes, or using e.g. `git add -p` or similar to include only the relevant parts in the commit). Not a critical issue, though.

app/src/cc/arduino/packages/formatter/AStyle.java
```diff
@@ -76,15 +76,27 @@ public void init(Editor editor) {
 
   @Override
   public void run() {
-    String originalText = editor.getCurrentTab().getText();
+    
+    String originalText = editor.getCurrentTab().getSelectedText();
+    boolean selection = true;
+    
+    // If no selection use all file
+    if(originalText == null || originalText.isEmpty()) {
```
matthijskooijman, line 84: I think there should be a space after `if` here, looks like that is the convention used here.

app/src/cc/arduino/packages/formatter/AStyle.java
ricardojlrufino: Done...

app/src/cc/arduino/packages/formatter/AStyle.java
ricardojlrufino: Done...

ricardojlrufino:
- About Eclipse formatter

Some things change unexpectedly, I use Eclipse. And apparently he took the "Arduino" formatter embedded in the settings, so he should obey the formatting ... I will be monitoring ..
--
One thing that needs to be changed is the maximum number of characters to break the line, 80 is very little, I think it should be 120

image

cmaglie:
Hi @ricardojlrufino !

If I try to format this:

image

I get this:

image

doesn't look correct to me, I expect the indentation to be right. If I select the whole function (up to external `{` `}`) then the indentation is correct. I think the context must be taken into account but don't know if this is easy to do with libAStyle

ricardojlrufino:
Thanks, @cmaglie, for the feedback
I did not analyze the library in depth, but looking over it I did not see this option.

I had seen this problem, and I find it easier to correct using TAB to adjust the indentation.
Trying to identify the context, and adjust it automatically, seems like it’s not that easy.

ricardojlrufino:
Hi, i found a hack: Add comments to disable the formatting..

## Before
image

## After:
image

In this case, you would have to insert these comments dynamically into the text ... and have them formatted, then remove ...
It seems a little ugly, but it was the way I found it

cmaglie:
The purpose of this PR is to make things simpler: select a piece of source code and indent just that (this is to avoid the need to indent everything). IMHO it's a useful feature and a good idea, but adding comments to disable autoindent will make the whole experience very awkward, and completely defeat the usefulness of this PR.

If there is no other way to obtain the right user experince I think we should just drop this PR.

ricardojlrufino:
> but adding comments to disable autoindent will make the whole experience very awkward

@cmaglie , this would be done programmatically and not by the user

cmaglie:
ahhhww, sorry I misunderstood, so your proposed algorithm is:

1. add `indent-off` tag at the beginning
2. wrap user selection between `indent-on` and `indent-off` tags
3. perform auto-indent
4. removed added `indent-off` and `indent-on` tags
5. redraw

Looking it that way it may work, I'm a bit concerned about the "details", like what happen if I partially select a line? maybe we should indent the whole line even if partially selected.
What if I select a real comment in the source code (how the `// INDENT-ON` will be inserted in the middle of a `/* ... */` comment for example)?

ricardojlrufino:
I managed to implement.
The following tests were done, and they all worked as expected.

image

image

image

image

image

matthijskooijman:
> What if I select a real comment in the source code (how the // INDENT-ON will be inserted in the middle of a /* ... */ comment for example)?

Did you also test this?

Would be nice to have some actual unittests for this behaviour too. I think there's already one for formatting the entire source, could be added there (see the README for how to run unittests).

app/src/cc/arduino/packages/formatter/AStyle.java
```diff
@@ -76,15 +80,64 @@ public void init(Editor editor) {
...
+        content.insertString(0, FORMAT_OFF, null);
+        content.insertString(fristLineOffset + offLength, FORMAT_ON, null);
+        content.insertString(lastLineOffset + offLength + onLength, FORMAT_OFF,null);
+        originalText = content.getText(0, content.getLength());
+
+        String formattedText = aStyleInterface.AStyleMain(originalText, formatterConfiguration);
+
+        // Remove format tags
+        formattedText = formattedText.replaceAll(Pattern.quote(FORMAT_OFF), "");
+        formattedText = formattedText.replaceAll(Pattern.quote(FORMAT_ON), "");
```
matthijskooijman, line 127: This seems too general: If the code already contained these format off tags, they would now be removed. I would suggest to remember the offset where the tag was inserted (offset from the end for the third tag), and remove it only from there. As a sanity check, you could check that astyle has indeed left all text up to and including the second tag (and from the third tag to the end) indeed unmodified, otherwise your offsets would not be valid anymore.

app/src/cc/arduino/packages/formatter/AStyle.java
```diff
@@ -76,15 +80,64 @@ public void init(Editor editor) {
...
+
+        textArea.setText(formattedText);
+        textArea.select(selStart, selStart);
```
matthijskooijman, line 130: AFAICS, this clears the selection and puts the cursor at the start? Might be good to keep the selection (requires recalculating the end of the selection, but that is probably easier if you solve my previous comment too).

app/src/cc/arduino/packages/formatter/AStyle.java
```diff
@@ -76,15 +80,64 @@ public void init(Editor editor) {
...
+        textArea.beginAtomicEdit();
+
+        int selStart = editor.getCurrentTab().getSelectionStart();
+        int selEnd = editor.getCurrentTab().getSelectionStop();
+        int lineStart = textArea.getLineOfOffset(selStart);
+        int lineEnd = textArea.getLineOfOffset(selEnd);
+
+        // Calculate offsets from begin and end of each line.
+        int fristLineOffset = textArea.getLineStartOffset(lineStart);
+        int lastLineOffset = textArea.getLineEndOffset(lineEnd);
```
matthijskooijman, line 112: Does this line offset stuff also work correctly when the selection boundaries are just before or just after the newline? Probably something to add a test for all four cases.

app/src/cc/arduino/packages/formatter/AStyle.java
ricardojlrufino: No problem, i put keyword 'DYN' to allow this and not conflict with pre-existing format tags

app/src/cc/arduino/packages/formatter/AStyle.java
ricardojlrufino: This will place the cursor at the beginning of the SELECTION. I would have to recalculate the positioning of the new format, to keep things simple and straightforward I decided not to do it.

ricardojlrufino: No negative point for usability

app/src/cc/arduino/packages/formatter/AStyle.java
matthijskooijman: Ah, I see. I would still think this is somewhat fragile (a user might also put "DYN" in there), but I guess it works well enough in practice. On advantage of doing some offset calculations is that you can at the same time verify that nothing outside of the selection was changed.

ricardojlrufino:
> What if I select a real comment in the source code (how the // INDENT-ON will be inserted in the middle of a /* ... */ comment for example)?

NO WOKS ... =[

image

image

app/src/cc/arduino/packages/formatter/AStyle.java
ricardojlrufino: Is this the example where I need to format 1 line only, and the cursor is at the beginning of the next line? It will force formatting on the 2 lines

matthijskooijman:
Hm. One other approach, in theory, is to format the entire file, and then replace everything outside of the selection with the original unformatted code. But since astyle might change the number of lines, it will be impossible to decide which old lines correspond to which new lines...

Any idea what astyle does with unfinished code? E.g. if you pass code that starts at the start of the file, but then ends halfway a function, comment or expression? If astyle handles that properly (as far as possible), then you could:
 1. Format the code **before** the selection. Save the both the original and the result, but do not apply it.
 2. Format the code from the start of the file to the end of the selection.
 3. Then, in the result of step 2, find the **result** of step 1 and replace it with the **original** of step 1.

Effectively, you'll be formatting everything up to the end of the selection and then undoing everything up to the start of the selection.

I guess this can still break, especially if the selection start ends up halfway an expression (then the result of 1 might not be a prefix of the result of 2 because step 2 has more information and might layout code differently), but if you keep the "expand selection to entire lines" step you have now, I suspect that this might actually produce reasonable results (and *if* it breaks, it can be detected and an error shown).

app/src/cc/arduino/packages/formatter/AStyle.java
matthijskooijman: I think so, yes. I think I would expect that if I select no characters on the second line, just the newline separating both, I would only format the first line, not the second?

For the start of the selection, it's probably mirrored: If I select just the preceding newline and no characters on the previous line, I would expect to not format the previous line.

ricardojlrufino:
I tried your alternative and others, it is complicated to predict all cases.
I ended up finding a simpler solution, is to detect if it is in a comment, and take off until the beginning of it.

Fixed: prevent erros if cursor is in multi-line docs comments, like:
image

I also added new test cases for selection formatting

Stopping here, a lot of work for little gain ..

matthijskooijman:
> I ended up finding a simpler solution, is to detect if it is in a comment, and take off until the beginning of it.

Hm, I would want to avoid that, since trying to parse source code is really tricky to get right (as can be seen by all the fallout of the autogenerated function prototypes...). Though I guess that *just* parsing comments can be done without understanding anything else of the code, and just for the formatting, it's ok if it is not entirely perfect...

Good to see tests, they look good at first glance.

ricardojlrufino:
> Hm, I would want to avoid that, since trying to parse source code is really tricky

In this case, not... see last commit..
RsyntaxTextArea provide a Token model.. making it much easier to know that the cursor is on top of a comment

matthijskooijman:
> RsyntaxTextArea provide a Token model.. making it much easier to know that the cursor is on top of a comment

Ah, that makes it a lot better indeed :-)

PR #10270, opened 2020-05-25
Fix double libraries rescan.

### LINKED ISSUE #10228: [IDE] Slow Startup - BaseNoGui.onBoardOrPortChange  trigger two rescanLibraries()
I noticed that the method is being called 2 twice:
librariesIndexer.rescanLibraries ()

image

I believe this is a bug, because the `librariesIndexer.setArchitecturePriority`, seems to have no influence on the process.

This makes the other issue (#10214) I found a little worse

cmaglie:
I've cherry-picked @ricardojlrufino commits to fix #10228 and:

* added a trivial fix to function call
* fixed indentation

### DIFF
--- a/app/test/cc/arduino/contributions/UpdatableLibraryTest.java
+++ b/app/test/cc/arduino/contributions/UpdatableLibraryTest.java
@@ -36,8 +36,7 @@ public class UpdatableLibraryTest {
     LibrariesIndexer indexer = new LibrariesIndexer(index_SD_only);
     BaseNoGui.librariesIndexer = indexer;
     indexer.parseIndex();
-    indexer.setLibrariesFolders(folders);
-    indexer.rescanLibraries();
+    indexer.setLibrariesFoldersAndRescan(folders);
 
     ContributedLibrary sdLib = indexer.getIndex().getInstalled("SD").get();
     assertTrue("SD lib is installed", sdLib.isLibraryInstalled());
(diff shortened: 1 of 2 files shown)


arduino-core/src/processing/app/BaseNoGui.java
```diff
@@ -678,6 +678,7 @@ static public void onBoardOrPortChange() {
     if (getTargetPlatform() != null) {
       librariesIndexer.setArchitecturePriority(getTargetPlatform().getId());
     }
+
```
matthijskooijman, line 681: This is an unneeded whitespace change.

cmaglie:
I've resolved the whitespace and squashed into a single commit.

> Refactor by renaming setLibrariesFolders to setLibrariesFoldersAndRescan, updating all call sites.

Right, that would be the best thing to do from the beginning, but doing it now it's just more work for a very little gain, so I'll leave it as is :-)

ricardojlrufino:
It has not yet been possible to avoid this, as there are several days of work and some corrections eventually need to be improved. End up having this, a commit, with an attempt that has not yet been resolved, but this will be discovered a little further ... (refactoring flow)

ricardojlrufino:
For me the only way to solve this, was to create another branch, and reapply all modifications, try to do the isolated tests, to have a cleaner commit. But many issues are the problem of having to rewrite everything, just to have a clean commit.

PR #10756, opened 2020-09-13
Use file passed on command line when using --upload or --verify

### LINKED ISSUE #10755: Incorrect "primary" file chosen when compiling via command line
When the Arduino IDE is invoked via the command line for a Verify or Upload task, such as when an external tool like [microsoft/vscode-arduino](/microsoft/vscode-arduino) invokes a build, but the name of the sketch file provided on the command line does not match its containing folder name, the first file from `FileUtils.listFiles` is used as the "primary" file, resulting in undefined behavior (the order returned by `listFiles` seems non-deterministic) and apparent obscure bugs such as the one I filed at microsoft/vscode-arduino#1100.

The requirement to have the `.ino` file match the folder name seems arbitrary and restrictive for command-line compilations when the IDE itself is not being used for development. I developed a simple fix for this on my own branch and plan on filing a PR.

briman0094:
This removes the requirement for .ino files to match their folder name when the Compiler is invoked via the command line with an explicit .ino file path. Fixes #10755 which in turn fixes microsoft/vscode-arduino#1100. When invoking compiler via `--upload` or `--verify`, it doesn't make sense to require the `.ino` sketch name to match the containing folder name, as other IDEs do not have a reason to enforce this requirement. This should allow tools like `microsoft/vscode-arduino` to eventually support multiple `.ino` files per project more gracefully as well.

### All Submissions:

### DIFF
--- a/arduino-core/src/processing/app/Sketch.java
+++ b/arduino-core/src/processing/app/Sketch.java
@@ -22,6 +22,8 @@ public class Sketch {
   public static final List<String> OTHER_ALLOWED_EXTENSIONS = Arrays.asList("c", "cpp", "h", "hh", "hpp", "s");
   public static final List<String> EXTENSIONS = Stream.concat(SKETCH_EXTENSIONS.stream(), OTHER_ALLOWED_EXTENSIONS.stream()).collect(Collectors.toList());
 
+  private final File initialFile;
+
   /**
    * folder that contains this sketch
    */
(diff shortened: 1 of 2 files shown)


matthijskooijman:
I believe this is the wrong approach to fix this.

With this change, the .ino file passed on the commandline becomes somehow special. So this means that you'll potentially get different results compiling a sketch, based on which of the multiple .ino files you pass on the commandline, which is not something that we would want. It also means that a sketch that is considered invalid by the GUI, is considered valid by the CLI, which is inconsistent.

*if* we would decide to lift the requirement, then I believe the compilation should still be deterministic, so the result should be the same, regardless of which .ino file (or the directory name maybe) you specify to indicate the sketch. I think this is also how `arduino-cli` implements this now.

I'm going to go ahead and close this PR for this reason. I'll add a comment to the issue in a minute discussing alternatives.

PR #10849, opened 2020-10-12
Show maintainer instead of author in lib manager

cmaglie:
/cc @luigigubello

### DIFF
--- a/app/src/cc/arduino/contributions/libraries/ui/ContributedLibraryTableCellJPanel.java
+++ b/app/src/cc/arduino/contributions/libraries/ui/ContributedLibraryTableCellJPanel.java
@@ -148,8 +148,8 @@ public class ContributedLibraryTableCellJPanel extends JPanel {
     installButtonPlaceholder.setVisible(!(installable || upgradable));
 
     String name = selected.getName();
-    String author = selected.getAuthor();
-    // String maintainer = selectedLib.getMaintainer();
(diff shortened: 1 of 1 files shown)


luigigubello (approved):

app/src/cc/arduino/contributions/libraries/ui/ContributedLibraryTableCellJPanel.java
```diff
@@ -196,7 +196,7 @@ public ContributedLibraryTableCellJPanel(JTable parentTable, Object value,
         desc += format("{0}", paragraph);
       desc += "<br />";
     }
-    if (author != null && !author.isEmpty()) {
+    if (maintainer != null && !maintainer.isEmpty()) {
       desc = setButtonOrLink(moreInfoButton, desc, moreInfoLbl, website);
     }
```
matthijskooijman, line 201: Not strictly related to this change,  but why does this if exist? Seems like you don't get a "More info" button/link when no author (or now maintainer) is set, but that doesn't make sense to me?

cmaglie: You're right, it should be `website != null && !website.isEmpty()`, I guess we may fix it together with this PR.

luigigubello:
Great work Tooling team! :100: :unicorn:

Issue #10580, opened 2020-08-01
Null Pointer Exception with no AVR Core installed


NicoHood:
Also see: https://bugs.archlinux.org/task/67452

The Arduino IDE should handle the case when no avr core is installed. I remember that worked before, so it must be a problem of some newer versions? Since not everybody wants to use the IDE with AVR we should not rely on it. Especially on archlinux I left the choice to the user to either use the upstream arduino avr core or the archlinux avr core.

NicoHood:
The issue still exists and archlinux users open bug reports/email me over and over again about this issue. I guess it is quite easy to fix, can you please have a look at this again?

NicoHood:
*push* Please are reporting this issue over and over again

cmaglie:
> The issue still exists and archlinux users open bug reports/email me over and over again about this issue. I guess it is quite easy to fix, can you please have a look at this again?

I tried a `light_bundle` build and the IDE seems to start just fine:
![image](https://user-images.githubusercontent.com/423515/115588830-72a0c180-a2cf-11eb-9f2f-a4fed113e92b.png)

did you have a way to reproduce the issue? maybe it's has been fixed after 1.8.13?

NicoHood:
We are also using a light bundle:
https://github.com/archlinux/svntogit-community/blob/packages/arduino/trunk/PKGBUILD

It is not fixed in the recent release, as users still reported the issue. You can reproduce this on a fresh archlinux install. The user folder must not have any arduino related content (like the avr boards) and the package itself should also not ship the arduino avr boards.

cmaglie:
if I build a light bundle from the 1.8.13 I got this error:

```
     [exec] java.lang.NullPointerException
     [exec] 	at processing.app.Base.rebuildProgrammerMenu(Base.java:1705)
     [exec] 	at processing.app.Base.<init>(Base.java:286)
     [exec] 	at processing.app.Base.main(Base.java:150)
```

If I build from `master` the IDE starts cleanly, so I assume it was fixed in a way or another. Is the error above that are you experiencing?

If not, may you send me a stack trace or something else to debug it? I'm having trouble reproducing it on the `master` branch.

NicoHood:
That is the error I get:
```
$ arduino
Picked up JAVA_TOOL_OPTIONS: 
Set log4j store directory /home/user/.arduino15
java.lang.NullPointerException
	at processing.app.Base.rebuildProgrammerMenu(Base.java:1705)
[...]

Issue #10887, opened 2020-10-26
Inconsistencies and unneeed complexity in board option menu handling


matthijskooijman:
While reviewing #10816 (recently used boards), I had a closer look at the custom board menu code and I believe there's a few things about that code that are fishy and fragile, and would deserve improvement. These are a few separate, but heavily related issues, so I'm listing them in detail below.

TL;DR: I would propose to:
 - Not translate menu titles
 - Do not preserve board options across board changes
 - Allow remembering board options for boards selected earlier only through the in-development recently-used-boards list
 - Simplify the plethora of current-board-related preferences and use a single `fqbn` property instead

One overall remark (that I made previously [here](https://github.com/arduino/Arduino/issues/10214#issuecomment-630070899)), is that maybe most of this code can be replaced by code that just lets `arduino-cli` figure out various options and menus available somewhere in the near future. If this code is refactored now, that option should certainly be considered as well, so any work done now can be somewhat smoothly migrated to that later.

### Menu titles are translated
The titles of board menus, as defined in `boards.txt`, are translated. This makes some sense for Arduino-supplied cores, where Arduino can both control the `boards.txt` entries and the translations in the IDE, but for third-party boards, which have no control over the translation, this will result in untranslated entries, mixed entries (some translated, some untranslated) or even incorrect translations (when the board menu use a word that is also used in the IDE, but in a different context).

I would suggest just not translating these titles, just like we did with platform names before (see #9238 and e00304935).

See:

https://github.com/arduino/Arduino/blob/a056a5fdc751b933453830434e36aeffae7512d5/app/src/processing/app/Base.java#L1596
https://github.com/arduino/Arduino/blob/a056a5fdc751b933453830434e36aeffae7512d5/app/src/processing/app/Base.java#L1472

### Comparing titles to look for options
Board menu JMenuItems are created in one place and then looked up in another place. This uses the menu title against the JMenuItem label to find the right one, which is fragile. If two different options within the same core would use the same title (which makes no real sense, but could happen), then options of both would be mixed in the same menu I think. Better to use the id of the menu instead (by storing the id as a client property, or maybe use an (id,platform)-to-menu-item map rather than searching a long list).

See:
 
https://github.com/arduino/Arduino/blob/a056a5fdc751b933453830434e36aeffae7512d5/app/src/processing/app/Base.java#L1596
[...]

PR #10922, opened 2020-11-04
rebuildProgrammerMenu: Handle no current board

matthijskooijman:
rebuildProgrammerMenu: Handle no current board

This can happen happen in some unlikely cases (such as when renaming a
platform in a way that breaks the "select a board when none is selected
logic").

Even though a board should always be selected, code should still handle
no selected board gracefully (rather than raising a NullPointerException
like this used to do).

See #10887 for the underlying issue that caused no board to be selected.

### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/arduino/Arduino/pulls?q=) for the same update/change?

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1702,18 +1702,20 @@ public class Base {
     ButtonGroup group = new ButtonGroup();
 
     TargetBoard board = BaseNoGui.getTargetBoard();
-    TargetPlatform boardPlatform = board.getContainerPlatform();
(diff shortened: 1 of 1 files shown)


NicoHood:
Does this also fix https://github.com/arduino/Arduino/issues/10580 ?

matthijskooijman:
@NicoHood this PR would certainly solve one problem caused by having no platforms installed. There might be more, I haven't tested tihs.

PR #11070, opened 2020-12-14
Typo in SerialPlotter.java

sceniclife:
Just fixing a typo.

### DIFF
--- a/app/src/processing/app/SerialPlotter.java
+++ b/app/src/processing/app/SerialPlotter.java
@@ -373,7 +373,7 @@ public class SerialPlotter extends AbstractMonitor {
     sendButton.addActionListener(listener);
   }
 
-  public void appyPreferences() {
+  public void applyPreferences() {
     // Apply line endings.
     if (PreferencesData.get("serial.line_ending") != null) {
       lineEndings.setSelectedIndex(PreferencesData.getInteger("serial.line_ending"));


matthijskooijman (commented):
Huh? You rename a method, but no calls to this method? It seems it was called under the old name, how does that work? Is it that the superclass defines the method by the proper name and SerialPlotter tried but failed to override this method due to the typo?

What does this mean in practice? Before this fix, the line endings preference was not applied to the plotter? Or was it just that the dropdown did not reflect the actual value, but the right value was used internally?

sceniclife:
@matthijskooijman The method is called on line https://github.com/arduino/Arduino/blob/f4e8a91f101773c47481ae5444e7b4e32ddda0a4/app/src/processing/app/SerialPlotter.java#L338

Because of the typo, the class function was never called and only the superclass is called.
In practice, the `applyPreferences` is just reading the preferences file and setting the dropdown boxes to the current running value. Because the class applyPreferences wasn't called, the line ending dropdown box is not displaying the correct line ending that the user had set.

Before the change, changing the line ending preference does change the preference. Line 333. However, it's possible that `No line ending` is used even if you had a line ending preference set. (send() class function uses the current dropdown selection)

sceniclife:
Change can also be seen here: https://github.com/arduino/Arduino/pull/7461

Maclain27:
> hi! human in coding mode  I am Maclain27 I need ride of this repo.

Issue #9507, opened 2019-12-07
Libraries in sketch folder still not working


jeroenst:
According to the manual libraries can be placed in a libraries folder in the sketch folder.

https://www.arduino.cc/en/Guide/Libraries

> Arduino libraries are managed in three different places: inside the IDE installation folder, inside the core folder and in the libraries folder inside your sketchbook. 

But it doesn't work

Related bug: https://github.com/arduino/Arduino/issues/4936

matthijskooijman:
I assume you refer to the "inside your sketchbook" part? The sketchbook is the directory where your sketches live (e.g. `Documents/Arduino` or something), and libraries can live in there (e.g. `Docuements/Arduino/libraries/my_library`). AFAIU the docs are not trying to say "libraries can be loaded from your **sketch** folder", so I think the docs are correct and this feature is (unfortunately) still missing.

I'm going to close this issue as invalid. If you feel this is not appropriate, let use know through a comment. However note #7755, which is an issue that requests this specific feature and should be used for any further discussion about the feature (so no need to keep this open as a feature request).

jeroenst:
The statement says literally libraries folder inside your sketchbook which imo is the project folder, not the projectS folder because it's not stated sketchbookS folder.

This feature is also very wanted, because for different projects sometimes the same lib is needed but with different version. That's now almost impossible.

Also when a project on github has submodules of libraries included they now have to be copied out of the sketchbook folder to the libraries folder in the folder containing all the sketchbooks.

But still even if the only location is in the sketchbook folder we also want to have global libraries locating in folder containing all the sketchbooks (like the behaviour is now).

matthijskooijman:
> The statement says literally libraries folder inside your sketchbook which imo is the project folder, not the projectS folder because it's not stated sketchbookS folder.

This is a misunderstanding on your side. Check for example the preferences of the Arduino IDE, which lets you set the *sketchbook* directory. The *sketchbook* contains a number of *sketches*, which is what you are calling projects.

One can argue about how intuitive or clearly documented this is, but I am pretty sure this is how the terminology was intended.

As for the rest of your argument: Agreed, but that is for #7755.

Issue #4936, opened 2016-05-10
Allow libraries to be loaded from sketch / project folder


vespakoen:
Hey!

This is an issue that must have wasted hours for many people, as some sources on the internet hint that there is a possibility to do it but it's actually not.
## What is possible now:
- Add libraries to arduino's libraries folder (`ARDUINO/lib/targets/libraries`)
- **NEW** (as we can read [here](https://www.arduino.cc/en/hacking/libraries)) Add libraries to `~/Documents/Arduino/libraries/` or `My Documents\Arduino\libraries` depending on your OS, actually after installing Arduino (using the built in install.sh on linux) it made an `Arduino` folder in my home directory, (`~/Arduino`) so, I should make a libraries folder in there?
- Add libraries to project directory, this seems to be the best "hack" but makes it very hard to find the actual application code.
## What we all want

```
~/MyProject
~/MyProject/libraries/Fat16/Fat16.cpp
~/MyProject/libraries/Fat16/Fat16.h
...long list of other support files...
~/MyProject/MyProject.ino
~/MyProject/FileActuallyRelatedToTheProject.cpp
```

This doesn't seem to be possible without a (big ass) Makefile.
## The problem with the current options:
- If I want to distribute my code, I have to add a **install** section that tells my users where to get and how to install libraries.
- If changes have to be made to libraries (to manually tree-shake as space is sometimes limited, anyone?) users have to determine the location of the libraries folder based on their arduino version and OS, download the modified libraries and put them into the correct folder.
- If 2 packages have a filename that collides, I have to move libraries out of the way and back when I need them again
- If I want to use 2 different versions of a library for different projects, what do I do?

Is there really no way we can have relative includes at the moment?

If not, please allow it or at the very least check if there is a libraries folder **inside of the sketch / project directory** and link those correctly
## Pains
- http://arduino.stackexchange.com/questions/8651/loading-local-libraries
- http://stackoverflow.com/questions/14413372/creating-libraries-in-arduino-sketch-folder?s=1|0.4971
- http://stackoverflow.com/questions/4705790/keeping-all-libraries-in-the-arduino-sketch-directory?s=3|0.3257
- http://stackoverflow.com/questions/20413851/git-combining-multiple-repos
- http://stackoverflow.com/questions/7422841/using-arduino-ethernet-h-under-eclipse
- The list goes on, more here: http://arduino.stackexchange.com/search?q=libraries+sketch
[...]

PR #11110, opened 2020-12-28
Add the ability to find libraries in sketch folder.

Steve132:
### Feature Explaination:

This is a patch to allow the user to search sketches by default in the <Sketch>/libraries folder if it exists.  This feature is referenced and/or requested here:

https://github.com/arduino/arduino-builder/pull/223
https://github.com/arduino/Arduino/issues/4936
https://github.com/arduino/Arduino/issues/4936#issuecomment-436803575
https://github.com/arduino/Arduino/issues/9507
https://github.com/arduino/Arduino/issues/7755
https://github.com/arduino/Arduino/issues/11106

This adds this feature to the Arduino IDE provided that the user has enabled it in their preferences.  This preference is disabled by default in order to avoid breaking past behavior by default (in the exceedingly unlikely case that the user had a libraries directory in their sketch that they did NOT wish to be compiled).

This feature is not needed to be implemented in arduino-cli, because the user of arduino-cli could always manually add "--libraries `pwd`" or similar very easily in their command line build.  However, it may be a good idea to add by default as a potential future feature.

### All Submissions:

* [X ] Have you followed the guidelines in our Contributing document?

### New Feature Submissions:

### DIFF
--- a/app/src/cc/arduino/view/preferences/Preferences.java
+++ b/app/src/cc/arduino/view/preferences/Preferences.java
@@ -135,6 +135,7 @@ public class Preferences extends javax.swing.JDialog {
     checkUpdatesBox = new javax.swing.JCheckBox();
     saveVerifyUploadBox = new javax.swing.JCheckBox();
     accessibleIDEBox = new javax.swing.JCheckBox();
+    loadSketchLibrariesBox = new javax.swing.JCheckBox();
     jLabel1 = new javax.swing.JLabel();
     jLabel2 = new javax.swing.JLabel();
     scaleSpinner = new javax.swing.JSpinner();
@@ -284,6 +285,9 @@ public class Preferences extends javax.swing.JDialog {
 
     accessibleIDEBox.setText(tr("Use accessibility features"));
     checkboxesContainer.add(accessibleIDEBox);
+
+	loadSketchLibrariesBox.setText(tr("Use libraries in sketch folder"));
+	checkboxesContainer.add(loadSketchLibrariesBox);
 
     jLabel1.setText(tr("Interface scale:"));
 
@@ -718,6 +722,7 @@ public class Preferences extends javax.swing.JDialog {
   private javax.swing.JButton browseButton;
   private javax.swing.JCheckBox checkUpdatesBox;
   private javax.swing.JCheckBox accessibleIDEBox;
+  private javax.swing.JCheckBox loadSketchLibrariesBox;
   private javax.swing.JPanel checkboxesContainer;
   private javax.swing.JComboBox comboLanguage;
   private javax.swing.JLabel comboLanguageLabel;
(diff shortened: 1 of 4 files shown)


app/src/cc/arduino/view/preferences/Preferences.java
```diff
@@ -833,6 +838,8 @@ private void savePreferencesData() {
 
     PreferencesData.setBoolean("ide.accessible", accessibleIDEBox.isSelected());
 
+    PreferencesData.setBoolean("build.load_sketch_libraries",loadSketchLibrariesBox.isSelected());
```
matthijskooijman, line 841: Should this be `use_sketch_libraries` instead? These libraries are not loaded by default, they are made available and used when required. "Use" is also the term used in the checkbox text?

matthijskooijman (commented):
I'd be very happy if this was finally supported. I hadn't realized that this would be so easy to implement.

> This preference is disabled by default in order to avoid breaking past behavior by default (in the exceedingly unlikely case that the user had a libraries directory in their sketch that they did NOT wish to be compiled).

I wonder if this shouldn't just be enabled by default. If this is enabled by default, it really makes it easier to distribute self-contained sketches, without having to instruct recipients to explicitly enable this extra preference.

As you say, it seems rather unlikely that sketches have an existing libraries directory (if they have it, I'd think it would contain libraries intended to be used, originally by copying into the sketchbook libraries dir, which would now just be easier. Some sketches might have modified libraries intended to be included with relative includes rather than through the Arduino-generated include paths, but those would typically be in `src/libraries` and even if not, such direct includes would still work and not trigger library detection for these libraries).

Going even further, I wonder if we need a visible preference for this at all? If this is just enabled by default, having a preference in `preferences.txt` (but not shown in the GUI) to disable it just in case could be fine.

> This feature is not needed to be implemented in arduino-cli, because the user of arduino-cli could always manually add "--libraries pwd" or similar very easily in their command line build. However, it may be a good idea to add by default as a potential future feature.

I would really want to see this in arduino-cli as well (also enabled by default IMHO), since then a single sketch will end up behaving the same between -cli and the IDE. I suspect that this is similarly easy to implement in arduino-cli, though.

Did you check what the priority of this extra libraries directory will be? IIRC arduino-builder has some code to apply different priorities to different library folders (In particular the platform-specific libraries are treated specially, probably using the order specified on the commandline for other libraries?).

I haven't tested this yet, will do that soon as well.

Steve132:
> Going even further, I wonder if we need a visible preference for this at all? If this is just enabled by default, having a preference in preferences.txt (but not shown in the GUI) to disable it just in case could be fine.

I honestly tend to agree, but since this is my first contribution to Arduino I thought it might be better to avoid any breaking changes and to make sure I dotted my i's and crossed my t's with regards to possibly making breaking changes.

> I would really want to see this in arduino-cli as well (also enabled by default IMHO), since then a single sketch will end up behaving the same between -cli and the IDE. I suspect that this is similarly easy to implement in arduino-cli, though.

This is true.    I didn't do this yet, but I can if this patch gets merged.   I don't want to make the change to arduino-cli until then.

> Did you check what the priority of this extra libraries directory will be? IIRC arduino-builder has some code to apply different priorities to different library folders (In particular the platform-specific libraries are treated specially, probably using the order specified on the commandline for other libraries?).

I saw that, but the only way I mitigated it is to make sure this comes first.

blaquezzz (approved):

jwktje:
I really would love this feature to be added.

Decee1:
Why is this still not added??!?
It is such a basic function of wanting to modify a library because it is missing a function or because you are working in a team and want to ensure your sketch shares the same library.

Levi-Lesches:
Hi, what's the status of this PR?

stg:
This is a simple, basic and completely obvious feature, that I for one think should be enabled by default. It needs to be implemented yesterday. Not having this has been a huge problem for so long I can't even remember when I first ran into it and wondered how it had been missed. The src/ workaround does not work when libraries reference other libraries.

PR #11200, opened 2021-01-24
Fix status bar custom board preferences disappearing

### LINKED ISSUE #11164: Satus bar is not properly updated
Hi and happy new year.

I've searched if this issue already exists (open or closed) but not found. That's why I open a new one.

When opening Arduino IDE the status bar display properly all settings:
image

When we change a settings (or select the same) then it is not properly updated and only show the board type and selected port:
image

Same after an upload (after a build/verify the status bar is preserved):
image

It seems linked to custom settings not taken in account after any status bar update.

I've tested also with official core: `Arduino Duemilanove or Diecimila`
image

image

Same issue the `Processor` settings is not displayed.

magedrifaat:
This commit fixes #11164

The cause of the bug was that although EditorLineStatus' "paintComponent" function correctly calls "getBoardPreferences" which accounts for any custom board preferences, Editor's "onBoardOrPortChange" incorrectly calls "getTargetBoard" which only account for the board name.

That's why the board settings appear at first but any change to the board settings makes them disappear.
My fix was to simply call "getBoardPreferences" in Editor's "onBoardOrPortChange" just like the implementation in EditorLineStatus.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -2587,12 +2587,7 @@ public class Editor extends JFrame implements RunnerListener {
   // . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
 
   protected void onBoardOrPortChange() {
-    TargetBoard board = BaseNoGui.getTargetBoard();
-    if (board != null)
-      lineStatus.setBoardName(board.getName());
-    else
-      lineStatus.setBoardName("-");
-    lineStatus.setPort(PreferencesData.get("serial.port"));
+    lineStatus.updateBoardAndPort();
     lineStatus.repaint();
   }
 
--- a/app/src/processing/app/EditorLineStatus.java
+++ b/app/src/processing/app/EditorLineStatus.java
@@ -92,12 +92,7 @@ public class EditorLineStatus extends JComponent {
   public void paintComponent(Graphics graphics) {
     Graphics2D g = Theme.setupGraphics2D(graphics);
     if (name.isEmpty() && port.isEmpty()) {
-      PreferencesMap boardPreferences = BaseNoGui.getBoardPreferences();
-      if (boardPreferences != null)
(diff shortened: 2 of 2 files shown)


matthijskooijman (commented):
I've noticed this issue before, good to see it fixed.

Looking at the code, this looks like a reasonable fix. Haven't tested, though.

One thing that might improve this could be to remove the duplicate code entirely, for example by adding a `updateBoardAndPort()` method to `EditorLineStatus` that can be called from both `onBoardOrPortChange` and `paintComponent` (the latter within the existing if).

magedrifaat:
> I've noticed this issue before, good to see it fixed.
> 
> Looking at the code, this looks like a reasonable fix. Haven't tested, though.
> 
> One thing that might improve this could be to remove the duplicate code entirely, for example by adding a `updateBoardAndPort()` method to `EditorLineStatus` that can be called from both `onBoardOrPortChange` and `paintComponent` (the latter within the existing if).

Thanks for the feedback.

The solution you are describing seems like the better option, but the reason I leaned towards this "dirty" fix is because having dealt with Arduino IDE code before, I found that keeping modifications to a minimum usually produces less bugs in the long run.

That being said, I think the reason this bug existed in the first place is because of the duplicated code, the feature was implemented in one place but not the other. So it makes more sense to get rid of the root of the problem to avoid it happening again. I'll just have to be more careful and mind any other "board changes" that might be happening somewhere else.

I'll try it out and update this pull request with the new fix.

magedrifaat:
This is turned out easier than I expected as I found that only the Editor deals with EditorLineStatus.

As per the contribution guidelines, I rebased the commit and made a force push in the fork to update the pull request instead of adding new commits.

matthijskooijman (approved):
Looks perfect to me!

facchinm:
LGTM!

facchinm (approved):

PR #11284, opened 2021-02-18
Optimize applyFilter() to Avoid Redundant String Operations

AdrianJBHughes:
if 'showingHint' == true, then many potentially expensive String operations are being executed on an empty string. This change wraps these operations in an if block which will only run them when needed.

### All Submissions:

### DIFF
--- a/app/src/cc/arduino/contributions/ui/FilterJTextField.java
+++ b/app/src/cc/arduino/contributions/ui/FilterJTextField.java
@@ -102,7 +102,7 @@ public class FilterJTextField extends JTextField {
 
   public void applyFilter() {
     if (!showingHint) {
-      String filter = getText().toLowerCase();+
+      String filter = getText().toLowerCase();
         
       // Replace anything but 0-9, a-z, or : with a space
       filter = filter.replaceAll("[^\\x30-\\x39^\\x61-\\x7a^\\x3a]", " ");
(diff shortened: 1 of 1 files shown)


cmaglie:
Hi @AdrianJBHughes

The change in this PR is not equivalent to the current version: if `showingHint` is `false` the `onFilter(..)` method must be called with an empty `String[]` array to reset the view but, in your version, is not called at all, this will leave the list of contributions "filtered" even if the user clears the text field.

app/src/cc/arduino/contributions/ui/FilterJTextField.java
```diff
@@ -101,13 +101,14 @@ private void spawnTimer() {
...
-    filter = filter.replaceAll("[^\\x30-\\x39^\\x61-\\x7a^\\x3a]", " ");
-
-    onFilter(filter.split(" "));
+    if (!showingHint) {
+      String filter = getText().toLowerCase();
+        
+      // Replace anything but 0-9, a-z, or : with a space
+      filter = filter.replaceAll("[^\\x30-\\x39^\\x61-\\x7a^\\x3a]", " ");
+      
+      onFilter(filter.split(" "));
```
cmaglie, line 110: As I said in the other comment this method should be called with an empty array if `showingHint` is `false`.

cmaglie (changes requested):

app/src/cc/arduino/contributions/ui/FilterJTextField.java
AdrianJBHughes: Thanks for the note, I have addressed the above and pushed the changes.

cmaglie (approved):

PR #11425, opened 2021-04-12
Make Custom menus scrollable

### LINKED ISSUE #11416: list are longer than screen height, and impossible to scroll down
Hi,
I have a similar issue than #11069
Arduino IDE 1.8.13 on Windows 10. Examples menu has the correct behavior but not in Tools menu.
The new STM32duino core release will provide more boards support but now list are longer than screen height, and impossible to scroll down the list like in the example menu.
image
image

I need to find a workaround for this else release is compromised.

/cc @matthijskooijman FYI

_Originally posted by @fpistm in https://github.com/arduino/Arduino/issues/11069#issuecomment-816870815_

facchinm:
Fixes #11416

The patch on MenuScroller.java is needed to avoid a clash between custom menus with the same label.
This behaviour artificially increases the amount of objects that the scroller will calculate.
Eg. if two boards share the same custom menu, that menu will contain the properties for both the boards (since it's parsed twice).

### All Submissions:

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you lint your code locally prior to submission?

### Changes to Core Features:

* [ ] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1473,6 +1473,7 @@ public class Base {
           customMenu.putClientProperty("platform", getPlatformUniqueId(targetPlatform));
           customMenu.putClientProperty("removeOnWindowDeactivation", true);
           boardsCustomMenus.add(customMenu);
+          MenuScroller.setScrollerFor(customMenu);
         }
       }
     }
--- a/app/src/processing/app/tools/MenuScroller.java
+++ b/app/src/processing/app/tools/MenuScroller.java
@@ -34,6 +34,7 @@ public class MenuScroller {
 
   private JPopupMenu menu;
   private Component[] menuItems;
+  private Component[] allMenuItems;
   private MenuScrollItem upItem;
   private MenuScrollItem downItem;
   private final MenuScrollListener menuListener = new MenuScrollListener();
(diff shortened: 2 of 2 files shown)


matthijskooijman (commented):
Right, so the problem is indeed the invisible items. I haven't tested, but I wonder if the filter you added is in the right place? AFAICS this now ensures that `menuItems` omits the invisible items, but `restoreItems` clears *all* items, including invisible ones and then restores all items in `menuItems`, so that restores only the visible ones? If this is indeed the case, I suspect that switching between boards might prevent showing items for the second board that did not exist for the first board?

Shouldn't `setMenuItems` still backup the full list and then have `refreshMenu` operate on the visible list? In practice, that could be done by modifying the loops in `refreshMenu` to skip invisible items (or use streaming functions to simplify that, mabye), or a lot simpler, by letting `setMenuItems` create two lists: One list with all items for restoration and one list with visible items for use in `refreshMenu`?

facchinm:
@matthijskooijman you are indeed (as always :slightly_smiling_face: ) right, but for some unknown reason I'm unable to trigger the expected wrong behaviour. 
May it depend on the destruction of the menu entries when it gets minimised (the `putClientProperty("removeOnWindowDeactivation", true);` line) ?

matthijskooijman:
Hm, I haven't tried reproducing it (and no time to dig in, I'm afraid). Maybe just apply a fix and not worry about reproducing the problem, though? :-)

cmaglie:
@matthijskooijman I've pushed a fix for the issue you described, the only way I managed to reproduce is this way:

1. Install `stm32` platform
2. Select `Generic STM32F0 series` board
3. Open the `Board part number` submenu: it shows a lot of entries
4. Select `Generic STM32F1 series` board
5. The 'Board part number' is gone -> bug

All the other platforms seem to be unaffected.
may you test it on some of your configurations to see if everything is behaving correctly?

matthijskooijman (approved):
Code looks good to me. I played around with the latest STM32 core a bit and it seems to work as expected.

fpistm:
Thanks a lot @facchinm, @cmaglie and @matthijskooijman

PR #11487, opened 2021-05-06
Do not delete unused 3rd party index files

cmaglie:
Previously, after removing a 3rd party URL from the preferences, the cached package_xxx_index.json was removed from the data directory. This fact would interfere with external tools like `arduino-cli` that may have a different set of external URLs.

Now the 3rd party indexes are left, this also leads to an overall simplification of the update/parsing process.


ubidefeo (approved):
tested and working

PR #11564, opened 2021-05-27
Gracefully handle malformed additional board manager urls

cmaglie:
If a malformed URL is added to the Additional Board Manager URL the IDE won't start. This is a regression from 1.8.13.

This PR fix the problem


facchinm (approved):

PaulStoffregen:
1.8.16?

cmaglie:
> 1.8.16?

yes... I'm not sure about the release deadline yet...

